PR #5, opened 2021-09-26
[FLINK-1] Adds the iteration module and API

gaoyunhaii:
This PR initializes the module structure for the iteration.

This PR also increases the relied flink version to 1.13.2, which is the latest stable version. We would further change it to 1.14.0 after it is released. This is necessary since we relies on Flink's checkpoint mechanism to also support failover with iterations, and for bounded iteration, we would need checkpoints after some tasks finished (FLIP-147), which is first provided in 1.14.0.


lindong28:
Thanks for the PR @gaoyunhaii. Could you create JIRA for FLIP-176 and add JIRA to the PR title as appropriate?

And it seems that the CI build has failed. Could you take a look?

flink-ml-iteration/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.apache.flink</groupId>
+        <artifactId>flink-ml-parent</artifactId>
+        <version>0.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>flink-ml-iteration</artifactId>
```
lindong28, line 31: Is there any case that user will want to import just one of the flink-ml-iteration module and the flink-ml-api module, but not the other? Would it be simpler to merge the flink-ml-iteration module with the flink-ml-api module? We can rename the flink-ml-api module as appropriate.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/DataStreamList.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.flink.streaming.api.datastream.DataStream;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/** An utility class to maintain a list of {@link DataStream}, which might have different types. */
+public class DataStreamList {
+    public static DataStreamList of(DataStream<?>... streams) {
+        return new DataStreamList(Arrays.asList(streams));
```
lindong28, line 30: Hmm... It seems that this static method is added to the FLIP-176 since our last discussion of the API section0.

I am not sure we need this static method given that we already have `public DataStreamList(List<DataStream<?>> dataStreams)`. Could you help explain the motivation for this method?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBody.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.iteration;
+
+/** The builder of the subgraph that will be executed inside the iteration. */
+public interface IterationBody {
```
lindong28, line 22: Given that this is a public interface, do we need @PublicEvolving here?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
```diff
@@ -0,0 +1,167 @@
...
+ */
+
+package org.apache.flink.ml.iteration;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/** A helper class to create iterations. */
+@PublicEvolving
+public class Iterations {
+
```
lindong28, line 26: Hmm... it seems that the FLIP uses the class name `IterationUtils`. We will need to update the FLIP first if we agree to use this name.

I would slightly prefer `IterationUtils` because this class provides just static public methods. But I am not strong on this. Can you explain the rational for `Iterations`? We can also ask other reviewers for opinions.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBody.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * This method creates the graph for the iteration body. See {@link Iterations} for how the
```
lindong28, line 25: It seems that the Java doc is different from FLIP-176. It will be nice if we can keep Java doc here consistent with the FLIP.

Among other things, can we keep the requirement that "the number of feedback variable streams returned by this method must equal the number of variable streams given to this method."

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBodyResult.java
```diff
@@ -0,0 +1,74 @@
...
+/** The result of an iteration, specifying the feedbacks and the outputs. */
+public class IterationBodyResult {
+
+    /**
+     * A list of feedback variable streams. These streams will only be used during the iteration
+     * execution and will not be returned to the caller of the iteration body. It is assumed that
+     * the method which executes the iteration body will feed the records of the feedback variable
+     * streams back to the corresponding input variable streams.
+     */
+    private final DataStreamList feedbackVariableStreams;
```
lindong28, line 34: Could you update the FLIP-176 to include e.g. `private final` to keep the API consistent with FLIP-176?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBodyResult.java
```diff
@@ -0,0 +1,74 @@
...
+     * A list of output streams. These streams will be returned to the caller of the methods that
+     * execute the iteration body.
+     */
+    private final DataStreamList outputStreams;
+
+    /**
+     * An optional termination criteria stream. If this stream is not null, it will be used together
+     * with the feedback variable streams to determine when the iteration should terminate.
+     */
+    private final @Nullable DataStream<?> terminationCriteria;
```
lindong28, line 46: The FLIP-176 uses `Optional<DataStream<?>>`. Can you explain what is the rational to choose `@Nullable DataStream<?>`? For example, is it more consistent with the existing pattern in Flink?

And hopefully we can keep the PR consistent with the FLIP.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBodyResult.java
```diff
@@ -0,0 +1,74 @@
...
+    public IterationBodyResult(
+            DataStreamList feedbackVariableStreams,
+            DataStreamList outputStreams,
+            @Nullable DataStream<?> terminationCriteria) {
+        this.feedbackVariableStreams = feedbackVariableStreams;
+        this.outputStreams = outputStreams;
+        this.terminationCriteria = terminationCriteria;
+    }
+
+    public DataStreamList getFeedbackVariableStreams() {
```
lindong28, line 62: Since all members are `final`, can we just make them `public final` and remove the accessor methods here?

lindong28 (commented):
Thanks for the PR!

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * <p>The epoch values are determined as described below. See IterationListener for how the
+     * epoch values are used. 1) All records in the initial variable streams and initial data
+     * streams has epoch=0. 2) For any record emitted by this operator into a non-feedback stream,
+     * the epoch of this emitted record = the epoch of the input record that triggers this emission.
+     * If this record is emitted by onEpochWatermarkIncremented(), then the epoch of this record =
+     * epochWatermark. 3) For any record emitted by this operator into a feedback variable stream,
+     * the epoch of the emitted record = the epoch of the input record that triggers this emission +
+     * 1. If this record is emitted by onEpochWatermarkIncremented(), then the epoch of this record
+     * = epochWatermark.
```
becketqin, line 43: Before we jump in and explain how epoch is determined, can we first give some context on what the epoch stands for and used for? For example:

```
Each of the records involved in the iteration has an epoch, including the records in the variable stream, feedback stream, output stream and the side output emitted by IterationListener. The iterations use the record epoch to mark the progress of the iteration. Different types of iterations assign the epoch slightly different. 
```
Also I think an example would actually help a lot. We may refer the java doc to the FLIP where we have diagrams that are easier to understand.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
```diff
@@ -0,0 +1,167 @@
...
+     *
+     * <p>This method invokes the iteration body with the following parameters: 1) The 1st parameter
+     * is a list of input variable streams, which are created as the union of the initial variable
+     * streams and the corresponding feedback variable streams (returned by the iteration body). 2)
+     * The 2nd parameter is the data streams given to this method.
+     *
+     * <p>The epoch values are determined as described below. See IterationListener for how the
+     * epoch values are used. 1) All records in the initial variable streams has epoch=0. 2) All
+     * records in the data streams has epoch=0. 3) For any record emitted by this operator into a
+     * non-feedback stream, the epoch of this emitted record = the epoch of the input record that
```
becketqin, line 81: Can we just say output stream instead of non-feedback stream to make it consistent with that in the `IterationBodyResult`?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
```diff
@@ -0,0 +1,167 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.iteration;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/** A helper class to create iterations. */
+@PublicEvolving
+public class Iterations {
```
becketqin, line 25: The java doc in this class is critical as users will probably start to learn iterations from here. Can we make the java doc here more structured and well-formatted? For example, add some necessary overview and use bullet points instead of number indices in a long paragraph.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
```diff
@@ -0,0 +1,167 @@
...
+     * @param initVariableStreams The initial variable streams. These streams will be merged with
+     *     the feedback variable streams before being used as the 1st parameter to invoke the
+     *     iteration body.
+     * @param initDataStreams The initial data streams. Records from these streams will be
+     *     repeatedly replayed and used as the 2nd parameter to invoke the iteration body.
+     * @param body The computation logic which takes variable/data streams and returns
+     *     variable/output streams.
+     * @return The list of output streams returned by the iteration boy.
+     */
+    static DataStreamList iterateAndReplayBoundedStreamsUntilTermination(
```
becketqin, line 161: Why is this method package private? Also it seems to me that whether replaying the dataStreams or not is more of a configuration than a separate method.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
becketqin: Also, will there be cases, where there are multiple dataStreams and users only replay some of them. Should we provide a finer granularity on the dataStream replay?

tisonkun:
@gaoyunhaii how this PR can be "FLINK-1"?

becketqin:
@tisonkun I think this is just a placeholder. FLIP-176 has not been voted yet so there is no JIRA ticket opened for the subtasks of it. It should be replaced with the actual ticket number later.

flink-ml-iteration/pom.xml
gaoyunhaii: I agree with that we could first remove `ml` from the package name first. 

And for a separate module, Another reason for me is that I think the users of the ML library including two types: one constructs the pipelines with the existing algorithm provided, and the other type develops their own algorithms. I think the first type of users do not need to include the iteration related module.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/DataStreamList.java
gaoyunhaii: For users to pass the parameters and return IterationBodyResult, they need to construct the `DataStreamList` with several named datastreams. The utility avoids they have to write codes like

```
new DataStreamList(Arrays.asList(datastream1, datastream2));
```

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
gaoyunhaii: It seems to me `IterationUtils` seems more like some utility methods used inside the iteration instead of a factory to create iterations? The name `Iterations` followed some convention of the flink's name to the factory class like `RestartStrategies`.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBody.java
gaoyunhaii: @lindong28 I think this sentence should indeed be kept. Also refer to the following comments, I also agree we have a overall revise to the javadoc in this class.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBodyResult.java
gaoyunhaii: I'll update the FLIP~

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBodyResult.java
gaoyunhaii: The current flink coding style should not tend to use Optional as the class field: https://flink.apache.org/contributing/code-style-and-quality-java.html#java-optional.  In general, optional should be used for the return value of the public api if it does not cause performance issue.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationBodyResult.java
gaoyunhaii: Yes, I think we may first keep it as now~

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
```diff
@@ -0,0 +1,167 @@
...
+@PublicEvolving
+public class Iterations {
+
+    /**
+     * This method uses an iteration body to process records in unbounded data streams.
+     *
+     * <p>This method invokes the iteration body with the following parameters: 1) The 1st parameter
+     * is a list of input variable streams, which are created as the union of the initial variable
+     * streams and the corresponding feedback variable streams (returned by the iteration body). 2)
+     * The 2nd parameter is the data streams given to this method.
```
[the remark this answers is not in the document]
gaoyunhaii: I'll revise the java doc as a whole.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
gaoyunhaii: This is in fact provided as a lower-level functionality in the current implementation. 

The current per-round iteration api serves as a compatibility method to the original dataset api.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/Iterations.java
gaoyunhaii: This should be `public`.

> replaying the dataStreams or not is more of a configuration than a separate method.

In fact the lower-level api implements in this way and treat this method as a special case: https://github.com/apache/flink-ml/pull/16

It seems to me that as a whole, users use all-round operators for performance (since they have fully control on the data cache and replay) and use per-round operators for reusing the operators outside of the iteration (like reduce). Therefore, perhaps we might do some change to the api to 
1. Keep the iterateUnboundedStreams and iterateBoundedStreamsUntilTermination
2. Provides the ability to replay some of the input `dataStreams`. 
3. For `iterateBoundedStreamsUntilTermination`, allows users to specify a subgraph of the IterationBody to use per-round wrapper. 

In this cases, we could remove this method. And if we still want to provide compatibility to the dataset api, we might still need to do some modification to this method to explicitly support bulk / delta iteration.

gaoyunhaii:
> @gaoyunhaii how this PR can be "FLINK-1"?

> @tisonkun I think this is just a placeholder. FLIP-176 has not been voted yet so there is no JIRA ticket opened for the subtasks of it. It should be replaced with the actual ticket number later.

@tisonkun Yes, this is exactly the case...

guoweiM (commented):
Thank you very much @gaoyunhaii  for opening this pr, it looks good for me, I only mentioned some small comments below, you can read it first. 

In addition, as a whole, I have an idea. At present, the class of `Iterations` distinguishes between unbounded stream and bounded stream. From the perspective of streaming & batch unification, this is different from the style of `DataStream` and `Table`. They are only in source to distinguish between. 
Is there any key thinking or trade-off for this? Would you like to give more detailed explanation.
Thanks

PR #6, opened 2021-09-26
[FLINK-2][iteration] Add broadcast output to broadcast events to all the downstream tasks

gaoyunhaii:
Iteration requires to broadcast the epoch-watermark inside iteration no matter what partitioner is used. 

Currently the functionality is implementation via reflection. As a whole, the output of the an operator might be

1. ChainingOutput: if the following operators is chained with this one. In this case, we could direct output the event, which id equivalent to broadcast.
2. RecordWriterOutput: the records are written to an external task. The output wraps an `RecordWriter`, which provides the functionality of broadcast to support watermark / latencymark, and we need to proxy broadcasting to this object via reflection.
3. BroadcastingOutput: The operator itself have multiple downstream operators / tasks. `BroadcastingOutput` has a list of internal outputs, which must be 1 or 2.

We should also maintain the metric of `numOutputRecords` by counting the broadcasted event.


lindong28:
Thanks for the PR @gaoyunhaii. It is not clear why FLIP-176 needs this `BroadcastOutput`. Can you explain it? And is there any existing end-to-end test that shows how to support an iteration job using this `BroadcastOutput`?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/broadcast/BroadcastOutputFactory.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param numRecordsOut the counter for the number of output record.
+     * @return the wrapped broadcast output.
+     */
+    public static <OUT> BroadcastOutput<OUT> createBroadcastOutput(
+            Output<StreamRecord<OUT>> output, Counter numRecordsOut) {
+
+        OutputReflectionContext outputReflectionContext = new OutputReflectionContext();
+
+        List<BroadcastOutput<OUT>> internalOutputs = new ArrayList<>();
+        if (outputReflectionContext.isBroadcastingOutput(output)) {
```
becketqin, line 48: Is it possible to have nested `BroadcastingOutput`?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/broadcast/ChainingBroadcastOutput.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.apache.flink.ml.iteration.broadcast;
+
+import org.apache.flink.streaming.api.operators.Output;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.util.OutputTag;
+
+/** The broadcast output corresponding to a chained output. */
+public class ChainingBroadcastOutput<OUT> implements BroadcastOutput<OUT> {
```
becketqin, line 25: Is this BroadcastOutput only emitting to the downstream operator or to side output, but not both?

becketqin (commented):
I'd like to understand the usage of this broadcast output a bit more. So far my understanding is following:

1. Eventually all the operators inside the iteration loop will be wrapped with a _WrapperOperator_ which is responsible for handling the control events of the iteration. 
2. The control events in this case are just like ordinary `StreamRecords` to the runtime. They are only recognized by the _WrapperOperator_.
3. The broadcast output here is only for that _WrapperOperator_ to construct and use, so it can broadcast the control events to all the downstream subtasks, regardless of to which downstream subtask the data records are emitted.

If so is it the case, is this broadcast output only needed in case of the shuffle? Is my understanding correct?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/broadcast/OutputReflectionContext.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.flink.streaming.api.operators.Output;
+import org.apache.flink.streaming.runtime.io.RecordWriterOutput;
+import org.apache.flink.streaming.runtime.streamrecord.StreamElement;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.util.OutputTag;
+
+import java.lang.reflect.Field;
+
+/** The reflection utilities to parse the output and create the broadcast output */
+public class OutputReflectionContext {
```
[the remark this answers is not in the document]
gaoyunhaii: I'm a bit hesitate here in that I'm not sure java reflection library is thread safe. I think it should not cause to much performance penalty to keep one context for each operator.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/broadcast/BroadcastOutputFactory.java
gaoyunhaii: No, it is not possible with the current implementation.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/BroadcastOutputFactory.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.flink.util.OutputTag;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Factor that creates the corresponding {@link BroadcastOutput} from the {@link Output}. */
+public class BroadcastOutputFactory {
+
+    /**
+     * Creates the wrapper broadcast ooutput from {@code output}.
```
guoweiM, line 36: ooutput -> output

flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/ChainingBroadcastOutput.java
```diff
@@ -0,0 +1,43 @@
...
+    private final OutputTag outputTag;
+
+    ChainingBroadcastOutput(Output<StreamRecord<OUT>> rawOutput, OutputTag outputTag) {
+        this.rawOutput = rawOutput;
+        this.outputTag = outputTag;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void broadcastEmit(StreamRecord<OUT> record) {
```
guoweiM, line 36: If I understand correctly, broadcast means that no matter what the tag is, the data should be broadcast down. If this is the case, I understand that rawOutput.collect should be fine here?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/ChainingBroadcastOutput.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.iteration.broadcast;
+
+import org.apache.flink.streaming.api.operators.Output;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.util.OutputTag;
+
+/** The broadcast output corresponding to a chained output. */
```
guoweiM, line 24: Here I suggest that you can refer to specific classes, such as {@linkto ChainingOutput}

flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/OutputReflectionContext.java
```diff
@@ -0,0 +1,104 @@
...
+                    ReflectionUtils.getClassField(
+                            RecordWriterOutput.class, "serializationDelegate");
+            this.serializationDelegateSerializerField =
+                    ReflectionUtils.getClassField(SerializationDelegate.class, "serializer");
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to initialize the OutputReflectionContext", e);
+        }
+    }
+
+    public boolean isBroadcastingOutput(Output<?> bareOutput) {
```
guoweiM, line 70: bareOutput ---> rawOutput

flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/CountingBroadcastOutput.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public CountingBroadcastOutput(
+            Counter numRecordsOut, List<BroadcastOutput<OUT>> internalOutputs) {
+        this.numRecordsOut = numRecordsOut;
+        this.internalOutputs = internalOutputs;
+    }
+
+    @Override
+    public void broadcastEmit(StreamRecord<OUT> record) throws IOException {
+        numRecordsOut.inc();
```
guoweiM, line 44: Why does the broadcasted record only increment once?

guoweiM (commented):
Thanks @gaoyunhaii for opening this pr. I just give some minor comments below.
IMHO the broadcast event is a general pattern so I hope we could put this back to Flink in the futre.

PR #7, opened 2021-09-26
[FLINK-3][iteration] Implements the feedback mechanism and the controller operators.

gaoyunhaii:
This PR implements the feedback mechanism and the controller operators.

For feedback mechanism, we will rely on the feedback edge implemented in the Flink Statefun project, which wrap the feedback records via mails. Since processing mail have higher priority compared to processing normal record, it ensures we always first process the feedback records. The feedback edge currently does not support spilling to disk, we would add this support in the future versions. 

The controller operators include
1. Input operator: wrap user records into IterationRecord, which contains the round information.
2. Output operator: unwrap the IterationRecord into the user records, and discard the events inside the iteration body.
3. Tail operator: the operator increment the round by 1 and emit the record into the feedback edges.
4. Head operator: union the records from the normal input and feedback, and synchronized the epoch watermark message.

Since an iteration might have multiple head operators, the OperatorCoordinator would do the global synchronized via a static variable. 

Also note that the criteria stream would also have the corresponding head operator. 

Finally, the head operator would be used for all kinds of iterations.


lindong28:
Thanks for the PR @gaoyunhaii. Could you create JIRA for FLIP-176 and add JIRA to the PR title as appropriate?

And it seems that the CI build has failed. Could you take a look?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationRecord.java
```diff
@@ -0,0 +1,145 @@
...
+/** The wrapper for the records in iterative stream. */
+public class IterationRecord<T> {
+
+    /** The type of mini-batch stream records. */
+    public enum Type {
+        RECORD,
+
+        EPOCH_WATERMARK,
+
+        BARRIER
```
lindong28, line 31: hmm... I am not sure we need this barrier. Can you explain its use-case? We can also discuss offline.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/operator/HeadOperator.java
```diff
@@ -0,0 +1,222 @@
...
+                shouldTerminate = globallyAlignedEvent.isTerminated();
+                reusable.replace(
+                        IterationRecord.newEpochWatermark(
+                                globallyAlignedEvent.isTerminated()
+                                        ? Integer.MAX_VALUE
+                                        : globallyAlignedEvent.getRound(),
+                                OperatorUtils.getUniqueSenderId(
+                                        getOperatorID(),
+                                        getRuntimeContext().getIndexOfThisSubtask())),
+                        0);
```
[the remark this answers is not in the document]
gaoyunhaii: Currently we do not plan to provide the support of the timestamp in the first version, thus I think the value should makes no difference. We will consider how to support timestamp in the second phase, and by then we would modify the implementation here.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/IterationRecord.java
```diff
@@ -0,0 +1,145 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.iteration;
+
+import java.util.Objects;
+
+/** The wrapper for the records in iterative stream. */
+public class IterationRecord<T> {
```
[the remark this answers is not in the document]
gaoyunhaii: Logically it would be better, but it is impossible to add a new type of `StreamRecord` without modify the flink code directly.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/typeinfo/IterationRecordSerializer.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        @Override
+        protected int getCurrentOuterSnapshotVersion() {
+            return CURRENT_VERSION;
+        }
+
+        @Override
+        protected TypeSerializer<?>[] getNestedSerializers(
+                IterationRecordSerializer<T> tIterationRecordSerializer) {
```
guoweiM, line 244: tIterationRecordSerializer ---> typo?

flink-ml-iteration/src/test/java/org/apache/flink/iteration/typeinfo/IterationRecordSerializerTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+
+/** Tests the serialization and deserialization for the {@link IterationRecord}. */
+public class IterationRecordSerializerTest {
```
guoweiM, line 40: Maybe we need add a test for un-expected type?

guoweiM (commented):
Thanks @gaoyunhaii for opening this pr. I give some minor comment following.

PR #8, opened 2021-09-27
[FLINK-4][iteration] Add operator wrapper for all-round iterations.

gaoyunhaii:
This PR adds operator wrapper for all-round iterations. For such kind of iteration, one user operator would live across the whole period of the iteration. The wrapper operator would process the epoch watermark, and forward the normal records into the user operator. If the user operator has implemented the IterationListener, it would also notify the user operator on alignment.


flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/operator/allround/OneInputAllRoundWrapperOperator.java
```diff
@@ -0,0 +1,76 @@
...
+            case EPOCH_WATERMARK:
+                onEpochWatermarkEvent(0, element.getValue());
+                break;
+            default:
+                throw new FlinkRuntimeException("Not supported iteration record type: " + element);
+        }
+    }
+
+    @Override
+    public void processWatermark(Watermark mark) throws Exception {
```
becketqin, line 62: It is not clear to me how the watermark is going to look like when there is repeated events. Also not sure if this will cause events to be dropped if a datastream with large event time span is replayed.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/operator/allround/OneInputAllRoundWrapperOperator.java
```diff
@@ -0,0 +1,76 @@
...
+        wrappedOperator.processWatermark(mark);
+    }
+
+    @Override
+    public void processLatencyMarker(LatencyMarker latencyMarker) throws Exception {
+        wrappedOperator.processLatencyMarker(latencyMarker);
```
becketqin, line 68: In this case, will the latency marker will be emitted to the output stream? Or will it also participate in the iteration until the iteration terminates? The behavior seems not well defined.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/operator/allround/MultipleInputAllRoundWrapperOperator.java
```diff
@@ -0,0 +1,116 @@
...
+        for (int i = 0; i < wrappedOperator.getInputs().size(); ++i) {
+            // TODO: Note that here we relies on the assumption that the
+            // stream graph generator labels the input from 1 to n for
+            // the input array, which we map them from 0 to n - 1.
+            proxyInputs.add(new ProxyInput(i));
+        }
+        return proxyInputs;
+    }
+
+    public class ProxyInput<IN> implements Input<IterationRecord<IN>> {
```
becketqin, line 79: Ideally the `ProxyInput` should stay with `ProxyOutput`.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/EpochWatermarkAware.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.function.Supplier;
+
+/**
+ * Operators or UDF implements this method would be provided with an supplier that provides the
+ * current rounds of the current element.
+ */
+public interface EpochWatermarkAware {
+
+    void setEpochWatermarkSupplier(Supplier<Integer> epochWatermarkSupplier);
```
yunfengzhou-hub, line 29: `ProgressTrackerListener.onEpochWatermarkIncrement` provides operators with the information of latest epoch watermark. Why do we still need this class and method to provide the same information?

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/proxy/ProxyOutput.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public void emitLatencyMarker(LatencyMarker latencyMarker) {
+        output.emitLatencyMarker(latencyMarker);
+    }
+
+    @Override
+    public void collect(StreamRecord<T> tStreamRecord) {
+        reuseRecord.getValue().setValue(tStreamRecord.getValue());
+        reuseRecord.getValue().setRound(contextRound);
```
yunfengzhou-hub, line 84: The round or epoch of a record does not increase when it finishes a loop and returns to `HeadOperator`. Instead, it increases in this `ProxyOutput` with `contextRound`, whose value equals to epoch watermark. What is the consideration of this design choice over the other?

From my perspective, If the round of a record always equals to the previous epoch watermark, it might be unnecessary to assign `round` to each record. Operators inside the iteration could keep a context of the current round and make it the default round value for the incoming records.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/EpochWatermarkAware.java
gaoyunhaii: Sorry this class should be `EpochAware`. It is used to provide the epoch of current record.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/proxy/ProxyOutput.java
```diff
@@ -0,0 +1,111 @@
...
+        output.collect(sideOutputCache.tag, sideOutputCache.cachedRecord);
+    }
+
+    @Override
+    public void emitLatencyMarker(LatencyMarker latencyMarker) {
+        output.emitLatencyMarker(latencyMarker);
+    }
+
+    @Override
+    public void collect(StreamRecord<T> tStreamRecord) {
```
guoweiM, line 88: tStreamRecord or streamRecord?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/progresstrack/OperatorEpochWatermarkTracker.java
```diff
@@ -0,0 +1,145 @@
...
+            allInputsLowerBound.updateValue(inputIndex, inputStatus.getInputLowerBound());
+            if (allInputsLowerBound.getLowerBound() > oldLowerBound) {
+                progressTrackerListener.onEpochWatermarkIncrement(
+                        allInputsLowerBound.getLowerBound());
+            }
+        }
+    }
+
+    @VisibleForTesting
+    int[] getNumberOfInputs() {
```
guoweiM, line 74: I am not very sure but maybe we could remove this method, which just test the internal implementation.
What is in my mind is that we could test the `OperatorEpochWatermarkTrackerListener`.

PR #10, opened 2021-09-27
[FLINK-24354][FLIP-174] Improve the WithParams interface

lindong28:

This PR updates the parameter-related clsses according to [FLIP-174](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=181311361).

## Brief change log

This PR mades the following changes:
- Updated the `WithParams` interface.
- Removed the `ParamInfoFactory` and the `Params` classes.
- Replaced the `ParamInfo` class with the `Param` class.
- Added a few pre-defined `Param` subclasses  for common types such as int, double, float, string, double array etc. 
- Added `ParamValidators` to provide factory methods for common validation functions on numerical values.

## Verifying this change

The features added in this test are verified by the following unit tests:
- The `StageTest` verifies the parameter set/get operations and the stage save/load operations.
- The `PipelineTest` verifies Pipeline::fit and PipelineModel::transform. And it also verifies save/load of Pipeline and PipelineModel.

## Documentation


flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
```diff
@@ -97,17 +100,17 @@ public PipelineModel fit(Table... inputs) {
...
-        throw new UnsupportedOperationException();
+    public Map<Param<?>, Object> getUserDefinedParamMap() {
+        return paramMap;
     }
 
-    public static Pipeline load(String path) throws IOException {
-        throw new UnsupportedOperationException();
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.savePipeline(this, stages, path);
```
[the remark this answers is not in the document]
zhipeng93: Another concern is that can we assume that the client can access the stored model data when constructing the JobGraph?  Since you are using `FileReader` in `ReadWriterUtils#loadMetadata()`.

I am not sure but we need to let others know this.

flink-ml-api/pom.xml
```diff
@@ -38,6 +38,34 @@ under the License.
       <version>${flink.version}</version>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.flink</groupId>
+      <artifactId>flink-table-api-java-bridge_${scala.binary.version}</artifactId>
```
zhipeng93, line 44: One minor question here, why do we need these dependencies?

flink-ml-api/src/test/java/org/apache/flink/ml/api/core/StageTest.java
```diff
@@ -0,0 +1,345 @@
...
+
+        String tempDir = Files.createTempDirectory("").toString();
+        String path = Paths.get(tempDir, "test").toString();
+        stage.save(path);
+        try {
+            stage.save(path);
+            Assert.fail("Expected IOException");
+        } catch (IOException e) {
+            // This is expected.
+        }
```
becketqin, line 173: Can we have a separate test method like `testSaveMultipleTimes` for this case?

becketqin (approved):
@lindong28 Thanks for the patch. LGTM overall with a few minor comments. Let's maybe first get the FLIP passed before merging the PR.

flink-ml-api/pom.xml
lindong28: Thanks for the review Zhipeng. You are right that `flink-table-api-java-bridge` and `flink-java` are not needed here. The `flink-table-planner` and `flink-test-utils` can be changed to use `test` scope since they are only needed by unit tests (e.g. PipelineTest).

I will update the PR to fix this issue.

flink-ml-api/src/test/java/org/apache/flink/ml/api/core/StageTest.java
lindong28: Hmm... are you suggesting to add something like the code below? In my opinion this is a bit overkill to add a dedicated method that is only used in `validateStageSaveLoad`. The logic seems pretty simple here. I can add this if you prefer.

```
private static void validateExceptionOnSecondSave(Stage<?> stage, String path) throws IOException {
    stage.save(path);

    try {
        stage.save(path);
        Assert.fail("Expected IOException");
    } catch (IOException e) {
        // This is expected.
    }
}
```

flink-ml-api/src/main/java/org/apache/flink/ml/util/ReadWriteUtils.java
```diff
@@ -0,0 +1,324 @@
...
+    public static <T extends Stage<T>> T loadStageParam(String path) throws IOException {
+        Map<String, ?> metadata = loadMetadata(path, "");
+        String className = (String) metadata.get("className");
+        Map<String, String> paramMap = (Map<String, String>) metadata.get("paramMap");
+
+        try {
+            Class<?> clazz = Class.forName(className);
+            Constructor<T> constructor = (Constructor<T>) clazz.getConstructor();
+            constructor.setAccessible(true);
+            T instance = constructor.newInstance();
```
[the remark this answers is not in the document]
lindong28: Sounds good. I have updated the PR as suggested.

flink-ml-api/src/main/java/org/apache/flink/ml/util/ReadWriteUtils.java
```diff
@@ -0,0 +1,324 @@
...
+
+        if (clazz.getSuperclass() != null) {
+            result.addAll(getPublicParamFields(object, clazz.getSuperclass()));
+        }
+        for (Class<?> cls : clazz.getInterfaces()) {
+            result.addAll(getPublicParamFields(object, cls));
+        }
+        return result;
+    }
+}
```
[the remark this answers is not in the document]
lindong28: Good point. I have updated PR as suggested.

flink-ml-api/src/test/java/org/apache/flink/ml/api/core/StageTest.java
```diff
@@ -0,0 +1,345 @@
...
+            Assert.assertTrue(m2.containsKey(entry.getKey()));
+            Object v1 = entry.getValue();
+            Object v2 = m2.get(entry.getKey());
+            if (v1.getClass().isArray() && v2.getClass().isArray()) {
+                Assert.assertArrayEquals((Object[]) v1, (Object[]) v2);
+            } else {
+                Assert.assertEquals(v1, v2);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
lindong28: `HashMap.equals()` does not work here. If the value of an entry is List<...>, then HashMap.equals() will compare the reference of these two arrays instead of comparing the values of these two arrays. Instead, the existing approach explicitly uses `Assert.assertArrayEquals` to compare the values of these two arrays in this case.

I have verified this behavior by changing the code here to use `HashMap.equals()`.

flink-ml-api/pom.xml
lindong28: The PR has been updated to fix this issue.

lindong28:
Thanks @becketqin for the review!

The FLIP-174 has passed the voting process.

flink-ml-api/src/main/java/org/apache/flink/ml/param/Param.java
```diff
@@ -0,0 +1,98 @@
...
+            String description,
+            T defaultValue,
+            ParamValidator<T> validator) {
+        this.name = name;
+        this.clazz = clazz;
+        this.description = description;
+        this.defaultValue = defaultValue;
+        this.validator = validator;
+
+        if (defaultValue != null && !validator.validate(defaultValue)) {
```
zhipeng93, line 54: We should probably remove `defaultValue != null` here and check whether it is null in each subclass of ParamValidator.

flink-ml-api/src/main/java/org/apache/flink/ml/param/ParamValidators.java
```diff
@@ -0,0 +1,95 @@
...
+public class ParamValidators {
+
+    // Always return true.
+    public static <T> ParamValidator<T> alwaysTrue() {
+        return (value) -> true;
+    }
+
+    // Check if the parameter value is greater than lowerBound.
+    public static <T> ParamValidator<T> gt(double lowerBound) {
+        return (value) -> ((Number) value).doubleValue() > lowerBound;
```
zhipeng93, line 33: should we check whether value is null here?

lindong28: Good catch. The PR has been updated to check `value != null` for all validators except the `notNull` validator. And the unit test `testValidators` has been updated to cover this behavior.

flink-ml-api/src/main/java/org/apache/flink/ml/param/Param.java
lindong28: Hmm.. I think we can not remove the `defaultValue != null` here. The reason is that we should allow the default value of a parameter to be null. This should be OK as long as user explicitly sets a non-null value for this parameter later.

But you are right the PR does not ensure this behavior and needs to be fixed. I have updated `WithParams::get` to validate the parameter using the validator if the parameter value is null.

Now the latest PR will always check the final parameter value using the validator:
- If the parameter value is explicitly set by user, its value will be checked in the `WithParams:set(...)` call.
- Otherwise, if its default value is not null, its value will be checked in the Param constructor.
- Otherwise, its default value must be null, then its value will be checked in the `WithParams::get(...)` call.

zhipeng93: Well, I think removing `defaultValue != null` is fine.
If the default value of a parameter is null, then we use ParamValidators.alwaysTrue()?

lindong28: Discussed offline. We agreed to keep the check as it is. The reason is that we want to allow users not to set the default value (e.g. defaultValue = null) in the parameter definition and still be able to specify a validator which requires the parameter value to be not-null.

Note that the check is useful here in case users specify a not-null default value that does not meet the validator requirement.

pom.xml
```diff
@@ -53,11 +53,8 @@ under the License.
 
   <modules>
     <module>flink-ml-api</module>
-    <module>flink-ml-lib</module>
```
zhipeng93, line 56: We should probably add flink-ml-lib back.

lindong28: Compilation will fail after I add this module back in the PR. I will ask Yun to merge https://github.com/apache/flink-ml/pull/19 first.

lindong28:
Thank you @becketqin and @zhipeng93 for your review!

PR #11, opened 2021-09-27
[FLINK-24649][iteration] Add DraftExecutionEnvironment to support wrapping operators during compile time

gaoyunhaii:
This PR adds a DraftExecutionEnvironment to support wrapping operators during compile time.

The users could first create a subgraph inside the DraftExecutionEnvironment and copy the transformations back to the actual execution environment. During the copying, it could apply some changes to the original graph structure.

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/compile/DraftTransformationTranslator.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/compile/DraftTransformationTranslator.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.iteration.compile;
+
+import org.apache.flink.api.common.ExecutionConfig;
+import org.apache.flink.api.dag.Transformation;
+import org.apache.flink.iteration.operator.OperatorWrapper;
+
+/** Creates the actual transformation according to the draft transformation. */
+public interface DraftTransformationTranslator<TF extends Transformation<?>> {
+
+    Transformation<?> translate(
+            TF draftTransformation, OperatorWrapper<?, ?> operatorWrapper, Context context);
+
+    /** The context for {@link DraftTransformationTranslator}. */
+    interface Context {
+
+        Transformation<?> getActualTransformation(int draftId);
+
+        ExecutionConfig getExecutionConfig();
+
+        default Transformation<?> copyProperties(
+                Transformation<?> actual, Transformation<?> draft) {
+            actual.setName(draft.getName());
+            actual.setParallelism(draft.getParallelism());
+
+            if (draft.getMaxParallelism() > 0) {
+                actual.setMaxParallelism(draft.getMaxParallelism());
+            }
+
+            if (draft.getBufferTimeout() > 0) {
+                actual.setBufferTimeout(draft.getBufferTimeout());
+            }
+
+            if (draft.getSlotSharingGroup().isPresent()) {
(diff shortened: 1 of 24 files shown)


flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/operator/OperatorWrapper.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.streaming.api.operators.StreamOperator;
+import org.apache.flink.streaming.api.operators.StreamOperatorFactory;
+import org.apache.flink.streaming.api.operators.StreamOperatorParameters;
+
+import java.io.Serializable;
+
+/** Wrappers for the given operator factory. */
+public interface OperatorWrapper<T, R> extends Serializable {
```
becketqin, line 32: Is this interface necessary? It looks that this should only have one implementation.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/compile/translator/BroadcastStateTransformationTranslator.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.flink.ml.iteration.compile.DraftTransformationTranslator;
+import org.apache.flink.ml.iteration.operator.OperatorWrapper;
+import org.apache.flink.ml.iteration.operator.WrapperOperatorFactory;
+import org.apache.flink.streaming.api.operators.SimpleOperatorFactory;
+import org.apache.flink.streaming.api.operators.co.CoBroadcastWithNonKeyedOperator;
+import org.apache.flink.streaming.api.transformations.BroadcastStateTransformation;
+import org.apache.flink.streaming.api.transformations.TwoInputTransformation;
+
+/** Draft translator for the {@link BroadcastStateTransformation}. */
+public class BroadcastStateTransformationTranslator
```
becketqin, line 32: Can we make the name of the class clear that it is for the transformation in iteration? e.g. add "Iteration" at the beginning of the class name, or maybe just "IterXXX" would make it clearer.

becketqin (commented):
I personally feel it is a little confusing to have the concept of draft subgraph. Would the following work?

1. Have an `IterationExecutionEnvironment` almost the same as `DraftExecutionEnvironment` except that it overrides the `addOperator()` method with a new `Transformation` instance of the same type but using the `IterationWrapperOperatorFactory` and `IterationRecordTypeInformation`.
2. For all the variable streams and data streams in the iteration declaration, we replace their `StreamExecutionEnvironment` with the `IterationStreamExecutionEnvironment`, and switch it back for the output data streams including the side outputs.

gaoyunhaii:
Very thanks @becketqin for the review! 

> I personally feel it is a little confusing to have the concept of draft subgraph. Would the following work?

It seems to me this would not work since the `StreamExecutionEnvironment` is created by the users, like

```
StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
// add some transformations to the env
// add the iteration
```

Unless we make `IterationExecutionEnvironment` visible to the users and ask users to create the environment initially, we might not be able to achieve both between `switch to a second environment` and `add the transformation back to the original environment`? 

The only possible method to me might be have a second `IterationExecutionEnvironment`, when adding transformation to it, it adds the translated transformation into the original environment, but it seems to me have too much difference with the current `DraftExecutionEnvironment`? Since not all transformation would be added by the `addOperator`, the new `IterationExecutionEnvironment` would still need to search for a subgraph that need to be translated.

flink-ml-iteration/src/main/java/org/apache/flink/ml/iteration/operator/OperatorWrapper.java
gaoyunhaii: There are two implements in the following PRs as a whole:
1. AllRoundOperatorWrapper: with this wrapper all the round is processed by a single operator instance.
2. PerRoundOperatorWrapper: with this wrapper each round of data is processed by a new operator instance. This enables us to reuse the operators defined outside of the iteration.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/compile/DraftTransformationTranslator.java
```diff
@@ -0,0 +1,69 @@
...
+                actual.setBufferTimeout(draft.getBufferTimeout());
+            }
+
+            if (draft.getSlotSharingGroup().isPresent()) {
+                actual.setSlotSharingGroup(draft.getSlotSharingGroup().get());
+            }
+            actual.setCoLocationGroupKey(draft.getCoLocationGroupKey());
+
+            actual.setUid(draft.getUid());
+            if (draft.getUserProvidedNodeHash() != null) {
```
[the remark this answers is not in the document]
guoweiM: ok

guoweiM (approved):

gaoyunhaii:
Very thanks @guoweiM and @becketqin  for the review! Will merge...

PR #22, opened 2021-11-02
[FLINK-24722][iteration] Fixes the issues on supporting keyed stream

gaoyunhaii:
This PR fixes that 
1. currently we also `setCurrentKey` for both normal records and control events (namely EpochWatermark), which might call NullPointerException. 
2. Currently for the wrapper and wrapped operators, we both uses the wrapped key selectors, which would cause errors if the wrapped operators need to process key. 
2. Not initialize the streampartitioner correctly.
3. Typo in getCurrentKey()


guoweiM:
In general I am fine with this fix. But I have a little problem —— Do you think that we should also add a test for `partitionCustom`

guoweiM (approved):
Thanks @gaoyunhaii for adding the test.

PR #25, opened 2021-11-06
[FLINK-24807][iteration] Fix the issues in checkpoints with iteration

gaoyunhaii:
We have met some more issues in supporting checkpoints with iteration:

1. Support snapshotting Replayer Operator and per-round wrapper.
2. Fix the issues that after head tasks finished, the coordinator continues emitting CoordinatorCheckpointEvent, which made the checkpoint fails due to events not sending.
3. Support raw operator state inside the iteration. 

Specially for the second point,  Currently the HeadCoordinator would emit CoordinatorCheckpointEvent to the tasks so that the GloballyAlignedEvent would not be interleave with the checkpoint barrier. However, if the tasks are finished and we continue emitting the event, the checkpoint would fail due to there are failed operator events. To address this issue, we would stop emitting CoordinatorCheckpointEvent after the head operator is terminating, namely it received the GloballyAlignedEvent marking terminating.

The third point is required since operators like `withBroadcast` might rely on raw operator state to snapshot the cached records. This is necessary since the normal operator state always resides in the memory, which might be not enough.

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/ReplayOperator.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/ReplayOperator.java
@@ -18,17 +18,24 @@
 
 package org.apache.flink.iteration.operator;
 
+import org.apache.flink.api.common.operators.MailboxExecutor;
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.common.typeutils.base.IntSerializer;
 import org.apache.flink.core.fs.FileSystem;
 import org.apache.flink.core.fs.Path;
 import org.apache.flink.iteration.IterationRecord;
 import org.apache.flink.iteration.datacache.nonkeyed.DataCacheReader;
+import org.apache.flink.iteration.datacache.nonkeyed.DataCacheSnapshot;
 import org.apache.flink.iteration.datacache.nonkeyed.DataCacheWriter;
 import org.apache.flink.iteration.progresstrack.OperatorEpochWatermarkTracker;
 import org.apache.flink.iteration.progresstrack.OperatorEpochWatermarkTrackerFactory;
 import org.apache.flink.iteration.progresstrack.OperatorEpochWatermarkTrackerListener;
 import org.apache.flink.iteration.typeinfo.IterationRecordSerializer;
 import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.runtime.state.StatePartitionStreamProvider;
+import org.apache.flink.runtime.state.StateSnapshotContext;
 import org.apache.flink.streaming.api.graph.StreamConfig;
 import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
 import org.apache.flink.streaming.api.operators.BoundedMultiInput;
@@ -36,12 +43,18 @@ import org.apache.flink.streaming.api.operators.Output;
 import org.apache.flink.streaming.api.operators.TwoInputStreamOperator;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.tasks.StreamTask;
+import org.apache.flink.streaming.runtime.tasks.mailbox.TaskMailbox;
 import org.apache.flink.util.ExceptionUtils;
+import org.apache.flink.util.FlinkRuntimeException;
+import org.apache.flink.util.function.SupplierWithException;
+
+import org.apache.commons.collections.IteratorUtils;
+
+import javax.annotation.Nullable;
 
 import java.io.IOException;
-import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.Collections;
+import java.util.List;
 
 import static org.apache.flink.util.Preconditions.checkState;
 
@@ -60,11 +73,21 @@ public class ReplayOperator<T> extends AbstractStreamOperator<IterationRecord<T>
 
     private TypeSerializer<T> typeSerializer;
 
-    private Executor dataReplayerExecutor;
+    private MailboxExecutor mailboxExecutor;
 
     private DataCacheWriter<T> dataCacheWriter;
 
-    private AtomicReference<DataCacheReader<T>> currentDataCacheReader;
+    @Nullable private DataCacheReader<T> currentDataCacheReader;
+
+    private int currentEpoch;
+
+    // ------------- states -------------------
+
+    /** Stores the parallelism of the last run so that we could check if parallelism is changed. */
+    private ListState<Integer> parallelismState;
+
+    /** Stores the current epoch. */
+    private ListState<Integer> currentEpochState;
 
     @Override
     public void setup(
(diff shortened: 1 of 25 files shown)


flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/ReplayOperator.java
```diff
@@ -60,11 +73,19 @@
...
     private DataCacheWriter<T> dataCacheWriter;
 
-    private AtomicReference<DataCacheReader<T>> currentDataCacheReader;
+    @Nullable private DataCacheReader<T> currentDataCacheReader;
+
+    private int currentEpoch;
+
+    // ------------- states -------------------
+
+    private ListState<Integer> parallelismState;
```
guoweiM, line 86: Please add some comments

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/ReplayOperator.java
```diff
@@ -88,33 +109,108 @@ public void setup(
...
         }
     }
 
     @Override
-    public void initializeState(StateInitializationContext context) throws Exception {
-        super.initializeState(context);
+    public void snapshotState(StateSnapshotContext context) throws Exception {
+        super.snapshotState(context);
+
+        // Always clear the union list state before set value.
```
guoweiM, line 195: clear -> clears?

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/ReplayOperatorTest.java
```diff
@@ -50,12 +69,206 @@ public void testReplaying() throws Exception {
         final int numRecords = 10;
         OperatorID operatorId = new OperatorID();
 
+        createHarnessAndRun(
```
guoweiM, line 72: extend `TestLoger`

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/ReplayOperatorTest.java
```diff
@@ -50,12 +69,206 @@ public void testReplaying() throws Exception {
...
+                    return null;
+                });
+
+        assertOutputAllRecordsAndEpochWatermark(firstRoundOutput, numRecords, operatorId, 0);
+        assertOutputAllRecordsAndEpochWatermark(secondRoundOutput, numRecords, operatorId, 1);
+    }
+
+    @Test
+    public void testSnapshotAndRestoreOnSecondEpoch() throws Exception {
+        final int numRecords = 20;
```
guoweiM, line 187: Would you like enlight me a little why this test case sets the `numRecords` 20 but others set it 10.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/perround/AbstractPerRoundWrapperOperator.java
```diff
@@ -100,6 +110,15 @@
 
     private transient KeySelector<?, ?> stateKeySelector2;
 
+    // --------------- state ---------------------------
+    private int latestEpochWatermark = -1;
```
guoweiM, line 118: `latestEpochWatermark` is not state?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/perround/AbstractPerRoundWrapperOperator.java
```diff
@@ -100,6 +110,15 @@
...
+
+    private ListState<Integer> parallelismState;
```
guoweiM, line 120: Would you like to add some comments for the following state variables?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/perround/AbstractPerRoundWrapperOperator.java
```diff
@@ -223,7 +246,44 @@ public void initializeState(StreamTaskStateInitializer streamTaskStateManager)
 
     @Override
     public void initializeState(StateInitializationContext context) throws Exception {
-        // Do thing for now since we do not have states.
+        parallelismState =
```
guoweiM, line 261: Do we need to call the super method?

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedPerRoundCheckpointITCase.java
```diff
@@ -0,0 +1,213 @@
...
+    public void setup() {
+        result = sharedObjects.add(new ArrayList<>());
+    }
+
+    @Test
+    public void testFailoverAndRestore() throws Exception {
+        try (MiniCluster miniCluster = new MiniCluster(createMiniClusterConfiguration(2, 2))) {
+            miniCluster.start();
+
+            // Create the test job
```
guoweiM, line 95: Create --> Creates?

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/operators/TwoInputReducePerRoundOperator.java
```diff
@@ -18,11 +18,19 @@
...
+import org.apache.flink.runtime.state.StateSnapshotContext;
 import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
 import org.apache.flink.streaming.api.operators.TwoInputStreamOperator;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.util.OutputTag;
 
+import java.util.Collections;
+
 /**
  * An operators that reduce the received numbers and emit the result into the output, and also emit
```
guoweiM, line 35: operators --> operator

guoweiM (commented):
Thanks @gaoyunhaii for opening this pr. I have some following minor comments below.
BTW I just think that we could not support `Batch` execution mode. So I think it would be nice to tell the user when they set the execution mode to `Batch`. I am sorry that I find it so late. But I think you could open a separate pr for this.
Thanks

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/ReplayOperatorTest.java
gaoyunhaii: It could also be 10, I changed it to 10 to be consistent.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/perround/AbstractPerRoundWrapperOperator.java
gaoyunhaii: Perhaps this is not necessary since the super is empty?

guoweiM: But I think it might be some potential bug, right? So I prefer to call it even if the super is empty WDYT?

gaoyunhaii: I also agree we'd better always call super method, but when I add the code it suddenly comes to the that the operator does not extend the `AbstractStreamOperator`, thus in this case we do not have the super method to call~

guoweiM: thanks for your explaination!

guoweiM (approved):

PR #31, opened 2021-11-11
[FLINK-24842][iteration] Make outputs depends on tails for the iteration body

gaoyunhaii:
This PR forces that the outputs depends on all the tails.

If we want to allow users to use table api with the output of the iteration, since the table api would crop the stream graph and only leave the operators that the output depends on. However, for the iteration we always required the whole iteration participate for alignement


guoweiM (approved):
Thanks for resolving the comments.

PR #32, opened 2021-11-11
[FLINK-24817] Add Estimator and Transformer for Naive Bayes

yunfengzhou-hub:
This PR adds the implementation of Naive Bayes algorithm to Flink ML. This algorithm is implemented in reference to that implemented in Alibaba Alink and Apache Spark, but it uses Flink ML's framework and Flink Datastream.

## Brief change log
This PR adds public classes NaiveBayes. Users can use NaiveBayes to do training and inference according to the algorithm with the same name. Users can also use MultiStringIndexer to convert strings into indices.

## Verifying this change
The changes are tested by unit tests in NaiveBayesTest.

Dependencies (does it add or upgrade a dependency): (yes)
The public API, i.e., is any changed class annotated with @Public(Evolving): (yes)

## Documentation

### DIFF
--- a/flink-ml-core/pom.xml
+++ b/flink-ml-core/pom.xml
@@ -65,5 +65,16 @@ under the License.
       <artifactId>flink-shaded-jackson</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>dev.ludovic.netlib</groupId>
+      <artifactId>blas</artifactId>
+      <version>2.2.0</version>
+    </dependency>
+      <dependency>
+          <groupId>org.apache.flink</groupId>
+          <artifactId>flink-table-api-java-bridge_2.11</artifactId>
+          <version>1.14.0</version>
+      </dependency>
   </dependencies>
 </project>
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.linalg;
+
+import org.apache.flink.util.Preconditions;
+
+/** A utility class that provides BLAS routines over matrices and vectors. */
+public class BLAS {
+    /** For level-1 function dspmv, use javaBLAS for better performance. */
+    private static final dev.ludovic.netlib.BLAS JAVA_BLAS =
+            dev.ludovic.netlib.JavaBLAS.getInstance();
+
+    /** y += a * x . */
+    public static void axpy(double a, DenseVector x, DenseVector y) {
+        Preconditions.checkArgument(x.size() == y.size(), "Vector size mismatched.");
+        JAVA_BLAS.daxpy(x.size(), a, x.values, 1, y.values, 1);
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
@@ -1,0 +1,360 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.classification.naivebayes;
+
+import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.functions.MapPartitionFunction;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeinfo.Types;
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.api.java.tuple.Tuple3;
+import org.apache.flink.api.java.tuple.Tuple4;
+import org.apache.flink.ml.api.Estimator;
+import org.apache.flink.ml.common.datastream.MapPartitionFunctionWrapper;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An Estimator which implements the naive bayes classification algorithm.
+ *
+ * <p>See https://en.wikipedia.org/wiki/Naive_Bayes_classifier.
+ */
+public class NaiveBayes
+        implements Estimator<NaiveBayes, NaiveBayesModel>, NaiveBayesParams<NaiveBayes> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public NaiveBayes() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public NaiveBayesModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+
+        final String featuresCol = getFeaturesCol();
+        final String labelCol = getLabelCol();
+        final double smoothing = getSmoothing();
+
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple2<Vector, Double>> input =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                new MapFunction<Row, Tuple2<Vector, Double>>() {
+                                    @Override
+                                    public Tuple2<Vector, Double> map(Row row) throws Exception {
+                                        Number number = (Number) row.getField(labelCol);
+                                        Preconditions.checkNotNull(
+                                                number, "Input data should contain label value.");
+                                        Preconditions.checkArgument(
+                                                number.intValue() == number.doubleValue(),
+                                                "Label value should be indexed number.");
+                                        return new Tuple2<>(
+                                                (Vector) row.getField(featuresCol),
+                                                number.doubleValue());
+                                    }
+                                });
+
+        DataStream<NaiveBayesModelData> modelData =
+                input.flatMap(new ExtractFeatureFunction())
+                        .keyBy(value -> new Tuple2<>(value.f0, value.f1).hashCode())
+                        .transform(
+                                "GenerateFeatureWeightMapFunction",
+                                Types.TUPLE(
+                                        Types.DOUBLE,
+                                        Types.INT,
+                                        Types.MAP(Types.DOUBLE, Types.DOUBLE),
+                                        Types.INT),
+                                new MapPartitionFunctionWrapper<>(
+                                        new GenerateFeatureWeightMapFunction()))
+                        .keyBy(value -> value.f0)
+                        .transform(
+                                "AggregateIntoArrayFunction",
+                                Types.TUPLE(
+                                        Types.DOUBLE,
+                                        Types.INT,
+                                        Types.OBJECT_ARRAY(Types.MAP(Types.DOUBLE, Types.DOUBLE))),
+                                new MapPartitionFunctionWrapper<>(new AggregateIntoArrayFunction()))
+                        .transform(
+                                "GenerateModelFunction",
+                                TypeInformation.of(NaiveBayesModelData.class),
+                                new MapPartitionFunctionWrapper<>(
+                                        new GenerateModelFunction(smoothing)))
+                        .setParallelism(1);
+
+        NaiveBayesModel model =
+                new NaiveBayesModel()
+                        .setModelData(NaiveBayesModelData.getModelDataTable(modelData));
+        ReadWriteUtils.updateExistingParams(model, paramMap);
+        return model;
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static NaiveBayes load(StreamExecutionEnvironment env, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /**
+     * Function to extract feature values from input rows.
+     *
+     * <p>Output records are tuples with the following fields in order:
+     *
+     * <ul>
+     *   <li>label value
+     *   <li>feature column index
+     *   <li>feature value
+     * </ul>
+     */
+    private static class ExtractFeatureFunction
+            implements FlatMapFunction<Tuple2<Vector, Double>, Tuple3<Double, Integer, Double>> {
+        @Override
+        public void flatMap(
+                Tuple2<Vector, Double> value,
+                Collector<Tuple3<Double, Integer, Double>> collector) {
+            Preconditions.checkNotNull(value.f1);
+            for (int i = 0; i < value.f0.size(); i++) {
+                collector.collect(new Tuple3<>(value.f1, i, value.f0.get(i)));
+            }
+        }
+    }
+
+    /**
+     * Function that aggregates entries of feature value and weight into maps.
+     *
+     * <p>Input records should have the same label value and feature column index.
+     *
+     * <p>Input records are tuples with the following fields in order:
+     *
+     * <ul>
+     *   <li>label value
+     *   <li>feature column index
+     *   <li>feature value
+     * </ul>
+     *
+     * <p>Output records are tuples with the following fields in order:
+     *
+     * <ul>
+     *   <li>label value
+     *   <li>feature column index
+     *   <li>map of (feature value, weight)
+     *   <li>number of records
+     * </ul>
+     */
+    private static class GenerateFeatureWeightMapFunction
+            implements MapPartitionFunction<
+                    Tuple3<Double, Integer, Double>,
+                    Tuple4<Double, Integer, Map<Double, Double>, Integer>> {
+
+        @Override
+        public void mapPartition(
+                Iterable<Tuple3<Double, Integer, Double>> iterable,
+                Collector<Tuple4<Double, Integer, Map<Double, Double>, Integer>> collector) {
+            List<Tuple3<Double, Integer, Double>> list = new ArrayList<>();
+            iterable.iterator().forEachRemaining(list::add);
+
+            Map<Tuple2<Double, Integer>, Map<Double, Double>> accMap = new HashMap<>();
+            Map<Tuple2<Double, Integer>, Integer> numMap = new HashMap<>();
+            for (Tuple3<Double, Integer, Double> value : list) {
+                Tuple2<Double, Integer> key = new Tuple2<>(value.f0, value.f1);
+                Map<Double, Double> acc = accMap.computeIfAbsent(key, x -> new HashMap<>());
+                acc.put(value.f2, acc.getOrDefault(value.f2, 0.) + 1.0);
+                numMap.put(key, numMap.getOrDefault(key, 0) + 1);
+            }
+            for (Map.Entry<Tuple2<Double, Integer>, Map<Double, Double>> entry :
+                    accMap.entrySet()) {
+                collector.collect(
+                        new Tuple4<>(
+                                entry.getKey().f0,
+                                entry.getKey().f1,
+                                entry.getValue(),
+                                numMap.get(entry.getKey())));
+            }
+        }
+    }
+
+    /**
+     * Function that aggregates maps under the same label into arrays.
+     *
+     * <p>Length of the generated array equals to the number of feature columns.
+     *
+     * <p>Input records are tuples with the following fields in order:
+     *
+     * <ul>
+     *   <li>label value
+     *   <li>feature column index
+     *   <li>map of (feature value, weight)
+     *   <li>number of records
+     * </ul>
+     *
+     * <p>Output records are tuples with the following fields in order:
+     *
+     * <ul>
+     *   <li>label value
+     *   <li>number of records
+     *   <li>array of featureValue-weight maps of each feature
+     * </ul>
+     */
+    private static class AggregateIntoArrayFunction
+            implements MapPartitionFunction<
+                    Tuple4<Double, Integer, Map<Double, Double>, Integer>,
+                    Tuple3<Double, Integer, Map<Double, Double>[]>> {
+
+        @Override
+        public void mapPartition(
+                Iterable<Tuple4<Double, Integer, Map<Double, Double>, Integer>> iterable,
+                Collector<Tuple3<Double, Integer, Map<Double, Double>[]>> collector) {
+            Map<Double, List<Tuple4<Double, Integer, Map<Double, Double>, Integer>>> map =
+                    new HashMap<>();
+            for (Tuple4<Double, Integer, Map<Double, Double>, Integer> value : iterable) {
+                map.computeIfAbsent(value.f0, x -> new ArrayList<>()).add(value);
+            }
+
+            for (List<Tuple4<Double, Integer, Map<Double, Double>, Integer>> list : map.values()) {
+                final int featureSize =
+                        list.stream().map(x -> x.f1).max(Integer::compareTo).orElse(-1) + 1;
+
+                int minDocNum =
+                        list.stream()
+                                .map(x -> x.f3)
+                                .min(Integer::compareTo)
+                                .orElse(Integer.MAX_VALUE);
+                int maxDocNum =
+                        list.stream()
+                                .map(x -> x.f3)
+                                .max(Integer::compareTo)
+                                .orElse(Integer.MIN_VALUE);
+                Preconditions.checkArgument(
+                        minDocNum == maxDocNum, "Feature vectors should be of equal length.");
+
+                Map<Double, Integer> numMap = new HashMap<>();
+                Map<Double, Map<Double, Double>[]> featureWeightMap = new HashMap<>();
+                for (Tuple4<Double, Integer, Map<Double, Double>, Integer> value : list) {
+                    Map<Double, Double>[] featureWeight =
+                            featureWeightMap.computeIfAbsent(
+                                    value.f0, x -> new HashMap[featureSize]);
+                    numMap.put(value.f0, value.f3);
+                    featureWeight[value.f1] = value.f2;
+                }
+
+                for (double key : featureWeightMap.keySet()) {
+                    collector.collect(
+                            new Tuple3<>(key, numMap.get(key), featureWeightMap.get(key)));
+                }
+            }
+        }
+    }
+
+    /** Function to generate Naive Bayes model data. */
+    private static class GenerateModelFunction
+            implements MapPartitionFunction<
+                    Tuple3<Double, Integer, Map<Double, Double>[]>, NaiveBayesModelData> {
+        private final double smoothing;
+
+        private GenerateModelFunction(double smoothing) {
+            this.smoothing = smoothing;
+        }
+
+        @Override
+        public void mapPartition(
+                Iterable<Tuple3<Double, Integer, Map<Double, Double>[]>> iterable,
+                Collector<NaiveBayesModelData> collector) {
+            ArrayList<Tuple3<Double, Integer, Map<Double, Double>[]>> list = new ArrayList<>();
+            iterable.iterator().forEachRemaining(list::add);
+            final int featureSize = list.get(0).f2.length;
+            for (Tuple3<Double, Integer, Map<Double, Double>[]> tup : list) {
+                Preconditions.checkArgument(
+                        featureSize == tup.f2.length, "Feature vectors should be of equal length.");
+            }
+
+            double[] numDocs = new double[featureSize];
+            HashSet<Double>[] categoryNumbers = new HashSet[featureSize];
+            for (int i = 0; i < featureSize; i++) {
+                categoryNumbers[i] = new HashSet<>();
+            }
+            for (Tuple3<Double, Integer, Map<Double, Double>[]> tup : list) {
+                for (int i = 0; i < featureSize; i++) {
+                    numDocs[i] += tup.f1;
+                    categoryNumbers[i].addAll(tup.f2[i].keySet());
+                }
+            }
+
+            int[] categoryNumber = new int[featureSize];
+            double piLog = 0;
+            int numLabels = list.size();
+            for (int i = 0; i < featureSize; i++) {
+                categoryNumber[i] = categoryNumbers[i].size();
+                piLog += numDocs[i];
+            }
+            piLog = Math.log(piLog + numLabels * smoothing);
+
+            Map<Double, Double>[][] theta = new HashMap[numLabels][featureSize];
+            double[] piArray = new double[numLabels];
+            double[] labels = new double[numLabels];
+
+            // Consider smoothing.
+            for (int i = 0; i < numLabels; i++) {
+                Map<Double, Double>[] param = list.get(i).f2;
+                for (int j = 0; j < featureSize; j++) {
+                    Map<Double, Double> squareData = new HashMap<>();
+                    double thetaLog =
+                            Math.log(list.get(i).f1 * 1.0 + smoothing * categoryNumber[j]);
+                    for (Double cate : categoryNumbers[j]) {
+                        double value = param[j].getOrDefault(cate, 0.0);
+                        squareData.put(cate, Math.log(value + smoothing) - thetaLog);
+                    }
+                    theta[i][j] = squareData;
+                }
+
+                labels[i] = list.get(i).f0;
+                double weightSum = list.get(i).f1 * featureSize;
+                piArray[i] = Math.log(weightSum + smoothing) - piLog;
+            }
+
+            NaiveBayesModelData modelData = new NaiveBayesModelData(theta, piArray, labels);
+            collector.collect(modelData);
+        }
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
@@ -1,0 +1,187 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.classification.naivebayes;
+
+import org.apache.flink.api.common.functions.RichMapFunction;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Model;
+import org.apache.flink.ml.common.broadcast.BroadcastUtils;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.BLAS;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/** A Model which classifies data using the model data computed by {@link NaiveBayes}. */
+public class NaiveBayesModel
+        implements Model<NaiveBayesModel>, NaiveBayesModelParams<NaiveBayesModel> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private Table modelDataTable;
+
+    public NaiveBayesModel() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+
+        final String predictionCol = getPredictionCol();
+        final String featuresCol = getFeaturesCol();
+        final String broadcastModelKey = "NaiveBayesModelStream";
+
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(), TypeInformation.of(Double.class)),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), predictionCol));
+
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) modelDataTable).getTableEnvironment();
+        DataStream<NaiveBayesModelData> modelDataStream =
+                NaiveBayesModelData.getModelDataStream(modelDataTable);
+        DataStream<Row> input = tEnv.toDataStream(inputs[0]);
+
+        Map<String, DataStream<?>> broadcastMap = new HashMap<>();
+        broadcastMap.put(broadcastModelKey, modelDataStream);
+
+        Function<List<DataStream<?>>, DataStream<Row>> function =
+                dataStreams -> {
+                    DataStream stream = dataStreams.get(0);
+                    return stream.map(
+                            new PredictLabelFunction(featuresCol, broadcastModelKey),
+                            outputTypeInfo);
+                };
+        DataStream<Row> output =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(input),
+                        Collections.singletonMap(broadcastModelKey, modelDataStream),
+                        function);
+
+        Table outputTable = tEnv.fromDataStream(output);
+
+        return new Table[] {outputTable};
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+        ReadWriteUtils.saveModelData(
+                NaiveBayesModelData.getModelDataStream(modelDataTable),
+                path,
+                new NaiveBayesModelData.ModelDataEncoder());
+    }
+
+    public static NaiveBayesModel load(StreamExecutionEnvironment env, String path)
+            throws IOException {
+        NaiveBayesModel model = ReadWriteUtils.loadStageParam(path);
+        DataStream<NaiveBayesModelData> modelData =
+                ReadWriteUtils.loadModelData(
+                        env, path, new NaiveBayesModelData.ModelDataStreamFormat());
+        return model.setModelData(NaiveBayesModelData.getModelDataTable(modelData));
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    @Override
+    public NaiveBayesModel setModelData(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        modelDataTable = inputs[0];
+        return this;
+    }
+
+    @Override
+    public Table[] getModelData() {
+        return new Table[] {modelDataTable};
+    }
+
+    private static class PredictLabelFunction extends RichMapFunction<Row, Row> {
+        private final String featuresCol;
+        private final String broadcastModelKey;
+        private NaiveBayesModelData modelData = null;
+
+        public PredictLabelFunction(String featuresCol, String broadcastModelKey) {
+            this.featuresCol = featuresCol;
+            this.broadcastModelKey = broadcastModelKey;
+        }
+
+        @Override
+        public Row map(Row row) {
+            if (modelData == null) {
+                modelData =
+                        (NaiveBayesModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastModelKey).get(0);
+            }
+            Vector vector = (Vector) row.getField(featuresCol);
+            double label = findMaxProbLabel(calculateProb(modelData, vector), modelData.labels);
+            return Row.join(row, Row.of(label));
+        }
+    }
+
+    private static double findMaxProbLabel(DenseVector prob, Vector label) {
+        double result = 0.;
+        int probSize = prob.size();
+        double maxVal = Double.NEGATIVE_INFINITY;
+        for (int i = 0; i < probSize; ++i) {
+            if (maxVal < prob.values[i]) {
+                maxVal = prob.values[i];
+                result = label.get(i);
+            }
+        }
+        Preconditions.checkArgument(maxVal > Double.NEGATIVE_INFINITY);
+        return result;
+    }
+
+    /** Calculate probability of the input data. */
+    private static DenseVector calculateProb(NaiveBayesModelData modelData, Vector data) {
+        int labelSize = modelData.labels.size();
+        DenseVector probs = new DenseVector(new double[labelSize]);
+        for (int i = 0; i < labelSize; i++) {
+            Map<Double, Double>[] labelData = modelData.theta[i];
+            for (int j = 0; j < data.size(); j++) {
+                probs.values[i] += labelData[j].get(data.get(j));
+            }
+        }
+        BLAS.axpy(1, modelData.piArray, probs);
+        return probs;
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.classification.naivebayes;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.serialization.Encoder;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeutils.base.DoubleSerializer;
+import org.apache.flink.api.common.typeutils.base.MapSerializer;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.connector.file.src.reader.SimpleStreamFormat;
+import org.apache.flink.core.fs.FSDataInputStream;
+import org.apache.flink.core.memory.DataInputViewStreamWrapper;
+import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.ml.linalg.typeinfo.DenseVectorSerializer;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The model data of {@link NaiveBayesModel}.
+ *
+ * <p>This class also provides methods to convert model data between Table and Datastream, and
+ * classes to save/load model data.
+ */
+public class NaiveBayesModelData {
+    /**
+     * Log of class conditional probabilities, whose dimension is C (number of classes) by D (number
+     * of features).
+     */
+    public final Map<Double, Double>[][] theta;
+
+    /** Log of class priors, whose dimension is C (number of classes). */
+    public final DenseVector piArray;
+
+    /** Value of labels. */
+    public final DenseVector labels;
+
+    public NaiveBayesModelData(Map<Double, Double>[][] theta, double[] piArray, double[] labels) {
+        this(theta, Vectors.dense(piArray), Vectors.dense(labels));
+    }
+
+    public NaiveBayesModelData(
+            Map<Double, Double>[][] theta, DenseVector piArray, DenseVector labels) {
+        this.theta = theta;
+        this.piArray = piArray;
+        this.labels = labels;
+    }
+
+    /** Converts the provided modelData Datastream into corresponding Table. */
+    public static Table getModelDataTable(DataStream<NaiveBayesModelData> stream) {
+        StreamTableEnvironment tEnv =
+                StreamTableEnvironment.create(stream.getExecutionEnvironment());
+        return tEnv.fromDataStream(stream);
+    }
+
+    /** Converts the provided modelData Table into corresponding DataStream. */
+    public static DataStream<NaiveBayesModelData> getModelDataStream(Table table) {
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) table).getTableEnvironment();
+        return tEnv.toDataStream(table)
+                .map(
+                        (MapFunction<Row, NaiveBayesModelData>)
+                                row -> (NaiveBayesModelData) row.getField("f0"));
+    }
+
+    /** Encoder for the {@link NaiveBayesModelData}. */
+    public static class ModelDataEncoder implements Encoder<NaiveBayesModelData> {
+        @Override
+        public void encode(NaiveBayesModelData modelData, OutputStream outputStream)
+                throws IOException {
+            DataOutputViewStreamWrapper output = new DataOutputViewStreamWrapper(outputStream);
+
+            DenseVectorSerializer denseVectorSerializer = new DenseVectorSerializer();
+            MapSerializer<Double, Double> mapSerializer =
+                    new MapSerializer<>(new DoubleSerializer(), new DoubleSerializer());
+
+            denseVectorSerializer.serialize(modelData.labels, output);
+
+            denseVectorSerializer.serialize(modelData.piArray, output);
+
+            output.writeInt(modelData.theta.length);
+            output.writeInt(modelData.theta[0].length);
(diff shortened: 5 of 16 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+            if (label == null) {
+                return;
+            }
+
+            for (int i = 0; i < featureSize; i++) {
+                Object feature = row.getField(featureColNames[i]);
+                if (feature == null) {
+                    continue;
+                }
+                collector.collect(new Tuple4<>(label, i, feature, 1.0));
```
zhipeng93, line 143: should we support weighted naive bayes or just remove 1.0 here?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/datastream/EndOfStreamWindows.java
```diff
@@ -0,0 +1,82 @@
...
+    private EndOfStreamWindows() {}
+
+    public static EndOfStreamWindows get() {
+        return INSTANCE;
+    }
+
+    @Override
+    public Collection<TimeWindow> assignWindows(
+            Object element, long timestamp, WindowAssignerContext context) {
+        return Collections.singletonList(new TimeWindow(Long.MIN_VALUE, Long.MAX_VALUE));
```
zhipeng93, line 60: why do you create a new TimeWindow on each call?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+import java.util.HashSet;
+import java.util.Map;
+
+/**
+ * Naive Bayes classifier is a simple probability classification algorithm using
+ * Bayes theorem based on independent assumption. It is an independent feature model.
+ * The input feature can be continual or categorical.
+ */
+public class NaiveBayes implements Estimator<NaiveBayes, NaiveBayesModel>,
+        NaiveBayesParams<NaiveBayes> {
```
zhipeng93, line 55: Shall we use different param classes for `NaiveBayes` and the `NaiveBayesModel`?

it may be confusing to users when call `new NaiveBayesModel().setSmoothing(...)` since `smoothing` is useless during the prediction.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,99 @@
...
+                    .build();
+
+    public NaiveBayesModelData(String[] featureNames, Map<Object, Double>[][] theta, double[] piArray, Object[] label) {
+        this.featureNames = featureNames;
+        this.theta = theta;
+        this.piArray = piArray;
+        this.label = label;
+    }
+
+    /** Encoder for the KMeans model data. */
```
zhipeng93, line 60: update the comment here.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,99 @@
...
+        @Override
+        public void encode(NaiveBayesModelData modelData, OutputStream outputStream) {
+            Kryo kryo = new Kryo();
+            Output output = new Output(outputStream);
+            kryo.writeObject(output, modelData);
+            output.flush();
+        }
+    }
+
+    /** Decoder for the KMeans model data. */
```
zhipeng93, line 71: update the comment here.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+                .keyBy((KeySelector<Tuple4<Object, Integer, Object, Double>, Object>) value -> new Tuple3<>(value.f0, value.f1, value.f2))
+                .window(EndOfStreamWindows.get())
+                .reduce((ReduceFunction<Tuple4<Object, Integer, Object, Double>>) (t0, t1) -> {t0.f3 += t1.f3; return t0; })
+                .keyBy((KeySelector<Tuple4<Object, Integer, Object, Double>, Object>) value -> new Tuple2<>(value.f0, value.f1))
+                .window(EndOfStreamWindows.get())
+                .aggregate(new ValueMapFunction())
+                .keyBy((KeySelector<Tuple4<Object, Integer, Map<Object, Double>, Double>, Object>) value -> value.f0)
+                .window(EndOfStreamWindows.get())
+                .aggregate(new MapArrayFunction(featureColNames.length))
+                .windowAll(EndOfStreamWindows.get())
```
zhipeng93, line 89: When we use `windowAll`, we are actually changing paralllelism to one and using one task to process all the elements. 

Are there any efficiency issues?

zhipeng93 (commented):
Thanks for the PR, Yunfeng!

Overall it looks good to me. I left some minor comments here.  

By the way, could you also reformat the code by `mvn spotless:apply`

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
yunfengzhou-hub: We may be able to support weighted training data in Naive Bayes in a later version, but in the current version I believe we only need to achieve the basic functionality of this algorithm.

Although we do not need weighted function here in this version, we still need to add this `1.0` to every training data, because we need to sum up this value and use `weightSum` to calculate the probability of each feature value and each label.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/datastream/EndOfStreamWindows.java
yunfengzhou-hub: I just confirmed with Lin Dong as this is his code. He agrees that we can create a global static variable here and reuse it every time we call `assignWindows()`. He will make this change in next version of his code.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
yunfengzhou-hub: There could be efficiency issues, but I think this is a necessity, because we need one operator to gather all the partial changes to form a `NaiveBayesModelData` before broadcasting it to `Model` operators. Before it is possible to avoid reducing parallelism to one, we need a way to always calculate and store a model data in a separated form.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesParams.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.shared.colname.HasFeatureCols;
+import org.apache.flink.ml.param.shared.colname.HasLabelCol;
+import org.apache.flink.ml.param.shared.colname.HasPredictionCol;
+
+/**
+ * Parameters of naive bayes training process.
+ */
+public interface NaiveBayesParams<T> extends
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll add this param to NaiveBayes.

One consideration is that while spark supports setting the type of a whole Naive Bayes model, Alink allows finer granularity to define the model type for each feature column. I agree that we just need to define the type of a whole model in the foreseeable future, but if users need to define type for each column they will need to create multiple NaiveBayesModel and use some bagging algorithm to unify them into one.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+    public NaiveBayesModel fit(Table... inputs) {
+        String[] featureColNames = getFeatureCols();
+        String labelColName = getLabelCol();
+        String predictionCol = getPredictionCol();
+        double smoothing = getSmoothing();
+
+        Preconditions.checkNotNull(inputs, "input table list should not be null");
+        Preconditions.checkArgument(inputs.length == 1, "input table list should contain only one argument");
+        Preconditions.checkArgument(
+                new HashSet<>(Arrays.asList(featureColNames)).size() == featureColNames.length,
```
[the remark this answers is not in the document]
yunfengzhou-hub: I agree we can remove this check.

My consideration was that, like we want to throw NPE instead of ignoring null values or labels, duplicated feature column also faces such choice. We could explicitly throw an exception and let users handle it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
zhipeng93: Is this helpful in this case?

https://github.com/apache/flink-ml/pull/30

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
yunfengzhou-hub: OK. I have fixed it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
yunfengzhou-hub: OK. I have fixed it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
yunfengzhou-hub: Yes this PR achieves similar functionality. If this PR is adopted I think we can make Naive Bayes follow this convention.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesParams.java
```diff
@@ -0,0 +1,49 @@
...
+        HasFeatureCols<T>,
+        HasLabelCol<T>,
+        HasPredictionCol<T> {
+    Param<Double> SMOOTHING =
+            new DoubleParam(
+                    "smoothing",
+                    "the smoothing factor",
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK I'll update the description. I use "smoothing factor" because Alink describes in this way. I'll update it into Spark's convention.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    @Override
+    public NaiveBayesModel fit(Table... inputs) {
+        String[] featureColNames = getFeatureCols();
+        String labelColName = getLabelCol();
+        String predictionCol = getPredictionCol();
+        double smoothing = getSmoothing();
+
+        Preconditions.checkNotNull(inputs, "input table list should not be null");
```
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. I have removed redundant checks and added Flink's `Precondition` checks.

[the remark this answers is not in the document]
yunfengzhou-hub: Got it. I have removed the redundant check.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+        Preconditions.checkArgument(inputs.length == 1, "input table list should contain only one argument");
+        Preconditions.checkArgument(
+                new HashSet<>(Arrays.asList(featureColNames)).size() == featureColNames.length,
+                "feature columns should not duplicate");
+        Preconditions.checkNotNull(labelColName, "label column should be set");
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I have added `ParamValidators::notNull` to `HasLabelCol` and calls `getLabelCol()` once `NaiveBayes.fit()` is called in order to invoke the validator.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+        private FlattenFunction(String[] featureColNames, String labelColName) {
+            this.labelColName = labelColName;
+            this.featureColNames = featureColNames;
+            this.featureSize = featureColNames.length;
+        }
+
+        @Override
+        public void flatMap(Row row, Collector<Tuple4<Object, Integer, Object, Double>> collector) {
+            Object label = row.getField(labelColName);
+            if (label == null) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. I have added Flink's `Precondition` checks.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+        public Tuple4<Object, Integer, Map<Object, Double>, Double> merge(
+                Tuple4<Object, Integer, Map<Object, Double>, Double> acc0,
+                Tuple4<Object, Integer, Map<Object, Double>, Double> acc1) {
+            if (acc0.f0.equals(new Object())) {
+                acc0.f0 = acc1.f0;
+            }
+            if (acc0.f1 == -1) {
+                acc0.f1 = acc1.f1;
+            }
+            for (Map.Entry<Object, Double> entry: acc1.f2.entrySet()) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+        public Tuple4<Object, Integer, Map<Object, Double>, Double> getResult(
+                Tuple4<Object, Integer, Map<Object, Double>, Double> acc) {
+            return acc;
+        }
+
+        @Override
+        public Tuple4<Object, Integer, Map<Object, Double>, Double> merge(
+                Tuple4<Object, Integer, Map<Object, Double>, Double> acc0,
+                Tuple4<Object, Integer, Map<Object, Double>, Double> acc1) {
+            if (acc0.f0.equals(new Object())) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Yes, sure.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+        @Override
+        public Tuple3<Object, Double[], Map<Object, Double>[]> merge(
+                Tuple3<Object, Double[], Map<Object, Double>[]> acc0,
+                Tuple3<Object, Double[], Map<Object, Double>[]> acc1) {
+            if (acc0.f0.equals(new Object())) {
+                acc0.f0 = acc1.f0;
+            }
+            for (int i = 0; i < featureSize; i++) {
+                acc0.f1[i] += acc1.f1[i];
+                if (acc1.f2[i] != null) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: We need to merge two arrays whose elements are not initialized. Thus it is possible to contain null values in this array.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+            this.featureSize = featureColNames.length;
+        }
+
+        @Override
+        public void apply(TimeWindow timeWindow,
+                          Iterable<Tuple3<Object, Double[], Map<Object, Double>[]>> values,
+                          Collector<NaiveBayesModelData> out) {
+            double[] numDocs = new double[featureSize];
+            ArrayList <Tuple3 <Object, Double[], Map <Object, Double>[]>> modelArray = new ArrayList <>();
+            HashSet <Object>[] categoryNumbers = new HashSet[featureSize];
```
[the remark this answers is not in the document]
yunfengzhou-hub: 'll run `mvn spotless:apply` to format the code.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+     * Function to aggregate feature value and weight into map from records with the same label and feature column index.
+     */
+    private static class ValueMapFunction implements AggregateFunction<
+            Tuple4<Object, Integer, Object, Double>,
+            Tuple4<Object, Integer, Map<Object, Double>, Double>,
+            Tuple4<Object, Integer, Map<Object, Double>, Double>> {
+
+        @Override
+        public Tuple4<Object, Integer, Map<Object, Double>, Double> createAccumulator() {
+            return new Tuple4<>(new Object(), -1, new HashMap<>(), 0.);
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I have removed the field.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /**
+     * Function to convert each column into tuples of label, feature column index, feature value, weight.
+     */
+    private static class FlattenFunction implements FlatMapFunction<Row, Tuple4<Object, Integer, Object, Double>> {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Following Spark's practice, I have changed input features into `Vector` and label into `double`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
yunfengzhou-hub: Got it. I have separated the two param classes.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesParams.java
```diff
@@ -0,0 +1,49 @@
...
+                    0.0,
```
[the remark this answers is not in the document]
yunfengzhou-hub: I think either value makes sense. But as Spark chooss 1.0 let's also follow this practice.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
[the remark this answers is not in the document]
yunfengzhou-hub: In latest commit I have replaced featureCols with featuresCol, from string array to a single string. So there is no need for duplicated check here.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,297 @@
...
+                        .as("weight", "features", "label");
+        Table predictTable =
+                tEnv.fromDataStream(
+                                env.fromElements(predictData)
+                                        .assignTimestampsAndWatermarks(
+                                                WatermarkStrategy.noWatermarks()),
+                                schema)
+                        .as("weight", "features", "label");
+
+        NaiveBayes estimator = new NaiveBayes();
```
lindong28, line 242: It is preferred to follow best practice when writing test code, since it is what users will read and probably copy.

We can use `NaiveBayes estimator = new NaiveBayes().setSmoothing(smoothing)` here.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,297 @@
...
+        estimator.setSmoothing(smoothing);
+        if (featuresCol != null) estimator.setFeaturesCol(featuresCol);
```
lindong28, line 244: Can we just remove the `if` and set the param directly?

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,297 @@
...
+
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.flink.table.api.Expressions.$;
+
+public class NaiveBayesTest {
```
lindong28, line 49: Can we add test to cover `setModelData`, `getModelData` and default param values? Maybe use KMeansTest for example.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,297 @@
...
+                    Row.of(1., Vectors.dense(2, 0., 1., 1.5, 2.), 11.0),
+                    Row.of(2., Vectors.dense(3, 1.5, 1., 0.5, 3.), 10.0),
+                    Row.of(1., Vectors.dense(1, 1., 1.5, 0., 1.), 10.0),
+                    Row.of(2., Vectors.dense(4, 1., 1., 0., 1.), 10.0)
+                };
+        runAndCheck();
+    }
+
+    @Test(expected = Exception.class)
+    public void testEmptyLabel() throws Exception {
```
lindong28, line 150: It seems a bit verbose to have a dedicated test for each param. We probably don't want to do this for every parameter of every Estimator/Transformer.

And the purpose of this test seems to have already been covered by `StageTest::testValidators`. We can add unit tests there if needed.

lindong28, line 150: It seems a bit verbose to have a dedicated test for each param. We probably don't want to do this for every parameter of every Estimator/Transformer.

And the purpose of this test seems to have already been covered by `StageTest::testValidators`. We can add unit tests there if needed.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,235 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/** Naive Bayes Predictor. */
```
lindong28, line 59: Can you rephrase the comments here following KMeans or Spark's convention? It is preferred not to introduce `Predictor` as a new concept here.

lindong28 (commented):

It looks like `mvn package` fails due to stylecheck error. Could you make sure `mvn package` works? I

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,349 @@
...
+                    Row.of(1., Vectors.dense(1, 1., 1., 1., 2.), 11.0, 11.0),
+                    Row.of(1., Vectors.dense(1, 1., 0., 1., 2.), 11.0, 11.0),
+                    Row.of(1., Vectors.dense(2, 0., 1., 1., 3.), 11.0, 11.0),
+                    Row.of(1., Vectors.dense(2, 0., 1., 1.5, 2.), 11.0, 11.0),
+                    Row.of(2., Vectors.dense(3, 1.5, 1., 0.5, 3.), 10.0, 10.0),
+                    Row.of(1., Vectors.dense(1, 1., 1.5, 0., 1.), 10.0, 10.0),
+                    Row.of(2., Vectors.dense(4, 1., 1., 0., 1.), 10.0, 10.0)
+                };
+
+        featuresCol = "features";
```
lindong28, line 111: nits: It looks like this variable is only used once in this class. Does this need to be a class-level variable? It is preferred to use local variable instead of class-level variable if possible.

Same for other variables.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,349 @@
...
+    private static Map<Object, Integer> getFrequencyMap(Row[] rows) {
+        Map<Object, Integer> map = new HashMap<>();
+        for (Row row : rows) {
+            List<Object> list = toList(row);
+            map.put(list, map.getOrDefault(list, 0) + 1);
+        }
+        return map;
+    }
+
+    private static List<Object> toList(Row row) {
```
lindong28, line 342: Would it be better to add a `List<Object> getFieldValues(Row row)` in `TableUtils`?

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,349 @@
...
+    private Row[] trainData;
+    private Row[] predictData;
+    private Row[] expectedOutput;
+    private String featuresCol;
+    private String labelCol;
+    private String predictCol;
+    private String modelType;
+    private double smoothing;
+    private boolean isSaveLoad;
+    private String errorMessage;
```
lindong28, line 66: It is very rare to pass `errorMessage` via through class-level variable. Can you try to pass it to runAndCheck directly?

For example, class-level variables in KMeansTest are only written in `before()`.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,349 @@
...
+        NaiveBayesModel modelA = getModel();
+
+        Table modelData = modelA.getModelData()[0];
+        NaiveBayesModel modelB = new NaiveBayesModel().setModelData(modelData);
+        ReadWriteUtils.updateExistingParams(modelB, modelA.getParamMap());
+
+        checkResult(modelB);
+    }
+
+    private void runAndCheck() throws Exception {
```
lindong28, line 268: It is probably preferred to setup a best practice of writing unit tests for algorithms.

Any chance we can re-use the test pattern (e.g. `executeAndCollect`) used in `KMeansTest`?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,330 @@
...
+        implements Estimator<NaiveBayes, NaiveBayesModel>, NaiveBayesParams<NaiveBayes> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public NaiveBayes() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public NaiveBayesModel fit(Table... inputs) {
+        final String featuresCol = getFeaturesCol();
```
lindong28, line 66: I realized that it is probably better to setup the best practice of always using `Preconditions.checkArgument(...)` to verify the number of inputs is expected, for every public APIs such as fit/transform/setModelData.

Could you help add this check for both NaiveBayes and KMeans?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,226 @@
...
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        final String predictionCol = getPredictionCol();
+        final String featuresCol = getFeaturesCol();
+
+        List<String> colNames = new ArrayList<>(inputs[0].getResolvedSchema().getColumnNames());
+        colNames.add(predictionCol);
+
+        List<TypeInformation<?>> colTypes =
```
lindong28, line 79: Would it be simpler and more consistent to use the approach in `KMeansModel`?

KMeansModel uses something like this:
```
        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
        RowTypeInfo outputTypeInfo =
                new RowTypeInfo(
                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), Types.INT),
                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), predictionCol));
```

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,226 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/** A Model which classifies data using the model data computed by {@link NaiveBayes}. */
+public class NaiveBayesModel
+        implements Model<NaiveBayesModel>, NaiveBayesModelParams<NaiveBayesModel> {
+    private static final long serialVersionUID = -4673084154965905629L;
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private Table modelTable;
+    private static final String broadcastModelKey = "NaiveBayesModelStream";
```
lindong28, line 65: `broadcastModelKey` is an implementation detail and it is only used in `NaiveBayesPredictOp`. Can this be passed to `NaiveBayesPredictOp` directly?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,226 @@
...
+                                new Path(dataPath), new NaiveBayesModelData.ModelDataEncoder())
+                        .withRollingPolicy(OnCheckpointRollingPolicy.build())
+                        .withBucketAssigner(new BasePathBucketAssigner<>())
+                        .build();
+        NaiveBayesModelData.toDataStream(tEnv, modelTable).sinkTo(sink);
+
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static NaiveBayesModel load(String path) throws IOException {
```
lindong28, line 135: Could you follow the example of KMeans and the Java doc in `Stage.java` to pass env as argument?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,226 @@
...
+        modelTable = inputs[0];
+        return this;
+    }
+
+    @Override
+    public Table[] getModelData() {
+        return new Table[] {modelTable};
+    }
+
+    private static class NaiveBayesPredictOp
```
lindong28, line 168: It is probably not necessary to have the algorithm name in this private static class. And it is preferred to make this name more meaningful.

How about changing this to something like `PredictLabelOperator`?

Same for other private classes in `NaiveBayes` and `NaiveBayesModel`.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
yunfengzhou-hub: OK. I'll add this check.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
yunfengzhou-hub: `broadcastModelKey` is also used when creating broadcastMap, but it's true that this variable is only used in one public method and its private methods. I'll make `broadcastModelKey` a local variable of `transform()`.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
yunfengzhou-hub: OK. I'll change it.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
yunfengzhou-hub: classes like this has been replaced by existing tools like `CollectionUtils` in `org.apache.commons`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private static class PredictLabelOperator
+            extends AbstractUdfStreamOperator<Row, PredictLabelFunction>
+            implements OneInputStreamOperator<Row, Row> {
+        public PredictLabelOperator(PredictLabelFunction userFunction) {
+            super(userFunction);
+        }
+
+        @Override
+        public void processElement(StreamRecord<Row> streamRecord) {
```
lindong28, line 171: Would it be simpler to remove `PredictLabelFunction` and move its logic into `PredictLabelOperator::processElement`?

lindong28 (commented):
Thanks for the update @yunfengzhou-hub. I left some minor comments regarding the code readability.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/KMeansTest.java
```diff
@@ -67,6 +69,18 @@
                     Vectors.dense(9.6, 0.0));
     private StreamExecutionEnvironment env;
     private StreamTableEnvironment tEnv;
+    private static List<Set<DenseVector>> expectedGroup =
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/KMeansTest.java
```diff
@@ -83,43 +97,19 @@ public void before() {
...
-                        clusterIdByPoints.get(DATA.get(group.get(i))));
+    private static List<Set<DenseVector>> executeAndCollect(
+            Table output, String featureCol, String predictionCol) {
+        Map<Integer, Set<DenseVector>> map = new HashMap<>();
+        for (CloseableIterator<Row> it = output.execute().collect(); it.hasNext(); ) {
+            Row row = it.next();
+            DenseVector vector = (DenseVector) row.getField(featureCol);
+            int predict = (Integer) row.getField(predictionCol);
+            if (!map.containsKey(predict)) {
+                map.put(predict, new HashSet<>());
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/KMeansTest.java
```diff
@@ -83,43 +97,19 @@ public void before() {
...
-    }
-
-    private static void verifyClusteringResult(
-            Map<DenseVector, Integer> clusterIdByPoints, List<List<Integer>> groups) {
-        for (List<Integer> group : groups) {
-            for (int i = 1; i < group.size(); i++) {
-                assertEquals(
-                        clusterIdByPoints.get(DATA.get(group.get(0))),
-                        clusterIdByPoints.get(DATA.get(group.get(i))));
+    private static List<Set<DenseVector>> executeAndCollect(
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/KMeansTest.java
```diff
@@ -158,10 +148,11 @@ public void testFeaturePredictionParam() throws Exception {
         assertEquals(
                 Arrays.asList("test_feature", "test_prediction"),
                 output.getResolvedSchema().getColumnNames());
-        Map<DenseVector, Integer> clusterIdByPoints =
-                executeAndCollect(output, kmeans.getFeaturesCol(), kmeans.getPredictionCol());
-        verifyClusteringResult(
-                clusterIdByPoints, Arrays.asList(Arrays.asList(0, 1, 2), Arrays.asList(3, 4, 5)));
+        assertTrue(
+                CollectionUtils.isEqualCollection(
+                        executeAndCollect(
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,291 @@
...
+        assertEquals("prediction", model.getPredictionCol());
+
+        model.setFeaturesCol("test_feature").setPredictionCol("test_prediction");
+
+        assertEquals("test_feature", model.getFeaturesCol());
+        assertEquals("test_prediction", model.getPredictionCol());
+    }
+
+    @Test
+    public void testNaiveBayes() throws Exception {
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+/** Tests {@link NaiveBayes} and {@link NaiveBayesModel}. */
+public class NaiveBayesTest {
+    private StreamExecutionEnvironment env;
+    private StreamTableEnvironment tEnv;
+    private Schema schema;
+    private List<Row> trainData;
+    private List<Row> predictData;
+    private List<Row> expectedOutput;
+    private boolean isSaveLoad;
```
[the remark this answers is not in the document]
yunfengzhou-hub: Yes you are right. I will make this change in `NaiveBayesTest` like what we have done in `KmeansTest`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+                                new MapFunction<Row, Tuple2<Vector, Double>>() {
+                                    @Override
+                                    public Tuple2<Vector, Double> map(Row row) throws Exception {
+                                        return new Tuple2<>(
+                                                (Vector) row.getField(featuresCol),
+                                                (Double) row.getField(labelCol));
+                                    }
+                                });
+
+        DataStream<NaiveBayesModelData> naiveBayesModel =
```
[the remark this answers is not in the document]
yunfengzhou-hub: Yes, `modeData` should be more clear. I'll make this change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+                input.flatMap(new FlattenFunction())
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it and add clearer comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+                Tuple3<Double, Integer, Map<Double, Double>> acc0,
+                Tuple3<Double, Integer, Map<Double, Double>> acc1) {
+            Preconditions.checkArgument(acc0.f1 != -1);
+            acc0.f2.putAll(acc1.f2);
+            return acc0;
+        }
+    }
+
+    /** Function to aggregate maps under the same label into arrays. array len = featureSize. */
+    private static class MapArrayFunction
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll do it.

flink-ml-lib/src/test/java/org/apache/flink/ml/util/StageTestUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+/** Utility methods for testing stages. */
+public class StageTestUtils {
+    /**
+     * Saves a stage to filesystem and reloads it with the static load() method a stage must
+     * implement.
+     */
+    public static <T extends Stage<T>> T saveAndReload(StreamExecutionEnvironment env, T stage)
+            throws Exception {
+        String tempDir = Files.createTempDirectory("").toString();
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll do it.

The common practice in Flink tests is to initialize the `TemporaryFolder` in each test class, and calls for initialization in `@Before` methods. Since the usage of `TemporaryFolder` in our case is merely in static utility functions. I think we can just put it as a static member of that class and initializes in static context.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasLabelCol.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.StringParam;
+import org.apache.flink.ml.param.WithParams;
+
+/**
+ * Param of the name of the label column in the input table.
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll fix it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,116 @@
...
+        this(null, null, null);
+    }
+
+    public NaiveBayesModelData(Map<Double, Double>[][] theta, double[] piArray, double[] labels) {
+        this.theta = theta;
+        this.piArray = piArray;
+        this.labels = labels;
+    }
+
+    public static Table fromDataStream(
```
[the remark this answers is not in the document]
yunfengzhou-hub: Sure. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+        return paramMap;
+    }
+
+    /**
+     * Function to convert each column into tuples of label, feature column index, feature value,
+     * weight.
+     */
+    private static class FlattenFunction
+            implements FlatMapFunction<
+                    Tuple2<Vector, Double>, Tuple4<Double, Integer, Double, Double>> {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Yes, that's right. I'll make the change.

I'll make it that users can still choose to input labels that are integer or double, but finally they will be converted to integer.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,333 @@
...
+            acc.f0 = value.f0;
+            acc.f1 = value.f1;
+            acc.f2.put(value.f2, value.f3);
+            return acc;
+        }
+
+        @Override
+        public Tuple4<Double, Integer, Map<Double, Double>, Double> getResult(
+                Tuple3<Double, Integer, Map<Double, Double>> acc) {
+            double weightSum = acc.f2.values().stream().mapToDouble(f -> f).sum();
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
yunfengzhou-hub: I'm afraid that we cannot achieve this by now. The API `withBroadcast` requires that I must pass to it a child class of `AbstractUdfStreamOperator`. A child class of `AbstractStreamOperator` would throw exception at runtime.

zhipeng93: Actually you can make the change if you use the RichMapFunction.
But I think making it a separate function is more clear to me.

zhipeng93: Actually you can make the change if you use the RichMapFunction.
But I think making it a separate function is more clear to me.

flink-ml-api/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.flink.ml.linalg;
+
+import org.apache.flink.util.Preconditions;
+
+/** A utility class that provides BLAS routines over matrices and vectors. */
+public class BLAS {
+    /**
+     * For level-2 and level-3 routines, we use the native BLAS.
+     *
```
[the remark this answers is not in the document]
yunfengzhou-hub: Comments have been updated.

I think that BLAS should be placed in the same package as other linear algorithm classes like `Vector`. Given that `Vector` currently lies in `flink-ml-api`, I also placed `BLAS` there. We can discuss offline if we need to change the placement of such packages.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
zhipeng93: > 

Ah, I see. 
Yes, the broadcast variables can only be accessed by `RichFunction#getRuntimeContext#getBroadcastVariable` now. 

A possible solution to put processing logic in `PredictLabelOperator::processElement` is to add a AbstractFunction but only use it for accessing broadcast variables. An examples can be found in `org.apache.flink.ml.common.broadcast.operator.TestOneInputOp`

yunfengzhou-hub: You are right. This method is feasible. I have make this change and removed `PredictLabelFunction`.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,327 @@
...
+        assertEquals("prediction", model.getPredictionCol());
+
+        model.setFeaturesCol("test_feature").setPredictionCol("test_prediction");
+
+        assertEquals("test_feature", model.getFeaturesCol());
+        assertEquals("test_prediction", model.getPredictionCol());
+    }
+
+    @Test
+    public void testFitAndPredict() throws Exception {
```
lindong28, line 182: Could you help update this test to use non-default column names for features, labels and predictions? I hope to do this test for every algorithm to make sure the user-specified columns names are enforced.

This could be similar to `KMeansTest::testFeaturePredictionParam`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,377 @@
...
+     * Function that aggregates maps under the same label into arrays.
+     *
+     * <p>Length of the generated array equals to the number of feature columns.
+     *
+     * <p>Input records are tuples with the following fields in order:
+     *
+     * <ul>
+     *   <li>label value
+     *   <li>feature column index
+     *   <li>map of (feature value, weight)
```
lindong28, line 236: It seems that the 4th field of the input is missing?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+/** A Model which classifies data using the model data computed by {@link NaiveBayes}. */
+public class NaiveBayesModel
+        implements Model<NaiveBayesModel>, NaiveBayesModelParams<NaiveBayesModel> {
+    private static final long serialVersionUID = -4673084154965905629L;
```
lindong28, line 62: Do you know what could go wrong if we don't specify `serialVersionUID` for `NaiveBayesModel` and `NaiveBayesModelData`? Spark does not seem to explicitly specify UID for its Estimator/Model subclasses.

If there is no known issue yet, maybe we should remove this UID for simplicity.

If there is known impact, we will need to add this UID for `NaiveBayes`, `KMeans` and `KMeansModel` etc.

lindong28 (commented):
Thanks for the update @yunfengzhou-hub. Left only some minor comments below.

flink-ml-api/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.flink.util.Preconditions;
+
+/** A utility class that provides BLAS routines over matrices and vectors. */
+public class BLAS {
+    /** For level-1 routines, we use Java implementation. */
+    private static final dev.ludovic.netlib.BLAS F2J_BLAS =
+            dev.ludovic.netlib.JavaBLAS.getInstance();
+
+    /** y += a * x . */
+    public static void axpy(double a, double[] x, double[] y) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: In `Vector` we defined a `double[] toArray();` method, so I believe that BLAS can work without direct support for `Vector` subclasses.

flink-ml-lib/src/test/java/org/apache/flink/ml/util/StageTestUtils.java
```diff
@@ -0,0 +1,62 @@
...
+    static {
+        try {
+            tempFolder.create();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * Saves a stage to filesystem and reloads it with the static load() method a stage must
```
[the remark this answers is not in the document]
yunfengzhou-hub: Yes it's reasonable. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
yunfengzhou-hub: Yes. I'll fix it.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
```diff
@@ -0,0 +1,327 @@
...
+        NaiveBayesModel model = estimator.fit(trainTable);
+        Table outputTable = model.transform(predictTable)[0];
+        Map<Vector, Integer> actualOutput =
+                executeAndCollect(outputTable, model.getFeaturesCol(), model.getPredictionCol());
+        assertEquals(expectedOutput, actualOutput);
+    }
+
+    @Test(expected = Exception.class)
+    public void testPredictUnseenFeature() throws Exception {
```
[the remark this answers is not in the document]
yunfengzhou-hub: I think `testPredictUnseenFeature` should fail naturally while the program cannot find corresponding key of the feature. Adding a check and set detailed error message might have a bit influence on the performance of this algorithm. Other tests that expect an exception needs to check the message and I'll make this change on those tests.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
yunfengzhou-hub: OK. I'll make the change.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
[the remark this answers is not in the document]
lindong28: Sounds good.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/NaiveBayesTest.java
lindong28: Sound good.

lindong28:
Thanks for the update! LGTM. @gaoyunhaii could you help review this PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,148 @@
...
+
+    public NaiveBayesModelData(Map<Double, Double>[][] theta, double[] piArray, int[] labels) {
+        this.theta = theta;
+        this.piArray = piArray;
+        this.labels = labels;
+    }
+
+    /** Converts the provided modelData Datastream into corresponding Table. */
+    public static Table getModelDataTable(
+            StreamTableEnvironment tEnv, DataStream<NaiveBayesModelData> stream) {
```
zhipeng93, line 60: Can we get `tEnv` from the input data stream rather than passing it as a param?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,148 @@
...
+        return tEnv.fromDataStream(stream);
+    }
+
+    /** Converts the provided modelData Table into corresponding DataStream. */
+    public static DataStream<NaiveBayesModelData> getModelDataStream(
+            StreamTableEnvironment tEnv, Table table) {
+        return tEnv.toDataStream(table)
```
zhipeng93, line 67: same as above.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
yunfengzhou-hub: Yes. it makes sense. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,213 @@
...
+
+        Table outputTable = tEnv.fromDataStream(output);
+
+        return new Table[] {outputTable};
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        String dataPath = ReadWriteUtils.getDataPath(path);
+        FileSink<NaiveBayesModelData> sink =
```
zhipeng93, line 115: nits: can you checkout `LogisticRegressionModelData#save/load`?

I am not sure which one is better, but I think we should follow same programming pattern.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,151 @@
...
+                                row -> (NaiveBayesModelData) row.getField("f0"));
+    }
+
+    /** Encoder for the {@link NaiveBayesModelData}. */
+    public static class ModelDataEncoder implements Encoder<NaiveBayesModelData> {
+        @Override
+        public void encode(NaiveBayesModelData modelData, OutputStream outputStream) {
+            Output output = new Output(outputStream);
+
+            output.writeInt(modelData.labels.length);
```
zhipeng93, line 82: nits: can you use Flink's serializer? Similar as` LogisticRegressionModelData#ModelDataEncoder`

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,213 @@
...
+        public PredictLabelOperator(String featuresCol, String broadcastModelKey) {
+            super(new AbstractRichFunction() {});
+            this.featuresCol = featuresCol;
+            this.broadcastModelKey = broadcastModelKey;
+        }
+
+        @Override
+        public void processElement(StreamRecord<Row> streamRecord) {
+            Row inputRow = streamRecord.getValue();
+            NaiveBayesModelData modelData =
```
zhipeng93, line 173: nits: can you cache the `modelData` instead of accessing repeatly from the broadcast context?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The model data of {@link NaiveBayesModel}.
+ *
+ * <p>This class also provides methods to convert model data between Table and Datastream, and
+ * classes to save/load model data.
+ */
+public class NaiveBayesModelData implements Serializable {
```
zhipeng93, line 48: can we remove `implements Serializable` here?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
```diff
@@ -0,0 +1,213 @@
...
+                dataStreams -> {
+                    DataStream stream = dataStreams.get(0);
+                    return stream.transform(
+                            this.getClass().getSimpleName(),
+                            outputTypeInfo,
+                            new PredictLabelOperator(featuresCol, broadcastModelKey));
+                };
+        DataStream<Row> output =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(input), broadcastMap, function);
```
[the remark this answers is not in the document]
yunfengzhou-hub: Yes, that would be a better choice. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
yunfengzhou-hub: OK. I'll make the change.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
yunfengzhou-hub: OK. I'll do it.

flink-ml-core/pom.xml
```diff
@@ -65,5 +65,12 @@ under the License.
       <artifactId>flink-shaded-jackson</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>dev.ludovic.netlib</groupId>
+      <artifactId>blas</artifactId>
+      <version>2.2.0</version>
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
yunfengzhou-hub: You are right. I'll remove the scope.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,382 @@
...
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple2<Vector, Integer>> input =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                new MapFunction<Row, Tuple2<Vector, Integer>>() {
+                                    @Override
+                                    public Tuple2<Vector, Integer> map(Row row) throws Exception {
+                                        Number number = (Number) row.getField(labelCol);
+                                        Preconditions.checkArgument(
+                                                number.intValue() == number.doubleValue());
```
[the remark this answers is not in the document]
yunfengzhou-hub: I wonder whether this check is necessary. If the number is null then the program will soon fail with NullPointerException. Adding an additional check seems not to change this behavior.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
```diff
@@ -0,0 +1,382 @@
...
+                                        return new Tuple2<>(
+                                                (Vector) row.getField(featuresCol),
+                                                number.intValue());
+                                    }
+                                });
+
+        DataStream<NaiveBayesModelData> modelData =
+                input.flatMap(new ExtractFeatureFunction())
+                        .keyBy(
+                                (KeySelector<Tuple4<Integer, Integer, Double, Double>, Object>)
```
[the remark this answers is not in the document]
yunfengzhou-hub: You are right. I'll remove the type cast.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.flink.ml.linalg;
+
+import org.apache.flink.util.Preconditions;
+
+/** A utility class that provides BLAS routines over matrices and vectors. */
+public class BLAS {
+    /** For level-1 routines, we use Java implementation. */
+    private static final dev.ludovic.netlib.BLAS F2J_BLAS =
```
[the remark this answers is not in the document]
yunfengzhou-hub: It means that the implementation is translated from Fortran to Java. You can refer to Alink or Spark for the difference between NATIVE_BLAS and F2J_BLAS.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
```diff
@@ -0,0 +1,151 @@
...
+    public final Map<Double, Double>[][] theta;
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll add these.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModel.java
yunfengzhou-hub: I agree. The implementation in LogisticRegression cannot exactly meet the needs for NaiveBayes and Kmeans, so I made some adjustment and applied this practice to these two algorithms.

Still, the current implementation cannot satisfy all possible needs of modelData, which could be more than one. We may need further discussion about how to make these utility methods cover more cases like this.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayesModelData.java
yunfengzhou-hub: Yes. I'll do it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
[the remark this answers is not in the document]
gaoyunhaii: Perhaps we could give a more meaningful exception instead of `NullPointerExceptioner`?

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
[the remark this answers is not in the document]
gaoyunhaii: https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/linalg/BLAS.scala

Sorry I do not see `F2J_BLAS` in the master spark code? Also, by googling it seems F2J_BLAS is referring to `com.github.fommil.netlib.F2jBLAS`, which seems not used here?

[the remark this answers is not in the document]
yunfengzhou-hub: Oh sorry. I also found that `F2J_BLAS` is only available in Alink and in `com.github.fommil.netlib`. I'll change the comment and naming to Spark's style.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/naivebayes/NaiveBayes.java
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll add one.

gaoyunhaii (approved):
Very thanks @yunfengzhou-hub for the updates! LGTM and +1 to merge

PR #33, opened 2021-11-14
 [FLINK-22915][FLIP-173] Updates the static load(...) method of Stage subclasses to take StreamExecutionEnvironment as parameter

lindong28:

This PR updates the static load(...) method of Stage subclasses to take StreamExecutionEnvironment as parameter. This is because the static load method will need to create DataStream or Table instances to read model data.

This PR also updates Model::setModelData(...) to return the Model instance itself. This could improve usability by allowing users to do e.g. `return new KMeansModel().setModelData(...)`.

## Brief change log

This PR mades the following changes:
- Updated static load method of all `Stage` subclasses to take StreamExecutionEnvironment as parameter.
- Updated a few methods in the `ReadWriteUtils` to take StreamExecutionEnvironment as parameter.
- Updated Model::setModelData(...) to return the Model instance itself.

## Verifying this change

The changes are validated by the existing unit tests.

## Documentation

### DIFF
--- a/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Stage.java
+++ b/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Stage.java
@@ -31,14 +31,15 @@ import java.io.Serializable;
  *
  * <p>Each stage is with parameters, and requires a public empty constructor for restoration.
  *
+ * <p>NOTE: every Stage subclass should implement a static method with signature {@code static T
+ * load(StreamExecutionEnvironment env, String path)}, where {@code T} refers to the concrete
+ * subclass. This static method should instantiate a new stage instance based on the data read from
+ * the given path.
+ *
  * @param <T> The class type of the Stage implementation itself.
  */
 @PublicEvolving
 public interface Stage<T extends Stage<T>> extends WithParams<T>, Serializable {
     /** Saves this stage to the given path. */
     void save(String path) throws IOException;
-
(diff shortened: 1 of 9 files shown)


lindong28:
@gaoyunhaii Could you review this PR?

flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Stage.java
```diff
@@ -39,6 +39,7 @@
     void save(String path) throws IOException;
 
     // NOTE: every Stage subclass should implement a static method with signature "static T
-    // load(String path)", where T refers to the concrete subclass. This static method should
-    // instantiate a new stage instance based on the data from the given path.
+    // load(StreamExecutionEnvironment env, String path)", where T refers to the concrete
```
gaoyunhaii, line 42: We might move this part of comment to the class comments? Then it would also appear in the java doc. 

Also change `static T load(StreamExecutionEnvironment env, String path)` -> `{@code static T load(StreamExecutionEnvironment env, String path)}` and `T` -> `{@code T}`

lindong28: Good point. The PR has been updated as suggested.

gaoyunhaii (approved):
Very thanks @lindong28 for the update! LGTM and +1 to merge

lindong28:
Thanks for the prompt review @gaoyunhaii!

PR #34, opened 2021-11-15
[hotfix] Fix nullpointer exception when broadcast variables are cleaned

zhipeng93:
This PR fixes the concurrency bug in the following case:

- Step1: BroadcastVariableReceiverOperator#endInput
- Step2: BroadcastVariableReceiverOperator#initializeState
- Step3: AbstractBroadcastWrapperOperator#close
- Step4: BroadcastVariableReceiverOperator#snapshot

In step4, we assumed that the broadcastVariable is still stored in `BroadcastContext`. However, it might have been removed by `AbstractBroadcastWrapperOperator` in Step3.

## Brief change log
- Fix bug in BroadcastVariableReceiverOperator.
- Add unit test in BroadcastVariableReceiverOperatorTest.

- If yes, how is the feature documented? (-)

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastContext.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastContext.java
@@ -72,7 +72,7 @@ public class BroadcastContext {
     }
 
     @VisibleForTesting
-    public static void markCacheFinished(String key) {
+    public static void notifyCacheFinished(String key) {
         BROADCAST_VARIABLES.computeIfPresent(
                 key,
                 (k, v) -> {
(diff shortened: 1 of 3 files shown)


lindong28:
Thanks for the PR @zhipeng93. Any chance you can add a unit test to verify the bug fix?

gaoyunhaii:
Very thanks @zhipeng93 for opening the PR! The fix roughly LGTM to me, perhaps we may also
1. `hotfix` usually means the PR that does not need to open issues, thus if we use `hotfix`, we might remove `[FLINK-24279]`, Or we leave `[FLINK-24279]` and remove `[hotfix]`. And we might also change the commit message accordingly.
2. Besides, I also think we might need to add a test for this case.

zhipeng93:
Hi @gaoyunhaii and @lindong28 ,

Thanks for the review. I have added a unit test. Please check out for `BroadcastVariableReceiverOperatorTest`.

gaoyunhaii:
Very thanks @zhipeng93 for the fix! LGTM and merging...

PR #41, opened 2021-12-20
[hotfix][FLIP-176] Updates onEpochWatermarkIncremented() and onIterationTerminated() to throw Exception

lindong28:

The purpose of this PR is to simplify the experience of implementing `onEpochWatermarkIncremented()` and `onIterationTerminated()`.

## Brief change log

This PR updated `onEpochWatermarkIncremented()` and `onIterationTerminated()` to throw Exception.

## Verifying this change

The changes are validated by the existing unit tests.

## Documentation

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
@@ -361,8 +361,8 @@ public class LogisticRegression
 
         @Override
         public void onEpochWatermarkIncremented(
-                int epochWatermark, Context context, Collector<double[]> collector) {
-            // TODO: let this method throws exception.
+                int epochWatermark, Context context, Collector<double[]> collector)
+                throws Exception {
             if (epochWatermark == 0) {
                 coefficient = new DenseVector(feedbackBuffer);
                 coefficientDim = coefficient.size();
@@ -372,12 +372,8 @@ public class LogisticRegression
                 updateModel();
             }
             Arrays.fill(gradient.values, 0);
-            try {
-                if (trainData == null) {
-                    trainData = IteratorUtils.toList(trainDataState.get().iterator());
-                }
-            } catch (Exception e) {
-                throw new RuntimeException(e);
+            if (trainData == null) {
+                trainData = IteratorUtils.toList(trainDataState.get().iterator());
             }
             miniBatchData = getMiniBatchData(trainData, localBatchSize);
             Tuple2<Double, Double> weightSumAndLossSum =
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/IterationListener.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/IterationListener.java
@@ -46,7 +46,8 @@ public interface IterationListener<T> {
      *     the invocation of this method.
      * @param collector The collector for returning result values.
      */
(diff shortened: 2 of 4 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
```diff
@@ -362,7 +362,6 @@ private void updateModel() {
         @Override
         public void onEpochWatermarkIncremented(
                 int epochWatermark, Context context, Collector<double[]> collector) {
-            // TODO: let this method throws exception.
```
zhipeng93, line 365: Could you also update this method and do not throw runtime exception in this method?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/Iterations.java
```diff
@@ -117,19 +117,16 @@
      */
     public static DataStreamList iterateUnboundedStreams(
             DataStreamList initVariableStreams, DataStreamList dataStreams, IterationBody body) {
-        return createIteration(
+        return iterateStreams(
```
zhipeng93, line 120: After this renaming, having `iterateUnboundedStreams` and `iterateStreams`, without `iterateBoundedStreamsUntilTermination` seems a bit confusing to me. These APIs are not symmetric for bounded and unbounded cases. A side effect of doing so is that we cannot restrict the following use case: supporting termination criteria on iterating unbounded streams.


Could you explain why do you propose to make the change here?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/Iterations.java
```diff
@@ -261,10 +256,6 @@ private static DataStreamList createIteration(
         }
 
         List<DataStream<?>> tailsAndCriteriaTails = new ArrayList<>(tails.getDataStreams());
-        checkState(
```
zhipeng93, line 264: Do we suppose to support termination criteria in unbouned iterations? If we are not doing this, we should probably keep this check.

zhipeng93 (commented):

lindong28:
@gaoyunhaii Could you help review this PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
lindong28: hmm... why do we need to do that?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/Iterations.java
lindong28: This change is reverted.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/Iterations.java
lindong28: This change is reverted.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
lindong28: Thanks for pointing out this issue. The issue has been fixed.

gaoyunhaii:
Thanks @lindong28 for opening the PR and @zhipeng93 for the review! I think it would indeed makes users easier and keep consistency with other UDF interfaces~

gaoyunhaii (approved):

PR #49, opened 2021-12-30
[hotfix] Add scala version as suffix of all modules

zhipeng93:
This PR adds scala version as suffix of all flink-ml sub-modules.


lindong28:
Thanks for the PR. LGTM.

PR #54, opened 2022-01-11
[FLINK-25552] Add Estimator and Transformer for MinMaxScaler  in FlinkML

weibozhao:
Add Transformer and Estimator of MinMaxScaler in FlinkML.


weibozhao:
@zhipeng93 @lindong28 @gaoyunhaii Can you help review this PR?

lindong28:
Thanks for the PR.

Before we continue to more and more algorithm, what's our plan to making sure these algorithm's implementation could meet our performance target? I am concerned that if we add more algorithms before we have a benchmark plan, we might end up refactoring added algorithms in a unnecessary manner, which is probably not a good long term solution.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/** An Estimator which implements the MinMaxScaler algorithm. */
```
zhipeng93, line 42: Could you add a doc/link to explain what is a `minmaxscaler` as we did for other algorithms?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,165 @@
...
+                                DenseVector minVector = null;
+                                DenseVector maxVector = null;
+                                int vecSize = 0;
+                                for (Tuple2<DenseVector, DenseVector> dataPoint : dataPoints) {
+                                    if (maxVector == null) {
+                                        vecSize = dataPoint.f0.size();
+                                        maxVector = dataPoint.f1;
+                                        minVector = dataPoint.f0;
+                                    }
+                                    for (int i = 0; i < vecSize; ++i) {
```
zhipeng93, line 106: nits: vecSize could be replaced with `maxVector.size()`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Generates minMax scaler model data.
+     *
+     * @param minMaxVectors Input distributed minMaxVectors.
+     * @return MinMax scaler model data.
+     */
+    private static DataStream<MinMaxScalerModelData> genModelData(
+            DataStream<Tuple2<DenseVector, DenseVector>> minMaxVectors) {
+        DataStream<MinMaxScalerModelData> modelData =
+                DataStreamUtils.mapPartition(
```
zhipeng93, line 89: nits: using `mapPartition` here may be not that efficient since we cache all the input records in the state.

How about using a self-defined streamOperator as [1]?

[1] https://github.com/zhipeng93/flink-ml/blob/ccd3f9919dbdd3c5123f1cc88636f832f8215ffb/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java#L68

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * Computes max and min values of features.
+     *
+     * @param inputData Input data.
+     * @param featureCol Feature column name.
+     * @return Max and min values of features.
+     */
+    private DataStream<Tuple2<DenseVector, DenseVector>> computeMinMaxVectors(
```
zhipeng93, line 131: The logic of `computeMinMaxVectors` and `genModelData` seems almost the same. Is merging them into one a better implementation?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
```diff
@@ -0,0 +1,56 @@
...
+
+    default T setMax(Double value) {
+        return set(MAX, value);
+    }
+
+    Param<Double> MIN =
+            new DoubleParam(
+                    "min", "Lower bound after transformation.", 0.0, ParamValidators.notNull());
+
+    default Double getMIN() {
```
zhipeng93, line 49: nit: `getMIN` to `getMin`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
```diff
@@ -0,0 +1,56 @@
...
+        return get(MIN);
+    }
+
+    default T setMIN(Double value) {
```
zhipeng93, line 53: ditto.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasOutputCol.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.StringParam;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared outputCol param. */
+public interface HasOutputCol<T> extends WithParams<T> {
```
zhipeng93, line 27: Why do we need a new param here? How about using the existing one, i.e., `HasPredictionCol`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,179 @@
...
+        private MinMaxScalerModelData minMaxScalerModelData;
+        private final double max;
+        private final double min;
+        private final String broadcastKey;
+        private DenseVector maxVector;
+        private DenseVector minVector;
+
+        public PredictLabelFunction(
+                String broadcastKey, double max, double min, String featureCol) {
+            this.max = max;
```
zhipeng93, line 146: nit: Are `loweBound` and `upperBound` better names for `max` and `min`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,179 @@
...
+            if (minMaxScalerModelData == null) {
+                minMaxScalerModelData =
+                        (MinMaxScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
+                maxVector = minMaxScalerModelData.maxVector;
+                minVector = minMaxScalerModelData.minVector;
+            }
+            DenseVector feature = (DenseVector) row.getField(featureCol);
+            DenseVector outputVector = new DenseVector(maxVector.size());
+            if (feature != null) {
```
zhipeng93, line 163: Should we throw an exception here if the input data does not contain the `featureCol`?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,206 @@
...
+    private static void verifyPredictionResult(Table output, String outputCol) throws Exception {
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) output).getTableEnvironment();
+        DataStream<DenseVector> stream =
+                tEnv.toDataStream(output)
+                        .map(
+                                (MapFunction<Row, DenseVector>)
+                                        row -> (DenseVector) row.getField(outputCol));
+        List<DenseVector> result = IteratorUtils.toList(stream.executeAndCollect());
+        for (DenseVector t2 : result) {
```
zhipeng93, line 98: Is this loop necessary here?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,206 @@
...
+    public void testParam() {
+        MinMaxScaler minMaxScaler = new MinMaxScaler();
+        assertEquals("features", minMaxScaler.getFeaturesCol());
+        assertEquals(1.0, minMaxScaler.getMax(), 0.0001);
+        assertEquals(0.0, minMaxScaler.getMIN(), 0.0001);
+        assertEquals("output", minMaxScaler.getOutputCol());
+        minMaxScaler
+                .setFeaturesCol("test_features")
+                .setMax(4.0)
+                .setMIN(0.0)
```
zhipeng93, line 113: How about use different default value for `min`?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,206 @@
...
+                        .setOutputCol("test_output");
+        MinMaxScalerModel model = minMaxScaler.fit(trainData.as("test_features"));
+        Table output = model.transform(predictData.as("test_features"))[0];
+        assertEquals(
+                Arrays.asList("test_features", "test_output"),
+                output.getResolvedSchema().getColumnNames());
+    }
+
+    @Test
+    public void testFewerDistinctPointsThanCluster() {
```
zhipeng93, line 137: Is this test necessary?

zhipeng93 (commented):
Thanks for the PR. It looks pretty good to me. Left some minor comments below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: computeMinMaxVectors execute in parallel mode, genModelData reduce distributed minMax values to one.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: This test checks when some worker has no data. I think it is needed.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: Spark uses setMax() and setMin(), alink also do as spark. min and max has the meaning of lowerBound and upperBound in this algo.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasOutputCol.java
weibozhao: PredictionCol may be better for ml algo. For feature engineer and data proc, outputCol may be better.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: ok

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
weibozhao: I define api as alink and spark.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: This mapPartition only has fewer record. For every worker only has one record. I think it's no efficient problem.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
weibozhao: ditto.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasOutputCol.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,198 @@
...
+        return model;
+    }
+
+    /**
+     * A stream operator to compute the min max vectors in each partition of the input bounded data
+     * stream.
+     */
+    private static class MinMaxReduceFunctionOperator extends AbstractStreamOperator<DenseVector>
+            implements OneInputStreamOperator<DenseVector, DenseVector>, BoundedOneInput {
+        private ListState<DenseVector> minDenseVectorState;
```
zhipeng93, line 113: How about using `minState/maxState` or `lowerBoundState` and `upperBoundState`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
zhipeng93: I am still a little bit concerned about the naming. What is the difference between the semantic of `min/max` and `minVector/maxVector`?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,204 @@
...
+import static org.junit.Assert.assertEquals;
+
+/** Tests {@link MinMaxScaler} and {@link MinMaxScalerModel}. */
+public class MinMaxScalerTest {
+    @Rule public final TemporaryFolder tempFolder = new TemporaryFolder();
+    private StreamExecutionEnvironment env;
+    private StreamTableEnvironment tEnv;
+    private Table trainData;
+    private Table predictData;
+    private static final List<Row> trainRows =
```
zhipeng93, line 62: nits: How about using `trainData`?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private static void verifyPredictionResult(Table output, String outputCol) throws Exception {
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) output).getTableEnvironment();
+        DataStream<DenseVector> stream =
+                tEnv.toDataStream(output)
+                        .map(
+                                (MapFunction<Row, DenseVector>)
+                                        row -> (DenseVector) row.getField(outputCol));
+        List<DenseVector> result = IteratorUtils.toList(stream.executeAndCollect());
```
zhipeng93, line 97: Shall we also check the size of `result`?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,204 @@
...
+        DataStream<Row> output = tEnv.toDataStream(modelData);
+        assertEquals("minVector", modelData.getResolvedSchema().getColumnNames().get(0));
+        assertEquals("maxVector", modelData.getResolvedSchema().getColumnNames().get(1));
+        List<Row> modelRows = IteratorUtils.toList(output.executeAndCollect());
+        MinMaxScalerModelData data =
+                new MinMaxScalerModelData(
+                        (DenseVector) modelRows.get(0).getField(0),
+                        (DenseVector) modelRows.get(0).getField(1));
+        Assert.assertNotNull(data);
+        assertEquals(data.maxVector.size(), data.maxVector.size());
```
zhipeng93, line 191: Could you please also check the value of model data?

zhipeng93 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,198 @@
...
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public MinMaxScalerModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        final String featureCol = getFeaturesCol();
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<DenseVector> vectors =
```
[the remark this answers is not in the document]
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,198 @@
...
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                (MapFunction<Row, DenseVector>)
+                                        value -> (DenseVector) value.getField(featureCol));
+        DataStream<DenseVector> minMaxVectors =
```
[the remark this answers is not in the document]
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * A stream operator to compute the min max vectors in each partition of the input bounded data
+     * stream.
+     */
+    private static class MinMaxReduceFunctionOperator extends AbstractStreamOperator<DenseVector>
+            implements OneInputStreamOperator<DenseVector, DenseVector>, BoundedOneInput {
+        private ListState<DenseVector> minDenseVectorState;
+        private ListState<DenseVector> maxDenseVectorState;
+
+        private DenseVector minDenseVector;
```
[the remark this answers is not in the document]
weibozhao: testFewerDistinctPointsThanCluster()  test this condition. Parallel is 4 and only 1 record trained.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,198 @@
...
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxDenseVectorState",
+                                            getOperatorConfig()
+                                                    .getTypeSerializerIn(
+                                                            0, getClass().getClassLoader())));
+            Iterator<DenseVector> minIterator = minDenseVectorState.get().iterator();
+            Iterator<DenseVector> maxIterator = maxDenseVectorState.get().iterator();
+            if (minIterator.hasNext()) {
```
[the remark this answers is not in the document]
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,197 @@
...
+            implements OneInputStreamOperator<DenseVector, DenseVector>, BoundedOneInput {
+        private ListState<DenseVector> minState;
+        private ListState<DenseVector> maxState;
+
+        private DenseVector minVector;
+        private DenseVector maxVector;
+
+        @Override
+        public void endInput() {
+            output.collect(new StreamRecord<>(minVector));
```
zhipeng93, line 120: If there is no data in one partition, it throw a NPE here.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
zhipeng93: Could you also verify the execution result here?

zhipeng93 (commented):
Thanks for the update. There is only one problem here. Please check out the comments...

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: OK

zhipeng93:
Thanks for the update. This PR looks pretty good to me. 

@lindong28 @yunfengzhou-hub Do you want to take a look at this PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,181 @@
...
+                for (int i = 0; i < maxVector.size(); ++i) {
```
yunfengzhou-hub, line 164: This for loop might also be replaced by BLAS operations.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,208 @@
...
+                        .setPredictionCol("test_output");
+        MinMaxScalerModel model = minMaxScaler.fit(trainDataTable.as("test_features"));
+        Table output = model.transform(predictDataTable.as("test_features"))[0];
+        assertEquals(
+                Arrays.asList("test_features", "test_output"),
+                output.getResolvedSchema().getColumnNames());
+    }
+
+    @Test
+    public void testFewerDistinctPointsThanCluster() throws Exception {
```
yunfengzhou-hub, line 137: Maybe we can refine this naming, as there is no concept of "cluster" in MinMaxScaler.

yunfengzhou-hub (commented):
Thanks for the update. The general structure looks good to me. Some minor comments are as follows.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,181 @@
...
+            if (minMaxScalerModelData == null) {
+                minMaxScalerModelData =
+                        (MinMaxScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
+                maxVector = minMaxScalerModelData.maxVector;
+                minVector = minMaxScalerModelData.minVector;
+            }
+            DenseVector feature = (DenseVector) row.getField(featureCol);
+            DenseVector outputVector = new DenseVector(maxVector.size());
+            if (feature != null) {
```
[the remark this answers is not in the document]
weibozhao: Thanks for your comments. 

I think, HasHandleInvalid is not needed here.
1. Spark and alink have no parameter about invalid. 
2. If feature range not agree each other, code will throw Exception with array range err. 
3. In HasHandleInvalid parameter:  KEEP and SKIP are improper here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,205 @@
...
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxState",
+                                            getOperatorConfig()
+                                                    .getTypeSerializerIn(
+                                                            0, getClass().getClassLoader())));
+            Iterator<DenseVector> minIterator = minState.get().iterator();
+            Iterator<DenseVector> maxIterator = maxState.get().iterator();
+            if (minIterator.hasNext()) {
```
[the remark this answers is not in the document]
weibozhao: OK, I will refine it later.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,181 @@
...
+                for (int i = 0; i < maxVector.size(); ++i) {
+                    if ((minVector.values[i] - maxVector.values[i]) != 0.0) {
+                        outputVector.values[i] =
+                                (feature.values[i] - minVector.values[i])
+                                                / (maxVector.values[i] - minVector.values[i])
```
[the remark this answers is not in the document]
weibozhao: I think spanVector maybe not very good. 
1. In Spark and alink, maxVector and minVector are saved for modelData. 
2. If we spanVector, code may be not clear as minVector and maxVector used.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: Kmeans and knn user this name already, I just take it from other ut.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,205 @@
...
+            if (minVector == null) {
+                int vecSize = currentValue.size();
+                minVector = new DenseVector(vecSize);
+                maxVector = new DenseVector(vecSize);
+                System.arraycopy(currentValue.values, 0, minVector.values, 0, vecSize);
+                System.arraycopy(currentValue.values, 0, maxVector.values, 0, vecSize);
+
+            } else {
+                for (int i = 0; i < currentValue.size(); ++i) {
+                    minVector.values[i] = Math.min(minVector.values[i], currentValue.values[i]);
```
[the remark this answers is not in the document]
weibozhao: As I Know, blas has no function like minVector or maxVector.
If user blas1 function(minVector is blas1 function), loop may not be avoided, too.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: There is a judge about max-min==0 exists in the loop, then we can not use blas here. 
If we want to use blas, we need another loop to prepare data. I think it is more expensive.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. If this is the case, then I think MinMaxScaler might not need this parameter either.

What do you think of the other part of the question above?
> if the feature vector exists, but its dimension is different from that of maxVector, or its max/min value exceeds maxVector/minVector's range, then maybe we should also throw exception.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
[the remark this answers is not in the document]
yunfengzhou-hub: I'm not sure this is the case for Spark. In Spark's MinMaxScaler.scala I found the following code:
```scala
values(i) = (values(i) - minArray(i)) * scaleArray(i) + minValue
```
which shows that Spark uses `minArray` and `scaleArray`, instead of `minArray` and `maxArray`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
yunfengzhou-hub: KMeans is a clustering algorithm, so I understand that there is the concept of cluster. But in MinMaxScaler there is no such "cluster" thing. I think it would be more proper if we rename it to `testMinMaxEquals()`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,201 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * An Estimator which implements the MinMaxScaler algorithm.
+ *
+ * <p>See https://en.wikipedia.org/wiki/Feature_scaling#Rescaling_(min-max_normalization).
```
lindong28, line 55: Since the link provided here does not explain the use of min/max parameter, it does not exactly describe what this class does.

Should we provide more detailed algorithm explanation here similar to Spark's MinMaxScaler Javadoc?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,181 @@
...
+        public Row map(Row row) {
+            if (minMaxScalerModelData == null) {
+                minMaxScalerModelData =
+                        (MinMaxScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
+                maxVector = minMaxScalerModelData.maxVector;
+                minVector = minMaxScalerModelData.minVector;
+            }
+            DenseVector feature = (DenseVector) row.getField(featureCol);
+            DenseVector outputVector = new DenseVector(maxVector.size());
```
lindong28, line 162: nits: would it be simpler to rename `outputVector` as `output` for simplicity and consistency with `feature`?

lindong28 (commented):
Thanks for the PR! Left some minor comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
yunfengzhou-hub: Just for this code, I think there is a simple way to do the max == min judgement just once, like follows.
```java

        public Row map(Row row) {
            if (minMaxScalerModelData == null) {
                minMaxScalerModelData =
                        (MinMaxScalerModelData)
                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
                maxVector = minMaxScalerModelData.maxVector;
                minVector = minMaxScalerModelData.minVector;
                for (int i = 0; i < maxVector.size(); ++i) {
                    if ((minVector.values[i] - maxVector.values[i]) == 0.0) {
                        maxVector.values[i] = maxVector.values[i] + 1.0;
                        minVector.values[i] = minVector.values[i] - 1.0;
                    }
                }
            }
            DenseVector feature = (DenseVector) row.getField(featureCol);
            DenseVector outputVector = new DenseVector(maxVector.size());
            if (feature != null) {
                for (int i = 0; i < maxVector.size(); ++i) {
                    outputVector.values[i] =
                            (feature.values[i] - minVector.values[i])
                                    / (maxVector.values[i] - minVector.values[i])
                                    * (upperBound - lowerBound)
                                    + lowerBound;
                }
                return Row.join(row, Row.of(outputVector));
            } else {
                throw new RuntimeException("Feature value is null, please check your input data.");
            }
        }
```
With this code, now it is possible to apply BLAS operations.

I noticed that Spark also uses for loop in this and the situation above, so I agree that we may postpone any further optimizations like this after this PR.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,208 @@
...
+                        env, minMaxScalerModel, tempFolder.newFolder().getAbsolutePath());
+        assertEquals(
+                Arrays.asList("minVector", "maxVector"),
+                minMaxScalerModel.getModelData()[0].getResolvedSchema().getColumnNames());
+        Table output = minMaxScalerModel.transform(predictDataTable)[0];
+        verifyPredictionResult(output, minMaxScaler.getPredictionCol(), Vectors.dense(0.75, 0.3));
+    }
+
+    @Test
+    public void testModelSaveLoadAndPredict() throws Exception {
```
lindong28, line 170: It seems that all functionalities covered in this test is already covered by `testSaveLoadAndPredict`. Could we remove this test for simplicity?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,208 @@
...
+        assertEquals(expected, result.get(0));
+    }
+
+    @Test
+    public void testParam() {
+        MinMaxScaler minMaxScaler = new MinMaxScaler();
+        assertEquals("features", minMaxScaler.getFeaturesCol());
+        assertEquals(1.0, minMaxScaler.getMax(), 0.0001);
+        assertEquals(0.0, minMaxScaler.getMin(), 0.0001);
+        assertEquals("prediction", minMaxScaler.getPredictionCol());
```
lindong28, line 109: nits: could we move `prediction` to be right after "features" for a bit more consistency, i.e. column name features are close to each other?

Same for other `assertEquals` usages and `testFeaturePredictionParam()`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,208 @@
...
+    private Table predictDataTable;
+    private static final List<Row> trainData =
+            new ArrayList<>(
+                    Arrays.asList(
+                            Row.of(Vectors.dense(0.0, 3.0)),
+                            Row.of(Vectors.dense(2.1, 0.0)),
+                            Row.of(Vectors.dense(4.1, 5.1)),
+                            Row.of(Vectors.dense(6.1, 8.1)),
+                            Row.of(Vectors.dense(200, 300))));
+    private static final List<Row> predictRows =
```
lindong28, line 70: Could we have multiple elements instead of just one element in the prediction stream? In general this could help provide more test coverage.

And could we rename `predictRows` as `predictData` for consistency with `trainData`?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK，I got it. I will refine it later.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,208 @@
...
+        config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
+        env = StreamExecutionEnvironment.getExecutionEnvironment(config);
+        env.setParallelism(4);
+        env.enableCheckpointing(100);
+        env.setRestartStrategy(RestartStrategies.noRestart());
+        tEnv = StreamTableEnvironment.create(env);
+        Schema schema = Schema.newBuilder().column("f0", DataTypes.of(DenseVector.class)).build();
+        DataStream<Row> dataStream = env.fromCollection(trainData);
+        trainDataTable = tEnv.fromDataStream(dataStream, schema).as("features");
+        DataStream<Row> predDataStream = env.fromCollection(predictRows);
```
lindong28, line 85: It appears that we don't need to explicitly create the schema. Would it be simpler to use the code below?

```
        trainDataTable = tEnv.fromDataStream(env.fromCollection(trainData)).as("features");
        predictDataTable = tEnv.fromDataStream(env.fromCollection(predictRows)).as("features");
```

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,201 @@
...
+            implements OneInputStreamOperator<DenseVector, DenseVector>, BoundedOneInput {
+        private ListState<DenseVector> minState;
+        private ListState<DenseVector> maxState;
+
+        private DenseVector minVector;
+        private DenseVector maxVector;
+
+        @Override
+        public void endInput() {
+            if (minVector != null) {
```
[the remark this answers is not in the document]
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: If featureValue[i] != maxVector[i], above code may not get the same result as original code.
In read world, data for train and predict are different, featureValue[i] != maxVector[i] is existed.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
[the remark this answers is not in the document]
weibozhao: val scaleArray = Array.tabulate(numFeatures) { i =>
      val range = originalMax(i) - originalMin(i)

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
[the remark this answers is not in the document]
weibozhao: OK, I have add exception.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
yunfengzhou-hub: > if the feature vector exists, but its dimension is different from that of maxVector, or its max/min value exceeds maxVector/minVector's range, then maybe we should also throw exception.

If `featureValue[i] != maxVector[i]`, the input data would have violated the rule above and would throw exceptions. So there would be no result generated in both cases, right?

From this, I think we should also add tests about illegal inputs, like `OneHotEncoderTest.testNonIndexedPredictData`.

weibozhao: I don‘t think so. 
If featureValue[i] != maxVector[i],  predict data may have mo problem. For train data defines the rule and we can't constrain predict data follow this rule. 

For example, in train data one feature is people's age, all feature values in the train data is 20 years old. If predict data has a sample which feature is 21 years old which will be scaled to 0.5*(max - min).

yunfengzhou-hub: Got it. Then how about this one?
```java
        @Override
        public Row map(Row row) {
            if (minMaxScalerModelData == null) {
                minMaxScalerModelData =
                        (MinMaxScalerModelData)
                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
                maxVector = minMaxScalerModelData.maxVector;
                minVector = minMaxScalerModelData.minVector;
                scaleVector = new DenseVector(minVector.size());
                offsetVector = new DenseVector(minVector.size());
                for (int i = 0; i < maxVector.size(); ++i) {
                    if ((minVector.values[i] - maxVector.values[i]) == 0.0) {
                        scaleVector.values[i] = Double.POSITIVE_INFINITY;
                        offsetVector.values[i] = 0.5;
                    } else {
                        scaleVector.values[i] = maxVector.values[i] - minVector.values[i];
                        offsetVector.values[i] = 0.0;
                    }
                }
            }
            DenseVector feature = (DenseVector) row.getField(featureCol);
            DenseVector outputVector = new DenseVector(maxVector.size());
            if (feature != null) {
                for (int i = 0; i < maxVector.size(); ++i) {
                    outputVector.values[i] =
                            ((feature.values[i] - minVector.values[i])
                                                    / scaleVector.values[i]
                                                    + offsetVector.values[i])
                                            * (upperBound - lowerBound)
                                    + lowerBound;
                }
                return Row.join(row, Row.of(outputVector));
            } else {
                throw new RuntimeException("Feature value is null, please check your input data.");
            }
        }
```

This code should meet the requirements without using if conditions.

yunfengzhou-hub: Besides, could you please add a test case about maxVector[i] == minVector[i] && featureValue[i] != maxVector[i]? I think it would be helpful to test these corner cases.

yunfengzhou-hub: Even better:
```java

        @Override
        public Row map(Row row) {
            if (minMaxScalerModelData == null) {
                minMaxScalerModelData =
                        (MinMaxScalerModelData)
                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
                maxVector = minMaxScalerModelData.maxVector;
                minVector = minMaxScalerModelData.minVector;
                scaleVector = new DenseVector(minVector.size());
                offsetVector = new DenseVector(minVector.size());
                for (int i = 0; i < maxVector.size(); ++i) {
                    if ((minVector.values[i] - maxVector.values[i]) == 0.0) {
                        scaleVector.values[i] = 0.0;
                        offsetVector.values[i] = (upperBound + lowerBound) / 2;
                    } else {
                        scaleVector.values[i] = (upperBound - lowerBound) / (maxVector.values[i] - minVector.values[i]);
                        offsetVector.values[i] = lowerBound - minVector.values[i] * scaleVector.values[i];
                    }
                }
            }
            DenseVector feature = (DenseVector) row.getField(featureCol);
            DenseVector outputVector = new DenseVector(maxVector.size());
            if (feature != null) {
                for (int i = 0; i < maxVector.size(); ++i) {
                    outputVector.values[i] = feature.values[i] * scaleVector.values[i] + offsetVector.values[i];
                }
                return Row.join(row, Row.of(outputVector));
            } else {
                throw new RuntimeException("Feature value is null, please check your input data.");
            }
        }
```

With this implementation, there would be only one multiplication and one addition in the for loop, which can definitely be replaced by `BLAS.axpy`. It would cause some precision loss (less then 1e-16 on this PR's test data), but I think the performance improvement worth it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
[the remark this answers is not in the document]
yunfengzhou-hub: Sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: I got it. But in BLAS.axpy, a is a double value which is a constant, not a value in vector.
If use BLAS, we need two BLAS func: vec.*vec and axpy. This means we need two vector loop.  I don't think it's more efficiency. 

Your code can avoid the judge of "if ((minVector.values[i] - maxVector.values[i]) == 0.0)" in transform, I will refine my code as you done.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
lindong28: Hmm... is this issue fixed?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
```diff
@@ -0,0 +1,56 @@
...
+
+    default Double getMax() {
+        return get(MAX);
+    }
+
+    default T setMax(Double value) {
+        return set(MAX, value);
+    }
+
+    Param<Double> MIN =
```
lindong28, line 45: nits: Could we move the definition of this MIN parameter to be above MAX?

This would be more consistent with the Java coding style, where we typically declare all class member variables before declaring methods.

And it would also make it easier for users to get all the parameters defined in this class.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModelData.java
```diff
@@ -0,0 +1,119 @@
...
+                .map(
+                        x ->
+                                new MinMaxScalerModelData(
+                                        (DenseVector) x.getField(0), (DenseVector) x.getField(1)));
+    }
+
+    /** Encoder for {@link MinMaxScalerModelData}. */
+    public static class ModelDataEncoder implements Encoder<MinMaxScalerModelData> {
+        @Override
+        public void encode(MinMaxScalerModelData minMaxScalerModelData, OutputStream outputStream)
```
lindong28, line 78: Super nits: all other classes (except `KnnModelData`) uses `modelData` as the input parameter name of this method. Any chance we can do the same here for consistency and shorten the code a little bit?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            minState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
```
lindong28, line 163: Hmm.. instead of using `getOperatorConfig().getTypeSerializerIn(...)`, would it be better to use the code below for simplicity and consistency with other usages of `getListState(...)` in Flink ML?

```
new ListStateDescriptor<>("minState", TypeInformation.of(DenseVector.class))
```

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
```diff
@@ -0,0 +1,210 @@
...
+        @Override
+        @SuppressWarnings("unchecked")
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            minState.clear();
+            maxState.clear();
+            if (minVector != null) {
+                minState.add(minVector);
+            }
+            if (maxVector != null) {
```
lindong28, line 190: nits: instead of using two `if` statements, would it be simpler to do the following?

```
if (minVector != null) {
    minState.add(minVector);
    maxState.add(maxVector);
}
```

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,189 @@
...
+        DataStream<Row> data = tEnv.toDataStream(inputs[0]);
+        DataStream<MinMaxScalerModelData> minMaxScalerModel =
+                MinMaxScalerModelData.getModelDataStream(modelDataTable);
+        final String broadcastModelKey = "broadcastModelKey";
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(),
+                                ExternalTypeInfo.of(DenseVector.class)),
```
lindong28, line 84: Hmm.. is there any reason in particular to use `ExternalTypeInfo`?

Would it be better to use `TypeInformation.of(DenseVector.class)` for consistency with e.g. `LogisticRegressionModel::transform(...)`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,189 @@
...
+                ReadWriteUtils.loadModelData(
+                        env, path, new MinMaxScalerModelData.ModelDataDecoder());
+        return model.setModelData(tEnv.fromDataStream(modelData));
+    }
+
+    /** This operator loads model data and predicts result. */
+    private static class PredictOutputFunction extends RichMapFunction<Row, Row> {
+        private final String featureCol;
+        private final String broadcastKey;
+        private MinMaxScalerModelData minMaxScalerModelData;
```
lindong28, line 138: Given that the code already stores `scaleVector` and `offsetVector`, do we still need to keep `minMaxScalerModelData` as a class member variable here?

Same for `minVector`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertEquals("prediction", minMaxScaler.getPredictionCol());
+        assertEquals(1.0, minMaxScaler.getMax(), 0.0001);
+        assertEquals(0.0, minMaxScaler.getMin(), 0.0001);
+        minMaxScaler
+                .setFeaturesCol("test_features")
+                .setPredictionCol("test_output")
+                .setMax(4.0)
+                .setMin(1.0);
+        assertEquals("test_features", minMaxScaler.getFeaturesCol());
+        assertEquals(1.0, minMaxScaler.getMin(), 0.0001);
```
lindong28, line 111: Maybe we should unify the value of epsilon used to check double values in all tests, instead of having each test uses different values.

How about we use `1e-5` here and all other tests? `1e-5` seems a bit stricter and also shorter than `0.0001`. It is used in `KMeansTest`.

We can change other tests in the future.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    @Test
+    public void testSaveLoadAndPredict() throws Exception {
+        MinMaxScaler minMaxScaler = new MinMaxScaler();
+        MinMaxScaler loadedMinMaxScaler =
+                StageTestUtils.saveAndReload(
+                        env, minMaxScaler, tempFolder.newFolder().getAbsolutePath());
+        MinMaxScalerModel minMaxScalerModel = loadedMinMaxScaler.fit(trainDataTable);
+        minMaxScalerModel =
```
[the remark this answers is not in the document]
weibozhao: Sounds good!

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,205 @@
...
+        verifyPredictionResult(output, minMaxScaler.getPredictionCol(), Vectors.dense(0.75, 0.225));
+    }
+
+    @Test
+    public void testGetModelData() throws Exception {
+        MinMaxScaler minMaxScaler = new MinMaxScaler();
+        MinMaxScalerModel minMaxScalerModel = minMaxScaler.fit(trainDataTable);
+        Table modelData = minMaxScalerModel.getModelData()[0];
+        DataStream<Row> output = tEnv.toDataStream(modelData);
+        assertEquals("minVector", modelData.getResolvedSchema().getColumnNames().get(0));
```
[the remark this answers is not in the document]
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertEquals("maxVector", modelData.getResolvedSchema().getColumnNames().get(1));
+        List<Row> modelRows = IteratorUtils.toList(output.executeAndCollect());
+        MinMaxScalerModelData data =
```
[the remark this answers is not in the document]
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Params for {@link MinMaxScaler}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface MinMaxScalerParams<T> extends HasFeaturesCol<T>, HasPredictionCol<T> {
+    Param<Double> MAX =
+            new DoubleParam(
+                    "max", "Upper bound after transformation.", 1.0, ParamValidators.notNull());
```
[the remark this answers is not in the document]
weibozhao: I agree with it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerParams.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModelData.java
weibozhao: OK, I will refine knnModelData, too.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScaler.java
weibozhao: OK

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: It's better.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: OK, I will refine it.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,205 @@
...
+    public void testParam() {
+        MinMaxScaler minMaxScaler = new MinMaxScaler();
+        assertEquals("features", minMaxScaler.getFeaturesCol());
+        assertEquals("prediction", minMaxScaler.getPredictionCol());
+        assertEquals(1.0, minMaxScaler.getMax(), 0.0001);
+        assertEquals(0.0, minMaxScaler.getMin(), 0.0001);
+        minMaxScaler
+                .setFeaturesCol("test_features")
+                .setPredictionCol("test_output")
+                .setMax(4.0)
```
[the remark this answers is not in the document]
weibozhao: OK, I will check this cases and refine them.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: fixed already.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
```diff
@@ -0,0 +1,187 @@
...
+            if (scaleVector == null) {
+                MinMaxScalerModelData minMaxScalerModelData =
+                        (MinMaxScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
+                DenseVector minVector = minMaxScalerModelData.minVector;
+                DenseVector maxVector = minMaxScalerModelData.maxVector;
+                scaleVector = new DenseVector(minVector.size());
+                offsetVector = new DenseVector(minVector.size());
+                for (int i = 0; i < maxVector.size(); ++i) {
+                    if ((minVector.values[i] - maxVector.values[i]) == 0.0) {
```
lindong28, line 162: According to the discussion in [1][2], it is could lead to unexpected bug if we use `==` to check whether two double values are equal.

How about we follow the recommendation in [3] and use the following code to check equality? The idea here is to explicitly specify an epsilon similar to what we did in tests.

```
Math.abs(minVector.values[i] - maxVector.values[i]) < 1e-5
```

[1] https://stackoverflow.com/questions/25160375/comparing-double-values-for-equality-in-java
[2] https://publicobject.com/2009/11/floating-point-equality.html
[3] https://www.baeldung.com/java-comparing-doubles

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
lindong28: @zhipeng93 @weibozhao Currently in all other model classes, we don't explicitly check whether the feature value is null. Should we follow the same practice here for consistency?

We could alternatively choose to consistently check whether feature value is null, and throw exception if it is null. But even if we don't check this, we will already have NullPointerException with proper stacktrace, which should be sufficient for us to locate the problem (i.e. feature value is null). I personally think it is simpler not to check this. What do you think?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,218 @@
...
+                (StreamTableEnvironment) ((TableImpl) output).getTableEnvironment();
+        DataStream<DenseVector> stream =
+                tEnv.toDataStream(output)
+                        .map(
+                                (MapFunction<Row, DenseVector>)
+                                        row -> (DenseVector) row.getField(outputCol));
+        List<DenseVector> result = IteratorUtils.toList(stream.executeAndCollect());
+
+        result.sort(
+                (first, second) -> {
```
lindong28, line 105: The Javadoc of `Comparator` says the following:

```
     * It is generally the case, but <i>not</i> strictly required that
     * <tt>(compare(x, y)==0) == (x.equals(y))</tt>.  Generally speaking,
     * any comparator that violates this condition should clearly indicate
     * this fact.  The recommended language is "Note: this comparator
     * imposes orderings that are inconsistent with equals."
```

It might be better to take the vector size into consideration so that the comparator returns 0 iff `first == second`. The main benefit is that the code is self-contained and could be copied/used in other places.

Alternatively we can just add the comment suggested by the comparator Javadoc.

lindong28 (commented):
Thanks for the update! Looks good overall. Left just some minor comments.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
```diff
@@ -0,0 +1,218 @@
...
+                    for (int x = 0; x < first.size(); x++) {
+                        int cmp = Double.compare(first.get(x), second.get(x));
+                        if (cmp != 0) {
+                            return cmp;
+                        }
+                    }
+                    return 0;
+                });
+        assertEquals(expected.size(), result.size());
```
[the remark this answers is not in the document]
weibozhao: OK, I will remove it.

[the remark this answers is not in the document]
weibozhao: I agree with you, I will remove it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: I think, [3] is much better. I will refine the code later.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/minmaxscaler/MinMaxScalerModel.java
weibozhao: I have discussed with @zhipeng93. We think, it's better to remove this exception. I will refine the code later.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: OK, I will add the javadoc.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MinMaxScalerTest.java
weibozhao: done

lindong28 (approved):
Thanks for the update! LGTM. @zhipeng93 do you want to take another look?

zhipeng93:
Thanks @weibozhao for the PR and @lindong28 @yunfengzhou-hub for the review. Merging now.

PR #56, opened 2022-01-20
[FLINK-25616] Add Transformer for VectorAssembler

weibozhao:
Add Transformer for VectorAssembler in FlinkML.

## Brief change log
Add VectorAssembler.

Does this pull request introduce a new feature? (no)

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
@@ -1,0 +1,182 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.vectorassembler;
+
+import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * A feature transformer that combines a given list of input columns into a vector column. Types of
+ * input columns must be either vector or numerical value.
+ */
+public class VectorAssembler
+        implements Transformer<VectorAssembler>, VectorAssemblerParams<VectorAssembler> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final double RATIO = 1.5;
+
+    public VectorAssembler() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(), TypeInformation.of(Vector.class)),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .flatMap(
+                                new AssemblerFunc(getInputCols(), getHandleInvalid()),
+                                outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    private static class AssemblerFunc implements FlatMapFunction<Row, Row> {
+        private final String[] inputCols;
+        private final String handleInvalid;
+
+        public AssemblerFunc(String[] inputCols, String handleInvalid) {
+            this.inputCols = inputCols;
+            this.handleInvalid = handleInvalid;
+        }
+
+        @Override
+        public void flatMap(Row value, Collector<Row> out) throws Exception {
+            try {
+                Object[] objects = new Object[inputCols.length];
+                for (int i = 0; i < objects.length; ++i) {
+                    objects[i] = value.getField(inputCols[i]);
+                }
+                Vector assembledVector = assemble(objects);
+                out.collect(Row.join(value, Row.of(assembledVector)));
+            } catch (Exception e) {
+                switch (handleInvalid) {
+                    case VectorAssemblerParams.ERROR_INVALID:
+                        throw e;
+                    case VectorAssemblerParams.SKIP_INVALID:
+                        return;
+                    case VectorAssemblerParams.KEEP_INVALID:
+                        out.collect(Row.join(value, Row.of((Object) null)));
+                        return;
+                    default:
+                        throw new UnsupportedOperationException(
+                                "handleInvalid=" + handleInvalid + " is not supported");
+                }
+            }
+        }
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static VectorAssembler load(StreamTableEnvironment env, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    private static Vector assemble(Object[] objects) {
+        int offset = 0;
+        Map<Integer, Double> map = new LinkedHashMap<>(objects.length);
+        for (Object object : objects) {
+            Preconditions.checkNotNull(object, "Input column value should not be null.");
+            if (object instanceof Number) {
+                map.put(offset++, ((Number) object).doubleValue());
+            } else if (object instanceof Vector) {
+                offset = appendVector((Vector) object, map, offset);
+            } else {
+                throw new IllegalArgumentException("Input type has not been supported yet.");
+            }
+        }
+
+        if (map.size() * RATIO > offset) {
+            DenseVector assembledVector = new DenseVector(offset);
+            for (int key : map.keySet()) {
+                assembledVector.values[key] = map.get(key);
+            }
+            return assembledVector;
+        } else {
+            return convertMapToSparseVector(offset, map);
+        }
+    }
+
+    private static int appendVector(Vector vec, Map<Integer, Double> map, int offset) {
+        if (vec instanceof SparseVector) {
+            SparseVector sparseVector = (SparseVector) vec;
+            int[] indices = sparseVector.indices;
+            double[] values = sparseVector.values;
+            for (int i = 0; i < indices.length; ++i) {
+                map.put(offset + indices[i], values[i]);
+            }
+            offset += sparseVector.size();
+        } else {
+            DenseVector denseVector = (DenseVector) vec;
+            for (int i = 0; i < denseVector.size(); ++i) {
+                map.put(offset++, denseVector.values[i]);
+            }
+        }
+        return offset;
+    }
+
+    private static SparseVector convertMapToSparseVector(int size, Map<Integer, Double> map) {
+        int[] indices = new int[map.size()];
+        double[] values = new double[map.size()];
+        int offset = 0;
+        for (Map.Entry<Integer, Double> entry : map.entrySet()) {
+            indices[offset] = entry.getKey();
+            values[offset++] = entry.getValue();
+        }
+        return new SparseVector(size, indices, values);
+    }
+}
--- a/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
+++ b/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
@@ -1,0 +1,178 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


weibozhao:
@zhipeng93  @lindong28  @gaoyunhaii  Can you help review this PR?

lindong28:
Before we continue to more and more algorithm, what's our plan to making sure these algorithm's implementation could meet our performance target?

weibozhao:
I get your idea

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,174 @@
...
+                new RowTypeInfo(
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(), TypeInformation.of(Vector.class)),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                (MapFunction<Row, Row>)
+                                        row -> {
+                                            Object[] objects = new Object[getInputCols().length];
```
yunfengzhou-hub, line 79: It might be better to make this lambda function a static class, and read the value of `getInputCols` and `getHandleInvalid` just once in its constructor. Every time `getInputCols` is invoked, there would be at least a query to a map and and `ParamValidator` check, which could affect performance a bit.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    private static Vector assemble(Object[] objects, String handleInvalid) {
+        int offset = 0;
+        Map<Integer, Double> map = new HashMap<>(objects.length);
```
yunfengzhou-hub, line 109: How about using a `TreeMap` or `LinkedHashMap` here? When converting the map to `SparseVector`, this could help keeping the indexes from `map.entrySet()` in order.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,144 @@
...
+                        .build();
+        DataStream<Row> dataStream = env.fromCollection(INPUT_DATA);
+        inputDataTable =
+                tEnv.fromDataStream(dataStream, schema).as("id", "vec", "num", "sparseVec");
+    }
+
+    private void verifyPredictionResult(Table output, String outputCol) throws Exception {
+        DataStream<Row> dataStream = tEnv.toDataStream(output);
+        List<Row> results = IteratorUtils.toList(dataStream.executeAndCollect());
+        for (Row result : results) {
```
yunfengzhou-hub, line 97: Can we check the number of results generated? If the `toList` returns an empty list, the test cases would still pass.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.rules.TemporaryFolder;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/** Tests VectorAssembler. */
+public class VectorAssemblerTest {
```
yunfengzhou-hub, line 51: It might be better to add test cases for invalid inputs, to test whether `handleInvalid` functions correctly.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,144 @@
...
+    private StreamExecutionEnvironment env;
+    private StreamTableEnvironment tEnv;
+    private Table inputDataTable;
+    private static final List<Row> INPUT_DATA =
+            Arrays.asList(
+                    Row.of(
+                            0,
+                            Vectors.dense(2.1, 3.1),
+                            1.0,
+                            Vectors.sparse(5, new int[] {3}, new double[] {1.0})),
```
yunfengzhou-hub, line 63: Can we test if the values in the sparse vector is not ordered? For example,
```java
Vectors.sparse(5, new int[]{4,2,3,1}, new double[]{4.0,2.0,3.0,1.0})
```

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
yunfengzhou-hub, line 51: We can have `VectorAssemblerTest extends AbstractTestBase` to possibly accelerate test cases.

yunfengzhou-hub (commented):
Thanks for the PR. Generally LGTM. Left some comments as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: Sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,174 @@
...
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static VectorAssembler load(StreamExecutionEnvironment env, String path)
```
yunfengzhou-hub, line 97: We have updated all `load` method to use `StreamTableEnvironment`.

yunfengzhou-hub, line 97: We have updated all `load` method to use `StreamTableEnvironment`.

yunfengzhou-hub (commented):
Left one comment as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: OK, It's better.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+            return assembledVector;
+        } else {
+            return convertMapToSparseVector(offset, map);
+        }
+    }
+
+    private static int appendVector(Vector vec, Map<Integer, Double> map, int offset) {
+        if (vec == null) {
+            throw new RuntimeException("VectorAssembler Error: vector data is null.");
```
yunfengzhou-hub, line 151: nit: `Preconditions.checkNotNull(vec, "VectorAssembler Error:...")`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,184 @@
...
+        Map<Integer, Double> map = new LinkedHashMap<>(objects.length);
+        for (Object object : objects) {
+            try {
+                if (object instanceof Number) {
+                    map.put(offset++, ((Number) object).doubleValue());
+                } else if (object instanceof Vector) {
+                    offset = appendVector((Vector) object, map, offset);
+                } else {
+                    throw new UnsupportedOperationException(
+                            "Vector assembler : input type not support yet.");
```
yunfengzhou-hub, line 125: nit: `"has not been supported yet"`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    private static int appendVector(Vector vec, Map<Integer, Double> map, int offset) {
+        if (vec == null) {
+            throw new RuntimeException("VectorAssembler Error: vector data is null.");
+        }
+        if (vec instanceof SparseVector) {
+            SparseVector sparseVector = (SparseVector) vec;
+            if (sparseVector.size() <= 0) {
+                throw new RuntimeException("The append sparse vector must have a positive size.");
```
yunfengzhou-hub, line 156: nit: `"The appended sparse ..."`

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,168 @@
...
+                new VectorAssembler()
+                        .setInputCols("vec", "num", "sparseVec")
+                        .setOutputCol("assembledVec")
+                        .setHandleInvalid(HasHandleInvalid.ERROR_INVALID);
+        try {
+            Table outputTable = vectorAssembler.transform(inputDataTable)[0];
+            outputTable.execute().collect().next();
+            Assert.fail("Expected IllegalArgumentException");
+        } catch (Exception e) {
+            assertEquals(RuntimeException.class, ((Throwable) e).getClass());
```
yunfengzhou-hub, line 151: Can we check the detailed exception message? For example in `LogisticRegressionTest`
```java

            assertEquals(
                    "Multinomial classification is not supported yet. Supported options: [auto, binomial].",
                    e.getCause().getCause().getMessage());
```

yunfengzhou-hub (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,175 @@
...
+        tEnv = StreamTableEnvironment.create(env);
+        Schema schema =
+                Schema.newBuilder()
+                        .column("f0", DataTypes.INT())
+                        .column("f1", DataTypes.of(DenseVector.class))
+                        .column("f2", DataTypes.DOUBLE())
+                        .column("f3", DataTypes.of(SparseVector.class))
+                        .build();
+        DataStream<Row> dataStream = env.fromCollection(INPUT_DATA);
+        inputDataTable =
```
lindong28, line 94: We can just do the following without using `schema`.

```
inputDataTable = tEnv.fromDataStream(dataStream).as("id", "vec", "num", "sparseVec")
```

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,175 @@
...
+                            Vectors.sparse(5, new int[] {3}, new double[] {1.0})),
+                    Row.of(
+                            1,
+                            Vectors.dense(2.1, 3.1),
+                            1.0,
+                            Vectors.sparse(
+                                    5, new int[] {4, 2, 3, 1}, new double[] {4.0, 2.0, 3.0, 1.0})),
+                    Row.of(2, null, 1.0, null));
+
+    private static final SparseVector EXPECTED_DATA_1 =
```
lindong28, line 72: nits: To be more consistent with `INPUT_DATA`, how about renaming it as `EXPECTED_OUTPUT_DATA_1`?

lindong28 (commented):
Thanks for the update! Left some comments below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,183 @@
...
+            Vector assembledVector = null;
+            try {
+                assembledVector = assemble(objects);
+            } catch (Exception e) {
+                switch (handleInvalid) {
+                    case HasHandleInvalid.ERROR_INVALID:
+                        throw e;
+                    case HasHandleInvalid.SKIP_INVALID:
+                        return;
+                    case HasHandleInvalid.KEEP_INVALID:
```
lindong28, line 106: When `keep` is specified for this transformer, do we keep all types of invalidity (e.g. missing column, types that are neither number nor vector), or just a selected list of invalid types?

And what is the expected output of the transform(...) if it contains invalid input rows? For example, the output row's outputColumn is null?

If we plan to keep only a selected types of `invalidity`, as it affects the algorithm behavior, we probably need to explicitly specify it in the Java doc of this class.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,183 @@
...
+                        out.collect(Row.join(value, Row.of(assembledVector)));
+                        return;
+                    default:
```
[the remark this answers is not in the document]
weibozhao: I agree with throw new UnsupportOperationException

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasHandleInvalid.java
```diff
@@ -37,13 +37,14 @@
 public interface HasHandleInvalid<T> extends WithParams<T> {
     String ERROR_INVALID = "error";
     String SKIP_INVALID = "skip";
+    String KEEP_INVALID = "keep";
```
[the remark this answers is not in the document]
weibozhao: I get your idea. I think it's better defining this parameter in the VectorAssemblerParams, for keep has no meaning in OnehotEncoder fit and transform.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: Here，keep is just keep this input row values，the outputColumn has no meaning，so I give a null.

I have refine code for keep condition, including missing column, types err. and selected invalid types.

I add some doc for keep invalid.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -0,0 +1,163 @@
...
+                .setInputCols("vec", "num", "sparseVec")
+                .setOutputCol("assembledVec")
+                .setHandleInvalid(HasHandleInvalid.SKIP_INVALID);
+        assertArrayEquals(new String[] {"vec", "num", "sparseVec"}, vectorAssembler.getInputCols());
+        assertEquals(HasHandleInvalid.SKIP_INVALID, vectorAssembler.getHandleInvalid());
+        assertEquals("assembledVec", vectorAssembler.getOutputCol());
+    }
+
+    @Test
+    public void testTransform() throws Exception {
```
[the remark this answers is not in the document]
weibozhao: OK, I will refine code later.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
lindong28: Thanks for the update. Since the method here is an implementation detail of `VectorAssembler`, it is probably better to move the Java doc to the `HANDLE_INVALID` parameter in the `VectorAssemblerParams`. And we probably also need to explain the semantics of `error` and `skip`.

How about using the following Java doc for the `HANDLE_INVALID` parameter?

```
    /**
     * <p>Supported options and the corresponding behavior to handle invalid entries is listed as
     * follows.
     *
     * <ul>
     *   <li>error: raise an exception.
     *   <li>skip: filter out rows with bad values.
     *   <li>keep: output bad rows with output column's value set to null.
     * </ul>
     */
```

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasHandleInvalid.java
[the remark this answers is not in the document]
lindong28: Sounds good. Thanks for the update.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: I agree with it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -0,0 +1,187 @@
...
+                out.collect(Row.join(value, Row.of(assembledVector)));
+            } catch (Exception e) {
+                switch (handleInvalid) {
+                    case VectorAssemblerParams.ERROR_INVALID:
+                        throw e;
+                    case VectorAssemblerParams.SKIP_INVALID:
+                        return;
+                    case VectorAssemblerParams.KEEP_INVALID:
+                        /*
+                         * If KEEP_INVALID is set, all input row values will be reserved and a null
```
lindong28, line 106: Would it be simpler to remove this command? Or we can copy/paste the Java doc for "keep" from `VectorAssemblerParams`.

lindong28 (commented):
Thanks for the update! The PR looks pretty good now. Left just one minor comment.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: OK

PR #63, opened 2022-02-21
[FLINK-26263] Check data size in LogisticRegression

yunfengzhou-hub:

- Add check for data size in LogisticRegression

## Brief change log

- Add an if condition in LogisticRegression to check train data size, and add an early exit if the data size is 0.
- Add test case to test when parallelism is bigger than data size


zhipeng93 (commented):
Thanks for the PR. LGTM.

zhipeng93 (approved):

zhipeng93:
Merged.

PR #64, opened 2022-02-22
[FLINK-26266][ml][python] Support Vector and Matrix in ML Python API

HuangXingBo:


flink-ml-python/pyflink/ml/core/linalg.py
```diff
@@ -0,0 +1,544 @@
...
+    @abstractmethod
+    def num_rows(self) -> int:
+        pass
+
+    @abstractmethod
+    def num_cols(self) -> int:
+        pass
+
+    @abstractmethod
+    def get(self, i, j) -> float:
```
dianfu, line 482:
```suggestion
    def get(self, i: int, j: int) -> float:
```

flink-ml-python/pyflink/ml/core/linalg.py
```diff
@@ -0,0 +1,544 @@
...
+        self._num_cols = num_cols
+        self._values = np.asarray(values, dtype=np.float64)
+
+    def num_rows(self) -> int:
+        return self._num_rows
+
+    def num_cols(self) -> int:
+        return self._num_cols
+
+    def get(self, i, j) -> float:
```
dianfu, line 511: ditto

dianfu (approved):
@HuangXingBo Thanks a lot for the PR. LGTM. Just one minor comment. I guess you could fix it during merging the PR.

PR #66, opened 2022-02-24
[FLINK-26263] (followup) Check data size in LogisticRegression

zhipeng93:

- This is a followup of https://github.com/apache/flink-ml/pull/63

## Brief change log
- Put off the check of train data size and let each worker intanstiate variables even there is no input data.
- Get model dimension using parallelism 1.

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/AllReduceImpl.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/AllReduceImpl.java
@@ -37,7 +37,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Applies all-reduce on a data stream where each partition contains only one double array.
+ * Applies all-reduce on a data stream where each partition contains up to only one double array.
  *
  * <p>AllReduce is a communication primitive widely used in MPI. In this implementation, all workers
  * do reduce on a partition of the whole data and they all get the final reduce result. In detail,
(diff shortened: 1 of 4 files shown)


lindong28:
Thanks for PR! 

Could you confirm that the flaky test could be reproduced before this patch but not after this patch?

And could you update the AllReduceImpl's Java doc to replace `only one double array` with `up to one double array`? And maybe update allReduceSum() Java doc similarly?

zhipeng93:
Hi Dong, Thanks for the review.

> Could you confirm that the flaky test could be reproduced before this patch but not after this patch?

I have repeatedly run this flaky test for up to 100 times and there is no failures.
Do you mean that we need another test case for this bug fix? I think `LogisticRegressionTest#testMoreSubtaskThanData` already covers this case --- It sometimes fails before, but did not fail for 100 runs after this fix.
 
> And could you update the AllReduceImpl's Java doc to replace `only one double array` with `up to one double array`? And maybe update allReduceSum() Java doc similarly?

The java doc is updated and I also added one more test case to cover this.

lindong28:
@zhipeng93 It is generally preferred to have a test to reproduce this bug with high likelihood, or be able to reproduce this bug manually. This is generally done on a best-effort basis. The goal is to increase the confidence that we indeed fixed a bug with this PR.

If the flaky test is really hard to reproduce and there is no known way to cover the bug with a new test, we can still choose to merge the PR based on our understanding. This is just less than ideal.

lindong28:
Thanks for the update! LGTM.

PR #68, opened 2022-03-01
[FLINK-26404] support non-local file systems

mumuhhh:
[FLINK-26404](https://issues.apache.org/jira/browse/FLINK-26404) Flink-ML should support non-local file systems

### DIFF
--- a/flink-ml-core/pom.xml
+++ b/flink-ml-core/pom.xml
@@ -31,6 +31,10 @@ under the License.
   <artifactId>flink-ml-core_${scala.binary.version}</artifactId>
   <name>Flink ML : Core</name>
 
+  <properties>
+    <hadoop.version>2.4.1</hadoop.version>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.apache.flink</groupId>
@@ -107,6 +111,64 @@ under the License.
       <version>2.21.0</version>
       <type>jar</type>
       <scope>test</scope>
+    </dependency>
+
+    <!-- hdfs is required for the data cache test -->
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-hdfs</artifactId>
+      <scope>test</scope>
+      <version>${hadoop.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <scope>test</scope>
+      <version>${hadoop.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- This dependency is no longer shipped with the JDK since Java 9.-->
+          <groupId>jdk.tools</groupId>
+          <artifactId>jdk.tools</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-minicluster</artifactId>
+      <scope>test</scope>
+      <version>${hadoop.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- This dependency is no longer shipped with the JDK since Java 9.-->
+          <groupId>jdk.tools</groupId>
+          <artifactId>jdk.tools</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
   </dependencies>
 </project>
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/util/ReadWriteUtils.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/util/ReadWriteUtils.java
@@ -24,6 +24,8 @@ import org.apache.flink.api.connector.source.Source;
 import org.apache.flink.connector.file.sink.FileSink;
 import org.apache.flink.connector.file.src.FileSource;
 import org.apache.flink.connector.file.src.reader.SimpleStreamFormat;
+import org.apache.flink.core.fs.FileSystem;
+import org.apache.flink.core.fs.Path;
 import org.apache.flink.ml.api.Stage;
 import org.apache.flink.ml.builder.Graph;
 import org.apache.flink.ml.builder.GraphData;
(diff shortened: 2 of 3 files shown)


yunfengzhou-hub:
Hi @mumuhhh , Thanks for contributing to Flink ML. Could you please add more description about the motivation and the implementation principles for this PR? It would be even better if unit tests could be added to verify that Flink ML will be able to function correctly when saving to non-local file systems.

mumuhhh:
> Hi @mumuhhh , Thanks for contributing to Flink ML. Could you please add more description about the motivation and the implementation principles for this PR? It would be even better if unit tests could be added to verify that Flink ML will be able to function correctly when saving to non-local file systems.

ok

flink-ml-core/pom.xml
```diff
@@ -108,5 +112,83 @@ under the License.
...
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <!-- This dependency is no longer shipped with the JDK since Java 9.-->
+          <groupId>jdk.tools</groupId>
+          <artifactId>jdk.tools</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>    <dependency>
```
yunfengzhou-hub, line 187: nit: split them into different lines.

flink-ml-core/pom.xml
```diff
@@ -108,5 +112,83 @@ under the License.
...
+          <artifactId>jdk.tools</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-hdfs</artifactId>
+      <scope>test</scope>
+      <type>test-jar</type>
```
yunfengzhou-hub, line 156: I notices that you added two dependencies for `hadoop-hdfs` and `hadoop-common`, and the difference between them is this `<type>test-jar</type>`. Could you please illustrate why do we need two separate dependencies? I tried removing either of them, and the resulting project can still execute the unit tests.

flink-ml-core/src/main/java/org/apache/flink/ml/util/ReadWriteUtils.java
```diff
@@ -186,7 +174,7 @@ private static String getDataPath(String path) {
     private static String getPathForPipelineStage(int stageIdx, int numStages, String parentPath) {
         String format = String.format("%%0%dd", String.valueOf(numStages).length());
         String fileName = String.format(format, stageIdx);
-        return Paths.get(parentPath, "stages", fileName).toString();
+        return new Path(parentPath, new Path("stages", fileName)).toString();
```
yunfengzhou-hub, line 177: I understand that `Paths.get()` might not work in hdfs and thus should be replaced. Do we have an even better choice than `new Path(parentPath, new Path("stages", fileName))`? It seems that we would need N-1 `new Path()` in order to join N strings, which might not be good enough.

yunfengzhou-hub (commented):
Hi @mumuhhh, thanks for the update. It generally looks good to me except for a few comments as below.

Besides, could you please rebase your commits each time after you updated your PR, so that there is only one commit? Flink ML has been following this convention as it would make the commit history cleaner.

flink-ml-core/pom.xml
```diff
@@ -31,6 +31,10 @@ under the License.
   <artifactId>flink-ml-core_${scala.binary.version}</artifactId>
   <name>Flink ML : Core</name>
 
+  <properties>
+    <hadoop.version>2.4.1</hadoop.version>
```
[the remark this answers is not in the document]
mumuhhh: I referred to flink-ml-iteration’s pom.xml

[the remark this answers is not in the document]
yunfengzhou-hub: Sounds good. Let's keep Hadoop's version the same as flink-ml-iteration.

yunfengzhou-hub:
Thanks for the update. Could you please fix the problems revealed through the checks? You may reproduce the errors by running `mvn clean install` locally.

mumuhhh:
sorry,  I found it was the import problem；  Idea optimizes import and replaces import

yunfengzhou-hub:
It seems that `mvn clean install` still throws errors. Could you please help running this command locally to make the check?

mumuhhh:
There seems to be a problem with my local environment. I can only conduct some tests. I confirm that the problem is in the spotless check.  I fixed it

yunfengzhou-hub:
It seems that there still remains errors after spotless check is fixed. You can configure your forked Flink ML repo to run the Github Actions to help reveal the problems, if your local environment is unable to do it. For example, here is a check executed on one of my recent PRs.

https://github.com/yunfengzhou-hub/flink-ml/actions/workflows/java8-build.yml

mumuhhh:
I'm not sure what caused this problem
`The parameters 'rules' for goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M1:enforce are missing or invalid`

flink-ml-core/src/test/java/org/apache/flink/ml/util/ReadWriteUtilsTest.java
```diff
@@ -0,0 +1,117 @@
...
+        hdfsCluster = builder.build();
+    }
+
+    @After
+    public void after() {
+        hdfsCluster.shutdown();
+    }
+
+    @Test
+    public void testStageSaveLoad() throws IOException {
```
lindong28, line 80: Is the focus of this test to ensure that we can save/load metadata and model data from HDFS? If so, would it be simpler to just save and load a `SumModel` in this test?

`SumModel::save(...)` would invoke `ReadWriteUtils.saveModelData` and `ReadWriteUtils.saveMetadata`, which is exactly what we want to change and test in this PR.

And maybe remove `testModelSaveLoad` and the test described above is sufficient for this PR.

lindong28 (commented):
Thanks for @mumuhhh for the PR! And thanks @yunfengzhou-hub for the review.

LGTM overall. Left just one minor comment.

flink-ml-core/src/test/java/org/apache/flink/ml/util/ReadWriteUtilsTest.java
mumuhhh: That is a good idea

flink-ml-core/src/test/java/org/apache/flink/ml/util/ReadWriteUtilsTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @After
+    public void after() {
+        hdfsCluster.shutdown();
+    }
+
+    @Test
+    public void testModelSaveLoad() throws Exception {
+        // Builds a PipelineModel that increments input value by 60. This PipelineModel consists of
+        // three stages where each stage increments input value by 10, 20, and 30 respectively.
+        ExampleStages.SumModel modelA =
```
lindong28, line 83: Thanks for the update.

Would it be simpler to just save and load one SumModel instance, instead of saving/loading a pipeline of three SumModel instances here?

lindong28: Hmm... even if there is only one model, we can still set its model data, save it and load it, right? Do you mean there is something we can not cover by just saving/loading one `SumModel` instance?

mumuhhh: Let me revise it

lindong28:
Thanks for the update! LGTM.

PR #73, opened 2022-03-21
[FLINK-26626] Add Transformer and Estimator for StandardScaler

zhipeng93:
- Add Transformer and Estimator of StandardScaler in FlinkML. Standardization is a common requirement for machine learning training because they may behave badly if the individual features of a input do not look like standard normally distributed data (e.g. Gaussian with 0 mean and unit variance).

## Brief change log
- Add StandardScaler.
- Add StandardScalerModel.
- Add unit test for the proposed functionalities.

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
@@ -32,9 +32,31 @@ public class BLAS {
     }
 
     /** y += a * x . */
-    public static void axpy(double a, DenseVector x, DenseVector y) {
+    public static void axpy(double a, Vector x, DenseVector y) {
         Preconditions.checkArgument(x.size() == y.size(), "Vector size mismatched.");
-        JAVA_BLAS.daxpy(x.size(), a, x.values, 1, y.values, 1);
+        if (x instanceof SparseVector) {
+            axpy(a, (SparseVector) x, y);
+        } else {
+            axpy(a, (DenseVector) x, y);
+        }
+    }
+
+    /** Computes the hadamard product of the two vectors (y = y \hdot x). */
+    public static void hDot(Vector x, Vector y) {
+        Preconditions.checkArgument(x.size() == y.size(), "Vector size mismatched.");
+        if (x instanceof SparseVector) {
+            if (y instanceof SparseVector) {
+                hDot((SparseVector) x, (SparseVector) y);
+            } else {
+                hDot((SparseVector) x, (DenseVector) y);
+            }
+        } else {
+            if (y instanceof SparseVector) {
+                hDot((DenseVector) x, (SparseVector) y);
+            } else {
+                hDot((DenseVector) x, (DenseVector) y);
+            }
+        }
     }
 
     /** x \cdot y . */
@@ -88,5 +110,61 @@ public class BLAS {
                 beta,
                 y.values,
                 1);
+    }
+
+    private static void axpy(double a, DenseVector x, DenseVector y) {
+        JAVA_BLAS.daxpy(x.size(), a, x.values, 1, y.values, 1);
+    }
+
+    private static void axpy(double a, SparseVector x, DenseVector y) {
+        for (int i = 0; i < x.indices.length; i++) {
+            int index = x.indices[i];
+            y.values[index] += a * x.values[i];
+        }
+    }
+
+    private static void hDot(SparseVector x, SparseVector y) {
+        int idx = 0;
+        int idy = 0;
+        while (idx < x.indices.length && idy < y.indices.length) {
+            int indexX = x.indices[idx];
+            while (idy < y.indices.length && y.indices[idy] < indexX) {
+                y.values[idy] = 0;
+                idy++;
+            }
+            if (idy < y.indices.length && y.indices[idy] == indexX) {
+                y.values[idy] *= x.values[idx];
+                idy++;
+            }
+            idx++;
+        }
+        while (idy < y.indices.length) {
+            y.values[idy] = 0;
+            idy++;
+        }
+    }
+
+    private static void hDot(SparseVector x, DenseVector y) {
+        int idx = 0;
+        for (int i = 0; i < y.size(); i++) {
+            if (idx < x.indices.length && x.indices[idx] == i) {
+                y.values[i] *= x.values[idx];
+                idx++;
+            } else {
+                y.values[i] = 0;
+            }
+        }
+    }
+
+    private static void hDot(DenseVector x, SparseVector y) {
+        for (int i = 0; i < y.values.length; i++) {
+            y.values[i] *= x.values[y.indices[i]];
+        }
+    }
+
+    private static void hDot(DenseVector x, DenseVector y) {
+        for (int i = 0; i < x.values.length; i++) {
+            y.values[i] *= x.values[i];
+        }
     }
 }
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/DenseVector.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/DenseVector.java
@@ -68,4 +68,9 @@ public class DenseVector implements Vector {
     public int hashCode() {
         return Arrays.hashCode(values);
     }
+
+    @Override
+    public DenseVector clone() {
+        return new DenseVector(values.clone());
+    }
 }
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
@@ -84,7 +84,7 @@ public class StandardScaler
                                         TypeInformation.of(DenseVector.class),
                                         TypeInformation.of(DenseVector.class),
                                         BasicTypeInfo.LONG_TYPE_INFO),
-                                new ComputeMetaOperator(getFeaturesCol()));
+                                new ComputeMetaOperator(getInputCol()));
 
         DataStream<StandardScalerModelData> modelData =
                 sumAndSquaredSumAndWeight
@@ -206,10 +206,10 @@ public class StandardScaler
         private DenseVector squaredSum;
         private long numElements;
 
-        private final String featuresCol;
+        private final String inputCol;
 
-        public ComputeMetaOperator(String featuresCol) {
-            this.featuresCol = featuresCol;
+        public ComputeMetaOperator(String inputCol) {
+            this.inputCol = inputCol;
         }
 
         @Override
@@ -220,15 +220,15 @@ public class StandardScaler
         }
 
         @Override
-        public void processElement(StreamRecord<Row> element) throws Exception {
-            Vector feature = (Vector) element.getValue().getField(featuresCol);
+        public void processElement(StreamRecord<Row> element) {
+            Vector inputVec = (Vector) element.getValue().getField(inputCol);
             if (numElements == 0) {
-                sum = new DenseVector(feature.size());
-                squaredSum = new DenseVector(feature.size());
+                sum = new DenseVector(inputVec.size());
+                squaredSum = new DenseVector(inputVec.size());
             }
-            BLAS.axpy(1, feature, sum);
-            BLAS.hDot(feature, feature);
-            BLAS.axpy(1, feature, squaredSum);
+            BLAS.axpy(1, inputVec, sum);
+            BLAS.hDot(inputVec, inputVec);
+            BLAS.axpy(1, inputVec, squaredSum);
             numElements++;
         }
 
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
@@ -1,0 +1,288 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.standardscaler;
+
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.tuple.Tuple3;
+import org.apache.flink.api.java.typeutils.TupleTypeInfo;
+import org.apache.flink.iteration.operator.OperatorStateUtils;
+import org.apache.flink.ml.api.Estimator;
+import org.apache.flink.ml.linalg.BLAS;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.runtime.state.StateSnapshotContext;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
+import org.apache.flink.streaming.api.operators.BoundedOneInput;
+import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An Estimator which implements the standard scaling algorithm.
+ *
+ * <p>Standardization is a common requirement for machine learning training because they may behave
+ * badly if the individual features of a input do not look like standard normally distributed data
+ * (e.g. Gaussian with 0 mean and unit variance).
+ *
+ * <p>This estimator standardizes the input features by removing the mean and scaling each dimension
+ * to unit variance.
+ */
+public class StandardScaler
+        implements Estimator<StandardScaler, StandardScalerModel>,
+                StandardScalerParams<StandardScaler> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public StandardScaler() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public StandardScalerModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple3<DenseVector, DenseVector, Long>> sumAndSquaredSumAndWeight =
+                tEnv.toDataStream(inputs[0])
+                        .transform(
+                                "computeMeta",
+                                new TupleTypeInfo<>(
+                                        TypeInformation.of(DenseVector.class),
+                                        TypeInformation.of(DenseVector.class),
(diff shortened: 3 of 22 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
```diff
@@ -0,0 +1,290 @@
...
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple3<DenseVector, DenseVector, Long>> sumAndSquaredSumAndWeight =
+                tEnv.toDataStream(inputs[0])
+                        .transform(
+                                "computeMeta",
+                                new TupleTypeInfo<>(
+                                        TypeInformation.of(DenseVector.class),
+                                        TypeInformation.of(DenseVector.class),
+                                        BasicTypeInfo.LONG_TYPE_INFO),
+                                new ComputeMeta(getFeaturesCol()));
```
yunfengzhou-hub, line 86: How about using the following structure here?
```java
.keyBy(<random_key>)
.windowAll()
.aggregate()/.reduce()
```
The expected behavior is to accumulate computed results and output when all inputs are finished, which I think can be completed by the code above. The code above also preserves parallelism of your job. Besides, it would use higher abstraction-level Flink API so you need not care about some implementation details. Using a keyBy also allows you to use `ValueState` instead of `ListState` + `getUniqueElement`.

Same for the following part of code about `BuildModel`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
```diff
@@ -0,0 +1,290 @@
...
+        private DenseVector sum;
+        private DenseVector squaredSum;
+        private long numElements;
+
+        @Override
+        public void endInput() {
+            if (numElements > 0) {
+                BLAS.scal(1.0 / numElements, sum);
+                double[] mean = sum.values;
+                double[] std = squaredSum.values;
```
yunfengzhou-hub, line 121: nit: maybe `double[] std = new double[squaredSum.size()]` is better? Or just write the calculated results to `squaredSum.values`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
```diff
@@ -0,0 +1,185 @@
...
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/** A Model which transforms data using the model data computed by {@link StandardScaler}. */
+public class StandardScalerModel
```
yunfengzhou-hub, line 51: Do you think we can merge the model data and model operator of MinMaxScaler and Standard Scaler? I noticed that they are quite similar.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
```diff
@@ -0,0 +1,185 @@
...
+                    }
+                }
+            }
+
+            Vector feature = ((Vector) dataPoint.getField(featuresCol)).clone();
+
+            if (withMean && feature instanceof SparseVector) {
+                feature = Vectors.toDense((SparseVector) feature);
+            }
+            if (withMean) {
```
yunfengzhou-hub, line 137: If `withMean` or `withStd` is set to false, we may also be able to optimize the training process, as some values calculated during the training process is never used in prediction.

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -32,9 +32,31 @@ public static double asum(DenseVector x) {
...
-        JAVA_BLAS.daxpy(x.size(), a, x.values, 1, y.values, 1);
+        if (x instanceof SparseVector) {
+            axpy(a, (SparseVector) x, y);
+        } else {
+            axpy(a, (DenseVector) x, y);
+        }
+    }
+
+    /** Computes the hadamard product of the two vectors (y = y \hdot x). */
+    public static void hDot(Vector x, Vector y) {
```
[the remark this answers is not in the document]
zhipeng93: Good question! There has also been a discussion of the reason that BLAS does not provide a hadamard dot [1] --- Mostly saying that there are no performance gains.

However, I'd like to say that introducing `hDot` here is simply for convienice. I need to use it twice. 

By the way, we can replace the logic in MinMaxScaler if we have this method. This will certainly simplify the implementation.

What do you think?

[1] https://github.com/xianyi/OpenBLAS/issues/1083

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/DenseVector.java
```diff
@@ -68,4 +68,9 @@ public boolean equals(Object obj) {
     public int hashCode() {
         return Arrays.hashCode(values);
     }
+
+    @Override
+    public Vector clone() {
```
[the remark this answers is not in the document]
zhipeng93: Thanks for pointing this out. The `clone()` here is to create a new instance of this Vector for further in-place operations, in which case we may not want to update the vectors directly.

I think there could be two alternatives:
- How about we rename the method to `duplicate` or something like `copy`? 
- Or how about we remove the `clone()` in `Vector`, but keep it in `SparseVector` and `DenseVector`, given that we are indeed cloning vectors.

What do you think?

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/Vectors.java
```diff
@@ -30,4 +30,9 @@ public static DenseVector dense(double... values) {
     public static SparseVector sparse(int size, int[] indices, double[] values) {
         return new SparseVector(size, indices, values);
     }
+
+    /** Converts a sparse vector to a dense vector. */
+    public static DenseVector toDense(SparseVector vec) {
```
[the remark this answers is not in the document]
zhipeng93: Thanks for the comment. Let's follow the best practice :~

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
```diff
@@ -0,0 +1,185 @@
...
+                    for (int i = 0; i < scaleValues.length; i++) {
+                        scaleValues[i] = scaleValues[i] == 0 ? 0 : 1 / scaleValues[i];
+                    }
+                }
+            }
+
+            Vector feature = ((Vector) dataPoint.getField(featuresCol)).clone();
+
+            if (withMean && feature instanceof SparseVector) {
+                feature = Vectors.toDense((SparseVector) feature);
```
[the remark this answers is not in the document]
zhipeng93: `toDense` is not used for performance here...

It is because we can only support in-place adding a `SparseVector/DenseVector` to a `DenseVecor` in `axpy`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
zhipeng93: Good point. My concern is that using`windowAll/window` may lead to a large memory consumption. As I understand, window operations need to cache the data in state. (Please correct me if I am wrong...

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
zhipeng93: They are indeed similar in -- both of them are composed of two vectors. 
I am okay to merge them as one. My Concern is that what are we going to name the new class? Do you have some suggestions?

zhipeng93: They are indeed similar in -- both of them are composed of two vectors. 
I am okay to merge them as one. My Concern is that what are we going to name the new class? Do you have some suggestions?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
zhipeng93: Using `double[] std = new double[squaredSum.size()]` needs to allocate memory for the new matrix. If the dimension of the input is high, it is worse.

We are actually writing the calculated results back to `squaredSum.values`.  `std` here is a reference for better understanding.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
zhipeng93: Good point. I also thought about this. But I didn't do the optimization here because of the following use case:

A user calls StandardScaler#fit and get a StandardScalerModel, using `withStd as false`. Then the user wants to do a inference with the trained model using `withStd as true`. In this case, it fails as there is no std in the model data.

Another reason is that computing `std` and `mean` is not that expensive comparing with IO...

What do you think?

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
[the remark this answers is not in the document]
yunfengzhou-hub: I think it would be a good idea to replace the logic in MinMaxScaler. But Weibo was concerned about the performance, as the implementation would be changed from something like 
```java
for (int i = 0; i < n; i++) {
  y[i] += a[i] * x[i];
}
```
to 
```java
for (int i = 0; i < n; i++) {
  x[i] *= a[i];
}
for (int i = 0; i < n; i++) {
  y[i] += x[i];
}
```
Thus I would like to hear from Weibo's opinion about this replacement.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/DenseVector.java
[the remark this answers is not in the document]
yunfengzhou-hub: I think the second one would be better, as it reuses existing API and is more straight forward.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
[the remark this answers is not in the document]
yunfengzhou-hub: Sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
yunfengzhou-hub: Got it. Then I'm good with the current implementation.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
yunfengzhou-hub: That's a good point. A trained Model might be used in either `withStd = true` or `withStd = false` situations. I agree with the current implementation.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
yunfengzhou-hub: I think we need more offline discussion about this API. Another choice is to still keep `MinMaxScalerModel` and `StandardScalerModel`, but they actually extend the same parent class without overriding any one of the parent class's methods.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
yunfengzhou-hub: I think window operation would only cache data if the following operator is `apply`. In cases when the following is `aggregate/reduce`, the window operation processes a stream record as soon as it is received, so it would only need to cache the accumulator / reduce result.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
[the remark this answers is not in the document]
zhipeng93: I talked with Weibo. He is fine with the change and also thinks it is better to keep the code consistent.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
zhipeng93: Thanks for the detail explanation. When using `keyBy().window().aggregate()`, I came into another problem [1]. It seems that it is not solved yet. So let's use the current implementation for now.

[1] https://lists.apache.org/thread/7jm3sdh2h3y2gks3rp8kn9yb44b55dzv

yunfengzhou-hub (commented):
Thanks for the update. Generally LGTM. Left one small comment as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
```diff
@@ -0,0 +1,188 @@
...
+                    double[] scaleValues = scale.values;
+                    for (int i = 0; i < scaleValues.length; i++) {
+                        scaleValues[i] = scaleValues[i] == 0 ? 0 : 1 / scaleValues[i];
+                    }
+                }
+            }
+
+            Vector feature = (Vector) (dataPoint.getField(featuresCol));
+            Vector output;
+            if (feature instanceof DenseVector) {
```
lindong28, line 133: Would it be simpler to add the method `Vector::clone()` similar to Spark's `Vector::copy()`?

lindong28 (commented):
Thanks for the PR. Looks good overall. Left some minor comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerModel.java
zhipeng93: As we discussed offline, we agree to keep the code for now because it is an implementation issue and do not affect end users.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -89,4 +111,56 @@ public static void gemv(
...
+                idy++;
+            }
+            idx++;
+        }
+    }
+
+    private static void hDot(SparseVector x, DenseVector y) {
+        int idx = 0;
+        for (int i = 0; i < y.size(); i++) {
+            if (x.indices[idx] == i) {
```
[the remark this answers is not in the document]
zhipeng93: Thanks for pointing this out. It is fixed.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
```diff
@@ -0,0 +1,288 @@
...
+
+        private final String featuresCol;
+
+        public ComputeMetaOperator(String featuresCol) {
+            this.featuresCol = featuresCol;
+        }
+
+        @Override
+        public void endInput() {
+            if (numElements > 0) {
```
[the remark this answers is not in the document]
zhipeng93: It throws an exception here if we remove it when there is no data on this partition.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
```diff
@@ -0,0 +1,288 @@
...
+                            .orElse(null);
+            numElements =
+                    OperatorStateUtils.getUniqueElement(numElementsState, "numElementsState")
+                            .orElse(0L);
+        }
+
+        @Override
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            if (numElements > 0) {
```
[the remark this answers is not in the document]
zhipeng93: Good point. 

I agree with the behavior that when the train data is empty, we should output a `special` vector to indicate that we cannot do inference. I have updated the code to handle this case and also added one test case. (See for StandardScalerTest#testFitOnEmptyData).

For other checks about when numElements is zero or not, we probably need to keep it. 

For example, when the training data is not empty, there could be one worker that contains zero data points. If we remove the check, it throws an nullpointer exception.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -32,9 +32,31 @@ public static double asum(DenseVector x) {
...
+        } else {
+            axpy(a, (DenseVector) x, y);
+        }
+    }
+
+    /** Computes the hadamard product of the two vectors (y = y \hdot x). */
+    public static void hDot(Vector x, Vector y) {
+        Preconditions.checkArgument(x.size() == y.size(), "Vector size mismatched.");
+        if (y instanceof DenseVector) {
+            if (x instanceof SparseVector) {
```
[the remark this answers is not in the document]
zhipeng93: Thanks for the comments. I have reordered the order of `a instanceof b`.

BTW, we do not support `axpy` when `y` is a SparseVector because it may lead to some memory re-allocation.

zhipeng93:
@lindong28 Thanks for the review. I have adressed your comments in the latest commit.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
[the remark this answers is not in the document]
lindong28: Sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScaler.java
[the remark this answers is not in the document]
lindong28: After thinking about this more, I think it might be better to let `estimator.fit(...)` if input data is empty during runtime, so that the error could be moved closer to where it happens, i.e. in the operator that does training instead of the operator that does inference. And this could also make the logic in the model simpler and consistent with other existing models.

What do you think?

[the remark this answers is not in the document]
zhipeng93: Thanks for pointint out that "The error should be moved closer to where it happens" and I also believe it is reasonable to throw exception when the training set is empty.

I have addressed this in the latest commit.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerParams.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.flink.ml.common.param.HasPredictionCol;
+import org.apache.flink.ml.param.BooleanParam;
+import org.apache.flink.ml.param.Param;
+
+/**
+ * Params for {@link StandardScaler}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface StandardScalerParams<T> extends HasFeaturesCol<T>, HasPredictionCol<T> {
```
lindong28, line 31: Spark's `StandardScaler` uses `inputCol` and `outputCol` as its parameters. Would it be better to follow the same approach here?

Note that this `StandardScaler` is used to transform features rather than making predictions. Thus it does not seem intuitive to name its output as `predictionCol`. We can submit a followup PR to fix this problem.

`HasFeaturesCol` should probably be renamed to `HasFeatureCol` as it represents just one column.

And `HasInputCol`  is more consistent with `HasOutputCol` than `HasFeatureCol` does.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/StandardScalerTest.java
```diff
@@ -0,0 +1,279 @@
...
+                .setPredictionCol("test_prediction");
+
+        assertEquals("test_features", standardScaler.getFeaturesCol());
+        assertEquals(true, standardScaler.getWithMean());
+        assertEquals(false, standardScaler.getWithStd());
+        assertEquals("test_prediction", standardScaler.getPredictionCol());
+    }
+
+    @Test
+    public void testFeaturePredictionParam() {
```
lindong28, line 152: nits: suppose we change the parameter to inputCol and outputCol, it might not be intuitive to name this method as `testFeaturePredictionParam`. How about changing it to `testOutputSchema()`? We can use this name consistently across other tests.

lindong28 (commented):
Thanks for the update. Looks good overall. Left two minor comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/StandardScalerParams.java
zhipeng93: Thanks for raising up this discussion. 

As `StandardScaler` is an `estmator`, it aims to train on the input features and predict on the output. So I thought it might be ok to keep the current naming. So soes `MinMaxScaler`.

> HasFeaturesCol should probably be renamed to HasFeatureCol as it represents just one column.
Here `features` is one column.

zhipeng93: After some offline discussion, we agree to follow the following rules:
- use `inputCol/outputCol/inputCols/outputCos` when doing feature engineering. (e.g., code under package `org.apache.flink.ml.feature`)
- use `featuresCol/predictionCol` when doing (un)supervised machine learning training (e.g., code under package `org.apache.flink.ml.classification, org.apache.flink.ml.clustering`

lindong28:
Thanks for the update! LGTM.

PR #82, opened 2022-04-07
[FLINK-27072] Add Transformer for Bucketizer

zhipeng93:
- Add Transformer of Bucketizer in FlinkML. 

## Brief change log
- Reorganized the HasHandleInvalid param in StringIndexer, OnehotEncoder and VectorAssembler.
- Add Bucketizer.
- Add unit test for Bucketizer

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
@@ -16,16 +16,27 @@
  * limitations under the License.
  */
 
-package org.apache.flink.ml.feature.stringindexer;
+package org.apache.flink.ml.param;
 
-import org.apache.flink.ml.common.param.HasHandleInvalid;
-import org.apache.flink.ml.common.param.HasInputCols;
-import org.apache.flink.ml.common.param.HasOutputCols;
+import org.apache.flink.ml.util.ReadWriteUtils;
 
-/**
- * Params of {@link StringIndexerModel}.
- *
- * @param <T> The class type of this instance.
- */
-public interface StringIndexerModelParams<T>
-        extends HasInputCols<T>, HasOutputCols<T>, HasHandleInvalid<T> {}
+import java.io.IOException;
+
+/** Class for the array parameter. */
+public class ArrayArrayParam<T> extends Param<T[][]> {
+
+    public ArrayArrayParam(
+            String name,
+            Class<T[][]> clazz,
+            String description,
+            T[][] defaultValue,
+            ParamValidator<T[][]> validator) {
+        super(name, clazz, description, defaultValue, validator);
+    }
+
+    @Override
+    public T[][] jsonDecode(Object json) throws IOException {
+        String jsonStr = ReadWriteUtils.OBJECT_MAPPER.writeValueAsString(json);
+        return ReadWriteUtils.OBJECT_MAPPER.readValue(jsonStr, clazz);
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.bucketizer;
+
+import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.common.param.HasHandleInvalid;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Bucketizer is a transformer that maps multiple columns of continuous features to multiple columns
+ * of discrete features, i.e., buckets indices. The indices are in [0, numSplitsInThisColumn - 1].
+ *
+ * <p>The `keep` option of {@link HasHandleInvalid} means that we put the invalid data in the last
+ * bucket of the splits, whose index is the number of the buckets.
+ */
+public class Bucketizer implements Transformer<Bucketizer>, BucketizerParams<Bucketizer> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public Bucketizer() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        String[] inputCols = getInputCols();
+        String[] outputCols = getOutputCols();
+        Double[][] splitsArray = getSplitsArray();
+        Preconditions.checkArgument(inputCols.length == outputCols.length);
+        Preconditions.checkArgument(inputCols.length == splitsArray.length);
+
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        TypeInformation<?>[] outputTypes = new TypeInformation[outputCols.length];
+        Arrays.fill(outputTypes, BasicTypeInfo.INT_TYPE_INFO);
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), outputTypes),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCols()));
+
+        DataStream<Row> result =
+                tEnv.toDataStream(inputs[0])
+                        .flatMap(
+                                new FindBucketFunction(inputCols, splitsArray, getHandleInvalid()),
+                                outputTypeInfo);
+        return new Table[] {tEnv.fromDataStream(result)};
+    }
+
+    /** Finds the bucket index for each continuous feature of an input data point. */
+    private static class FindBucketFunction implements FlatMapFunction<Row, Row> {
+        private final String[] inputCols;
+        private final String handleInvalid;
+        private final Double[][] splitsArray;
+
+        public FindBucketFunction(
+                String[] inputCols, Double[][] splitsArray, String handleInvalid) {
+            this.inputCols = inputCols;
+            this.splitsArray = splitsArray;
+            this.handleInvalid = handleInvalid;
+        }
+
+        @Override
+        public void flatMap(Row value, Collector<Row> out) {
+            Row outputRow = new Row(inputCols.length);
+
+            for (int i = 0; i < inputCols.length; i++) {
+                double feature = ((Number) value.getField(inputCols[i])).doubleValue();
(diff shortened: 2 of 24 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
```diff
@@ -0,0 +1,179 @@
...
+            for (int j = 1; j < splits.length; j++) {
+                Preconditions.checkArgument(
+                        splits[j] > splits[j - 1],
+                        "Illegal value for "
+                                + BucketizerParams.SPLITS_ARRAY
+                                + ". See param "
+                                + BucketizerParams.SPLITS_ARRAY
+                                + " for details.");
+            }
+        }
```
yunfengzhou-hub, line 82: Shall we move this part of logic to a `ParamValidator` class? This class can be placed inside `BucketizerParams`.

Moreover, We would always need to do some checks in ParamValidator, and to do the other checks in `fit()`/`transform()`. I think we should place a validation in ParamValidator if it is generally applicable to all subclasses that inherit this parameter. If the validation needs to be computed across parameters, or depends on the input Tables, then it should be placed in `fit()`/`transform()`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
```diff
@@ -0,0 +1,179 @@
...
+            this.splitsArray = splitsArray;
+            this.handleInvalid = handleInvalid;
+        }
+
+        @Override
+        public void flatMap(Row value, Collector<Row> out) {
+            Row outputRow = new Row(inputCols.length);
+
+            for (int i = 0; i < inputCols.length; i++) {
+                double feature = ((Number) value.getField(inputCols[i])).doubleValue();
```
yunfengzhou-hub, line 121: If the field value is not a Number, the operator would throw exception, which means that `HasHandleInvalid` is always `ERROR` in this case.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
```diff
@@ -0,0 +1,179 @@
...
+                        case BucketizerParams.ERROR_INVALID:
+                            throw new RuntimeException(
+                                    "The input contains invalid value. See "
+                                            + BucketizerParams.HANDLE_INVALID
+                                            + " parameter for more options.");
+                        case BucketizerParams.SKIP_INVALID:
+                            return;
+                        case BucketizerParams.KEEP_INVALID:
+                            outputRow.setField(i, splits.length - 1);
+                            break;
```
yunfengzhou-hub, line 155: This should be `continue`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
```diff
@@ -0,0 +1,179 @@
...
+                        default:
+                            throw new IllegalStateException(
+                                    "Unsupported handleInvalid type: " + handleInvalid);
```
yunfengzhou-hub, line 158: Given that we used `BucketizerParams.SPLITS_ARRAY` in other error messages, shall we also use `BucketizerParams.HANDLE_INVALID` here?

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.flink.ml.param;
+
+import org.apache.flink.ml.util.ReadWriteUtils;
+
+import java.io.IOException;
+
+/** Class for the arrayOfArray-type parameters. */
+public class ArrayArrayParam<T> extends Param<T[][]> {
```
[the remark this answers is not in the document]
zhipeng93: We probably follow the user behavior here. 

In Bucketizer, users specify a split array for each input col. Moreover, the number of split points in different columns could be different. If we use 1D array to represent 2D array, we would also need to specify the dimension for each 1D array. This seems a bit complex to me.

Also, using 2D array is consistent with exisiting libraries [1].

[1] https://github.com/apache/spark/blob/1018f8b545a74f4fcaf123b5f15d52ac2afa2d3c/mllib/src/main/scala/org/apache/spark/ml/param/params.scala#L567

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/onehotencoder/OneHotEncoderParams.java
```diff
@@ -27,6 +27,8 @@
 /**
  * Params of OneHotEncoderModel.
  *
+ * <p>The `keep` option of {@link HasHandleInvalid} is not supported in {@link OneHotEncoderParams}.
```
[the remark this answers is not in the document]
zhipeng93: Yep. I added the comment here because we would support `SKIP` in the near future (I guess), but will never support `keep`.

Are you suggesting that we also mention that `SKIP` is not supported and remove it later in the near future?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/OneHotEncoderTest.java
```diff
@@ -205,7 +205,7 @@ public void testInputDataType() {
 
     @Test
     public void testNotSupportedHandleInvalidOptions() {
-        estimator.setHandleInvalid(HasHandleInvalid.SKIP_INVALID);
+        estimator.setHandleInvalid(OneHotEncoderParams.SKIP_INVALID);
```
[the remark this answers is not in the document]
zhipeng93: Thanks for pointing this out. I also think using `HasHandleInvalid` would be better.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
zhipeng93: `Bucketizer` has assumed that the input is always a `Number`. I think the default thrown exception is enough for debugging.

The `ERROR` is for two cases:
- `NaN` 
- The input values that do not fit in the given split arrays.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
zhipeng93: Hmm, could you explain a bit more about this?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
zhipeng93: Good point. I have move the checks to `BucketizerParams`.

flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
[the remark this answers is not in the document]
yunfengzhou-hub: I agree with it that on API we need to display getter/setter for 2D arrays, but as for internal implementation, I believe we can store the 2D array in a 1D format. This does not affect user experience and can reuse existing `ArrayParams`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/onehotencoder/OneHotEncoderParams.java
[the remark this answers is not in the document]
yunfengzhou-hub: I think it could be better if we make the JavaDoc correspond to the actual behavior. We can modify the JavaDoc when we add support for `SKIP` in future. Thus it might be better to comment that `OneHotEncoder` only supports `ERROR` in its JavaDoc for now.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
yunfengzhou-hub: For example, If the input have three columns, `[a, b, c]`. `a` and `c` are valid values mapping to `a*` and `c*`, while `b` is an invalid value. In this case the operator should output `[a*, splits.length-1, c*]`, but if the code here is `break`,  the operator would actually output `[a*,  splits.length-1, null]`.

flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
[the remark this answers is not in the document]
zhipeng93: > we can store the 2D array in a 1D format

I am a bit confused about this. How are we going to store the shape of each dimension? 
Maintaining another `array` for shapes seems a bit complex to me.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/onehotencoder/OneHotEncoderParams.java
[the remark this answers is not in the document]
zhipeng93: Sounds good.

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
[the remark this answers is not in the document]
yunfengzhou-hub: How about this?
```java
    DoubleArrayParam SPLITS_ARRAY =
            new DoubleArrayParam(
                    "splitsArray", "Array of split points.", null, ParamValidators.nonEmptyArray());

    default Double[][] getSplitsArray() {
        Double[] values = get(SPLITS_ARRAY);
        Double[][] result = new Double[values[0].intValue()][];
        int offset = 1;
        for (int i = 0; i < values[0].intValue(); i++) {
            result[i] = new Double[values[offset].intValue()];
            System.arraycopy(values, offset + 1, result[i], 0, values[offset].intValue());
            offset += values[0].intValue() + 1;
        }
        return result;
    }

    default T setSplitsArray(Double[][] value) {
        List<Double> list = new ArrayList<>();
        list.add((double) value.length);
        for (Double[] doubles : value) {
            list.add((double) doubles.length);
            list.addAll(Arrays.asList(doubles));
        }
        set(SPLITS_ARRAY, list.toArray(new Double[0]));
        return (T) this;
    }
```

My point is that we should try to simplify the internal implementation. Currently we have about 6 basic type params, including String, Boolean, Int, Long, Double, Float. For array-typed parameters, we introduced another 6 classes. For 2D-array typed parameters, we would also need 6 classes. Same for 3D, 4D cases. I don't think it would be a good idea to have so many classes for parameters, each of which shares a large proportion in common. For this problem, the solution I proposed in comments above is to reuse 1D classes for arrays of larger dimensions.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/BucketizerTest.java
```diff
@@ -0,0 +1,190 @@
...
+    @Test
+    public void testParam() {
+        Bucketizer bucketizer = new Bucketizer();
+        assertEquals(HasHandleInvalid.ERROR_INVALID, bucketizer.getHandleInvalid());
+
+        bucketizer
+                .setInputCols("f1", "f2", "f3")
+                .setOutputCols("o1", "o2", "o3")
+                .setHandleInvalid(HasHandleInvalid.SKIP_INVALID)
+                .setSplitsArray(splitsArray);
```
yunfengzhou-hub, line 115: @lindong28 How do you like the idea to add APIs to `Bucketizer` so that users can define metadata in the following format?
```java
bucketizer
    .addCol("f1", "o1", splitsArray[0])
    .addCol("f2", "o2", splitsArray[1]);
```
It is different from existing practice from Spark or Alink, but the readability would better. It would also be OK for me if it is better to leave this discussion to future and to focus on functionality correctness for now.

yunfengzhou-hub (commented):
Thanks for the updates. The PR generally looks good to me except that we should rebase the commit history.

@lindong28 Could you please help review this PR? Besides, I left a comment as below and would like to hear your idea about it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/bucketizer/Bucketizer.java
lindong28: @yunfengzhou-hub by the `break` here, the code breaks out of the switch loop, and the for loop continues. So in the case described above, the current implementation would still output `[a*, splits.length-1, c*]`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/BucketizerTest.java
lindong28: It seems better keep the current approach so that the user-experience with `setInputCols(..)` names is more consistent with `setInputCol(...)`.

IMO, it is not clear that the readability with the alternative approach is better than the current approach.

lindong28 (commented):
Thanks for the PR! Looks good overall. Left some minor comments.

flink-ml-core/src/main/java/org/apache/flink/ml/param/ArrayArrayParam.java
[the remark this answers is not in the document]
lindong28: I am not sure the alternative approach simplify the overall internal implementation.

The alternative approach does remove the need for DoubleArrayArrayParam. But for every parameter of type `double[][]`, it requires the setXXX() and getXXX() to have non-trivial implementation as describe above. We can add two public static methods to be shared by those setXXX() and getXXX(). But the addition of these two methods do not seem to be simpler than the existing `DoubleArrayArrayParam`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/onehotencoder/OneHotEncoderParams.java
```diff
@@ -27,6 +27,9 @@
 /**
  * Params of OneHotEncoderModel.
  *
+ * <p>The `keep` and `skip` option of {@link HasHandleInvalid} is not supported in {@link
```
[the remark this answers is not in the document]
zhipeng93: Thanks for the comment. I prefer putting the description in XXXParam class for the following two reaons:

- Putting the description to concrete classes may lead to duplicate comments. For example, `OneHotEncoder` and `OneHotEncoderModel`.
- I don't really think checking Java doc for XXXParams is a bad/not-user-friendly behavior. They are public interfaces by the way.

What do you think?

lindong28:
Thanks for the update! LGTM.

lindong28:
Merged to the master branch.

PR #83, opened 2022-04-11
[FLINK-27170] Add Transformer and Estimator for OnlineLogisticRegression

weibozhao:
Add Transformer and Estimator of OnlineLogisticRegression

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
@@ -19,21 +19,32 @@
 package org.apache.flink.ml.common.datastream;
 
 import org.apache.flink.annotation.Internal;
+import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.common.functions.MapFunction;
 import org.apache.flink.api.common.functions.MapPartitionFunction;
 import org.apache.flink.api.common.functions.ReduceFunction;
 import org.apache.flink.api.common.state.ListState;
 import org.apache.flink.api.common.state.ListStateDescriptor;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.api.java.typeutils.TypeExtractor;
 import org.apache.flink.iteration.operator.OperatorStateUtils;
 import org.apache.flink.runtime.state.StateInitializationContext;
 import org.apache.flink.runtime.state.StateSnapshotContext;
 import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.functions.windowing.AllWindowFunction;
 import org.apache.flink.streaming.api.operators.AbstractUdfStreamOperator;
 import org.apache.flink.streaming.api.operators.BoundedOneInput;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.api.operators.TimestampedCollector;
+import org.apache.flink.streaming.api.windowing.windows.GlobalWindow;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.util.Collector;
+
+import org.apache.commons.collections.IteratorUtils;
+
+import java.util.Arrays;
+import java.util.List;
 
 /** Provides utility functions for {@link DataStream}. */
 @Internal
@@ -179,6 +190,69 @@ public class DataStreamUtils {
             state.clear();
             if (result != null) {
                 state.add(result);
+            }
+        }
+    }
+
+    /**
+     * Splits the input data into global batches of batchSize. After splitting, each global batch is
+     * further split into local batches for downstream operators with each worker has one batch.
+     */
+    public static <T> DataStream<T[]> generateBatchData(
+            DataStream<T> inputData, final int downStreamParallelism, int batchSize) {
+        return inputData
+                .countWindowAll(batchSize)
+                .apply(new GlobalBatchCreator<>())
+                .flatMap(new GlobalBatchSplitter<>(downStreamParallelism))
+                .partitionCustom((chunkId, numPartitions) -> chunkId, x -> x.f0)
+                .map(
+                        new MapFunction<Tuple2<Integer, T[]>, T[]>() {
+                            @Override
+                            public T[] map(Tuple2<Integer, T[]> integerTuple2) throws Exception {
+                                return integerTuple2.f1;
+                            }
+                        });
+    }
+
+    /** Splits the input data into global batches. */
+    private static class GlobalBatchCreator<T> implements AllWindowFunction<T, T[], GlobalWindow> {
+        @Override
+        public void apply(GlobalWindow timeWindow, Iterable<T> iterable, Collector<T[]> collector) {
+            List<T> points = IteratorUtils.toList(iterable.iterator());
+            collector.collect(points.toArray((T[]) new Object[0]));
+        }
+    }
+
+    /**
+     * An operator that splits a global batch into evenly-sized local batches, and distributes them
+     * to downstream operator.
+     */
+    private static class GlobalBatchSplitter<T>
+            implements FlatMapFunction<T[], Tuple2<Integer, T[]>> {
+        private final int downStreamParallelism;
+
+        public GlobalBatchSplitter(int downStreamParallelism) {
+            this.downStreamParallelism = downStreamParallelism;
+        }
+
+        @Override
+        public void flatMap(T[] values, Collector<Tuple2<Integer, T[]>> collector) {
+            int div = values.length / downStreamParallelism;
+            int mod = values.length % downStreamParallelism;
+
+            int offset = 0;
+            int i = 0;
+
+            int size = div + 1;
+            for (; i < mod; i++) {
+                collector.collect(Tuple2.of(i, Arrays.copyOfRange(values, offset, offset + size)));
+                offset += size;
+            }
+
+            size = div;
+            for (; i < downStreamParallelism; i++) {
+                collector.collect(Tuple2.of(i, Arrays.copyOfRange(values, offset, offset + size)));
+                offset += size;
             }
         }
     }
--- a/flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
+++ b/flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
@@ -37,6 +37,7 @@ import org.junit.Test;
 import java.util.List;
 
 import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 /** Tests the {@link DataStreamUtils}. */
@@ -72,6 +73,18 @@ public class DataStreamUtilsTest {
                 DataStreamUtils.reduce(dataStream, (ReduceFunction<Long>) Long::sum);
         List<Long> sum = IteratorUtils.toList(result.executeAndCollect());
         assertArrayEquals(new long[] {190L}, sum.stream().mapToLong(Long::longValue).toArray());
+    }
+
+    @Test
+    public void testGenerateBatchData() throws Exception {
+        DataStream<Long> dataStream =
+                env.fromParallelCollection(new NumberSequenceIterator(0L, 19L), Types.LONG);
+        DataStream<Long[]> result = DataStreamUtils.generateBatchData(dataStream, 2, 4);
+        List<Long[]> batches = IteratorUtils.toList(result.executeAndCollect());
+        for (Long[] batch : batches) {
+            assertEquals(2, batch.length);
+        }
+        assertEquals(10, batches.size());
     }
 
     /** A simple implementation for a {@link MapPartitionFunction}. */
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
@@ -115,7 +115,7 @@ public class LogisticRegression
                 optimizer.optimize(initModelData, trainData, BinaryLogisticLoss.INSTANCE);
 
         DataStream<LogisticRegressionModelData> modelData =
-                rawModelData.map(LogisticRegressionModelData::new);
+                rawModelData.map(vector -> new LogisticRegressionModelData(vector, 0));
         LogisticRegressionModel model =
                 new LogisticRegressionModel().setModelData(tEnv.fromDataStream(modelData));
         ReadWriteUtils.updateExistingParams(model, paramMap);
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelData.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelData.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.ml.classification.logisticregression;
 
+import org.apache.flink.api.common.functions.MapFunction;
 import org.apache.flink.api.common.serialization.Encoder;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.configuration.Configuration;
@@ -25,10 +26,11 @@ import org.apache.flink.connector.file.src.reader.SimpleStreamFormat;
 import org.apache.flink.core.fs.FSDataInputStream;
 import org.apache.flink.core.memory.DataInputViewStreamWrapper;
 import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
+import org.apache.flink.ml.common.datastream.TableUtils;
 import org.apache.flink.ml.linalg.DenseVector;
-import org.apache.flink.ml.linalg.Vector;
 import org.apache.flink.ml.linalg.typeinfo.DenseVectorSerializer;
 import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.Table;
 import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
 import org.apache.flink.table.api.internal.TableImpl;
@@ -36,9 +38,10 @@ import org.apache.flink.table.api.internal.TableImpl;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.util.Random;
 
 /**
- * Model data of {@link LogisticRegressionModel}.
+ * Model data of {@link LogisticRegressionModel} and {@link OnlineLogisticRegressionModel}.
  *
  * <p>This class also provides methods to convert model data from Table to Datastream, and classes
  * to save/load model data.
@@ -46,12 +49,49 @@ import java.io.OutputStream;
 public class LogisticRegressionModelData {
 
     public DenseVector coefficient;
-
-    public LogisticRegressionModelData(DenseVector coefficient) {
-        this.coefficient = coefficient;
-    }
+    public long modelVersion;
 
     public LogisticRegressionModelData() {}
+
+    public LogisticRegressionModelData(DenseVector coefficient, long modelVersion) {
+        this.coefficient = coefficient;
+        this.modelVersion = modelVersion;
+    }
+
+    /**
+     * Generates a Table containing a {@link LogisticRegressionModelData} instance with randomly
+     * generated coefficient.
+     *
+     * @param tEnv The environment where to create the table.
+     * @param dim The size of generated coefficient.
+     * @param seed Random seed.
+     */
+    public static Table generateRandomModelData(StreamTableEnvironment tEnv, int dim, int seed) {
+        StreamExecutionEnvironment env = TableUtils.getExecutionEnvironment(tEnv);
+        return tEnv.fromDataStream(
+                env.fromElements(1).map(new RandomModelDataGenerator(dim, seed)));
+    }
+
+    private static class RandomModelDataGenerator
+            implements MapFunction<Integer, LogisticRegressionModelData> {
+        private final int dim;
+        private final int seed;
+
+        public RandomModelDataGenerator(int dim, int seed) {
+            this.dim = dim;
+            this.seed = seed;
+        }
+
+        @Override
+        public LogisticRegressionModelData map(Integer integer) throws Exception {
+            DenseVector vector = new DenseVector(dim);
+            Random random = new Random(seed);
+            for (int j = 0; j < dim; j++) {
+                vector.values[j] = random.nextDouble();
+            }
+            return new LogisticRegressionModelData(vector, 0L);
+        }
+    }
 
     /**
      * Converts the table model to a data stream.
@@ -63,21 +103,24 @@ public class LogisticRegressionModelData {
         StreamTableEnvironment tEnv =
                 (StreamTableEnvironment) ((TableImpl) modelData).getTableEnvironment();
         return tEnv.toDataStream(modelData)
-                .map(x -> new LogisticRegressionModelData(((Vector) x.getField(0)).toDense()));
+                .map(x -> new LogisticRegressionModelData(x.getFieldAs(0), x.getFieldAs(1)));
     }
 
-    /** Data encoder for {@link LogisticRegressionModel}. */
+    /** Data encoder for {@link LogisticRegression} and {@link OnlineLogisticRegression}. */
     public static class ModelDataEncoder implements Encoder<LogisticRegressionModelData> {
 
         @Override
         public void encode(LogisticRegressionModelData modelData, OutputStream outputStream)
                 throws IOException {
+            DataOutputViewStreamWrapper dataOutputViewStreamWrapper =
+                    new DataOutputViewStreamWrapper(outputStream);
             DenseVectorSerializer.INSTANCE.serialize(
-                    modelData.coefficient, new DataOutputViewStreamWrapper(outputStream));
+                    modelData.coefficient, dataOutputViewStreamWrapper);
+            dataOutputViewStreamWrapper.writeLong(modelData.modelVersion);
         }
     }
 
-    /** Data decoder for {@link LogisticRegressionModel}. */
+    /** Data decoder for {@link LogisticRegression} and {@link OnlineLogisticRegression}. */
     public static class ModelDataDecoder extends SimpleStreamFormat<LogisticRegressionModelData> {
 
         @Override
@@ -88,10 +131,13 @@ public class LogisticRegressionModelData {
                 @Override
                 public LogisticRegressionModelData read() throws IOException {
                     try {
+                        DataInputViewStreamWrapper dataInputViewStreamWrapper =
+                                new DataInputViewStreamWrapper(inputStream);
                         DenseVector coefficient =
                                 DenseVectorSerializer.INSTANCE.deserialize(
-                                        new DataInputViewStreamWrapper(inputStream));
-                        return new LogisticRegressionModelData(coefficient);
+                                        dataInputViewStreamWrapper);
+                        long modelVersion = dataInputViewStreamWrapper.readLong();
+                        return new LogisticRegressionModelData(coefficient, modelVersion);
                     } catch (EOFException e) {
                         return null;
                     }
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
@@ -1,0 +1,424 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.classification.logisticregression;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.functions.ReduceFunction;
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.ObjectArrayTypeInfo;
+import org.apache.flink.iteration.DataStreamList;
+import org.apache.flink.iteration.IterationBody;
+import org.apache.flink.iteration.IterationBodyResult;
+import org.apache.flink.iteration.Iterations;
+import org.apache.flink.iteration.operator.OperatorStateUtils;
+import org.apache.flink.ml.api.Estimator;
+import org.apache.flink.ml.common.datastream.DataStreamUtils;
+import org.apache.flink.ml.linalg.BLAS;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.runtime.state.FunctionInitializationContext;
+import org.apache.flink.runtime.state.FunctionSnapshotContext;
+import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.streaming.api.checkpoint.CheckpointedFunction;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
+import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
+import org.apache.flink.streaming.api.operators.TwoInputStreamOperator;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.collections.IteratorUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An Estimator which implements the online logistic regression algorithm. The online optimizer of
+ * this algorithm is The FTRL-Proximal proposed by H.Brendan McMahan et al.
+ *
+ * <p>See <a href="https://doi.org/10.1145/2487575.2488200">H. Brendan McMahan et al., Ad click
+ * prediction: a view from the trenches.</a>
+ */
+public class OnlineLogisticRegression
+        implements Estimator<OnlineLogisticRegression, OnlineLogisticRegressionModel>,
+                OnlineLogisticRegressionParams<OnlineLogisticRegression> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private Table initModelDataTable;
+
+    public OnlineLogisticRegression() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public OnlineLogisticRegressionModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<LogisticRegressionModelData> modelDataStream =
+                LogisticRegressionModelData.getModelDataStream(initModelDataTable);
+
+        DataStream<Row> points =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                new FeaturesLabelExtractor(
+                                        getFeaturesCol(), getLabelCol(), getWeightCol()));
+
+        DataStream<DenseVector> initModelData =
+                modelDataStream.map(
+                        (MapFunction<LogisticRegressionModelData, DenseVector>)
+                                value -> value.coefficient);
+
+        initModelData.getTransformation().setParallelism(1);
+
+        IterationBody body =
+                new FtrlIterationBody(
+                        getGlobalBatchSize(), getAlpha(), getBeta(), getReg(), getElasticNet());
+
+        DataStream<LogisticRegressionModelData> onlineModelData =
+                Iterations.iterateUnboundedStreams(
+                                DataStreamList.of(initModelData), DataStreamList.of(points), body)
+                        .get(0);
+
+        Table onlineModelDataTable = tEnv.fromDataStream(onlineModelData);
+        OnlineLogisticRegressionModel model =
+                new OnlineLogisticRegressionModel().setModelData(onlineModelDataTable);
+        ReadWriteUtils.updateExistingParams(model, paramMap);
+        return model;
+    }
+
+    private static class FeaturesLabelExtractor implements MapFunction<Row, Row> {
+        private final String featuresCol;
+        private final String labelCol;
+        private final String weightCol;
+
+        private FeaturesLabelExtractor(String featuresCol, String labelCol, String weightCol) {
+            this.featuresCol = featuresCol;
+            this.labelCol = labelCol;
+            this.weightCol = weightCol;
+        }
+
+        @Override
+        public Row map(Row row) throws Exception {
+            if (weightCol == null) {
+                return Row.of(row.getField(featuresCol), row.getField(labelCol));
+            } else {
+                return Row.of(
+                        row.getField(featuresCol), row.getField(labelCol), row.getField(weightCol));
+            }
+        }
+    }
+
+    /**
+     * In the implementation of ftrl optimizer, gradients are calculated in distributed workers and
+     * reduce them to one final gradient. The reduced gradient is used to update model by ftrl
+     * method. When the feature vector is dense, it can get the same result as tensorflow's ftrl. If
+     * feature vector is sparse, we use the mean value in every feature dim instead of mean value of
+     * whole vector, which can get a better convergence.
+     *
+     * <p>See https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Ftrl
+     *
+     * <p>todo: makes ftrl to be a common optimizer and place it in org.apache.flink.ml.common.
+     */
+    private static class FtrlIterationBody implements IterationBody {
+        private final int batchSize;
+        private final double alpha;
+        private final double beta;
+        private final double l1;
+        private final double l2;
+
+        public FtrlIterationBody(
+                int batchSize, double alpha, double beta, double reg, double elasticNet) {
+            this.batchSize = batchSize;
+            this.alpha = alpha;
+            this.beta = beta;
+            this.l1 = elasticNet * reg;
+            this.l2 = (1 - elasticNet) * reg;
+        }
+
+        @Override
+        public IterationBodyResult process(
+                DataStreamList variableStreams, DataStreamList dataStreams) {
+            DataStream<DenseVector> modelData = variableStreams.get(0);
+
+            DataStream<Row> points = dataStreams.get(0);
+            int parallelism = points.getParallelism();
+            Preconditions.checkState(
+                    parallelism <= batchSize,
+                    "There are more subtasks in the training process than the number "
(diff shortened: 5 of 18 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LinearModelData.java
```diff
@@ -37,60 +38,68 @@
...
 
 /**
- * Model data of {@link LogisticRegressionModel}.
+ * Model data of {@link LogisticRegressionModel}, {@link FtrlModel}.
  *
  * <p>This class also provides methods to convert model data from Table to Datastream, and classes
  * to save/load model data.
  */
-public class LogisticRegressionModelData {
+public class LinearModelData {
```
yunfengzhou-hub, line 46: Could you please illustrate the relationship between FTRL and LogisticRegression, and other algorithms like LinearRegression? I'm not sure why we would like to rename `LogisticRegressionModelData` as `LinearModelData`.

If after discussion we still agree that this renaming is reasonable, it would mean that the model data class neither belongs `logisticregresson` or `ftrl` package. We would need to place classes like this to a neutral package, like something named `common`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LinearModelData.java
```diff
@@ -37,60 +38,68 @@
...
 
     public DenseVector coefficient;
+    public long modelVersion;
```
yunfengzhou-hub, line 49: The versioning mechanism is different from that in `OnlineKMeans`. Shall we adopt the same practice across both online algorithms?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlModelParams.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.flink.ml.common.param.HasPredictionCol;
+import org.apache.flink.ml.common.param.HasRawPredictionCol;
+
+/**
+ * Params for {@link FtrlModel}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface FtrlModelParams<T>
+        extends HasFeaturesCol<T>, HasPredictionCol<T>, HasRawPredictionCol<T> {}
```
yunfengzhou-hub, line 31: This interface is identical to `LogisticRegressionModelParams`. If FTRL is an online version of LogisticRegression, shall we consider reorganizing code to avoid creating duplicate classes like this?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlParams.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+
+/**
+ * Params of {@link Ftrl}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface FtrlParams<T>
+        extends HasLabelCol<T>, HasBatchStrategy<T>, HasGlobalBatchSize<T>, HasFeaturesCol<T> {
```
yunfengzhou-hub, line 36: An `Estimator`'s param class should inherit the corresponding `Model`'s param class. In this current implementation, `Ftrl` would not be able to set the `rawPredictionCol` and `predictionCol` of the generated `FtrlModel`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlParams.java
```diff
@@ -0,0 +1,92 @@
...
+
+    Param<Integer> VECTOR_SIZE =
+            new IntParam("vectorSize", "The size of vector.", -1, ParamValidators.gt(-2));
+
+    default Integer getVectorSize() {
```
yunfengzhou-hub, line 41: This param seems unused.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlParams.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    default T setL1(Double value) {
+        return set(L_1, value);
+    }
+
+    Param<Double> L_2 =
+            new DoubleParam("l2", "The parameter l2 of ftrl.", 0.1, ParamValidators.gt(0.0));
+
+    default Double getL2() {
```
yunfengzhou-hub, line 63: It seems that these parameters are also used in other algorithms like soft max and multi-layer perception. Let's define them as common parameters.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlParams.java
yunfengzhou-hub, line 36: It seems that some parameters provided by LogisticRegression, like `weightCol`, `reg` and `multiClass`, are not supported by FTRL for now. Is FTRL supposed to have weaker functionality than LogisticRegression, if both used in offline training process? If not, do we have any plan of adding support for these parameters in future?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlModel.java
```diff
@@ -0,0 +1,196 @@
...
+            modelDataVersion = modelData.modelVersion;
+            // System.out.println("update model...");
+            // todo : receive model data.
+            // Preconditions.checkArgument(modelData.centroids.length <= k);
+            // centroids = modelData.centroids;
+            // modelDataVersion++;
+            // for (Row dataPoint : bufferedPointsState.get()) {
+            //	processElement1(new StreamRecord<>(dataPoint));
+            // }
+            // bufferedPointsState.clear();
```
yunfengzhou-hub, line 149: Let's reformat the code to remove unused codes.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlModel.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * The main logic that predicts one input record.
+     *
+     * @param feature The input feature.
+     * @param coefficient The model parameters.
+     * @return The prediction label and the raw probabilities.
+     */
+    public static Tuple2<Double, DenseVector> predictRaw(Vector feature, DenseVector coefficient)
```
yunfengzhou-hub, line 160: Methods like this are almost identical to that in `LogisticRegressionModel`. It could be better if we could reuse logics that have already been defined in `LogisticRegression` and `LogisticRegressionModel`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/Ftrl.java
```diff
@@ -0,0 +1,395 @@
...
+                            modelData
+                                    .map(
+                                            new MapFunction<DenseVector[], LinearModelData>() {
+                                                long iter = 0L;
+
+                                                @Override
+                                                public LinearModelData map(DenseVector[] value) {
+                                                    return new LinearModelData(value[0], iter++);
+                                                }
+                                            })
```
yunfengzhou-hub, line 159: Shall we move logics like this to a separated class or method? That could make the code look prettier.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/Ftrl.java
```diff
@@ -0,0 +1,395 @@
...
+                        p += modelData[0].values[idx] * sparseVector.values[i];
+                    }
+                    p = 1 / (1 + Math.exp(-p));
+                    for (int i = 0; i < sparseVector.indices.length; ++i) {
+                        int idx = sparseVector.indices[i];
+                        double g = (p - point.f1) * sparseVector.values[i];
+                        double sigma = (Math.sqrt(n[idx] + g * g) - Math.sqrt(n[idx])) / alpha;
+                        z[idx] += g - sigma * modelData[0].values[idx];
+                        n[idx] += g * g;
+                        weights.values[idx] += 1.0;
```
yunfengzhou-hub, line 305: It seems that weights are only updated but not used.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/Ftrl.java
```diff
@@ -0,0 +1,395 @@
...
+            alignAndComputeModelData();
+        }
+
+        @Override
+        public void processElement2(StreamRecord<DenseVector[]> modelDataRecord) throws Exception {
+            modelDataState.add(modelDataRecord.getValue());
+            alignAndComputeModelData();
+        }
+
+        private void alignAndComputeModelData() throws Exception {
```
yunfengzhou-hub, line 241: Would it be better to add some comments to this method, or divide this method further into several smaller methods? That could help improve readability.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/Ftrl.java
```diff
@@ -0,0 +1,395 @@
...
+            for (Tuple2<Vector, Double> point : points) {
+                if (n == null) {
+                    n = new double[point.f0.size()];
+                    z = new double[n.length];
+                    weights = new DenseVector(n.length);
+                }
+
+                double p = 0.0;
+                Arrays.fill(weights.values, 0.0);
+                if (point.f0 instanceof DenseVector) {
```
yunfengzhou-hub, line 265: It seems that some logic in `if` and in `else` are the same. Shall we move these codes out of the if-else condition?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/Ftrl.java
```diff
@@ -0,0 +1,395 @@
...
+                    }
+                }
+            }
+            output.collect(new StreamRecord<>(new DenseVector[] {modelData[0], weights}));
+        }
+    }
+
+    /** Parses samples of input data. */
+    public static class ParseSample extends RichMapFunction<Row, Tuple2<Vector, Double>> {
+        private static final long serialVersionUID = 3738888745125082777L;
```
yunfengzhou-hub, line 315: Is variables like this required?

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LinearModelData.java
weibozhao: LinearModelData is shared by ftrl, lr, svm, linearReg. For these algorithm has the same model data format. We need to talk about this rename. If rename, I agree to move this linear model data to a common place.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/ftrl/FtrlModelParams.java
weibozhao: FTRL is not an online version of LogisticRegression. It just has the same model and predict process. The update processes are different.

zhipeng93:
Should we make `FTRL` an optimizer or an algorithm? As I see, FTRL is an optimizer that could be used to optimize Logistic Regression, just as SGD.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,428 @@
...
+                        .get(0);
+
+        Table onlineModelDataTable = tEnv.fromDataStream(onlineModelData);
+        OnlineLogisticRegressionModel model =
+                new OnlineLogisticRegressionModel().setModelData(onlineModelDataTable);
+        ReadWriteUtils.updateExistingParams(model, paramMap);
+        return model;
+    }
+
+    /** Iteration body of ftrl algorithm. */
```
zhipeng93, line 118: It would be nice to if we have more introduction of FTRL method.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,428 @@
...
+                                            })
+                            .setParallelism(1);
+
+            DataStream<LogisticRegressionModelData> outputModelData =
+                    feedbackModelData
+                            .map(
+                                    (MapFunction<DenseVector[], LogisticRegressionModelData>)
+                                            value ->
+                                                    new LogisticRegressionModelData(
+                                                            value[0], System.nanoTime()))
```
zhipeng93, line 200: Can we add the model version when the infra for model version is ready?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,428 @@
...
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /**
+     * An operator that splits a global batch into evenly-sized local batches, and distributes them
+     * to downstream operator.
+     */
+    private static class GlobalBatchSplitter implements FlatMapFunction<Row[], Row[]> {
```
zhipeng93, line 382: Can we reuse the code snippet in `OnlineKmeans`, rather than copy-paste it here?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionModel.java
```diff
@@ -0,0 +1,225 @@
...
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(),
+                                Types.DOUBLE,
+                                TypeInformation.of(DenseVector.class),
+                                Types.LONG),
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldNames(),
+                                getPredictionCol(),
+                                getRawPredictionCol(),
+                                getModelVersionCol()));
```
zhipeng93, line 87: The behavior here is inconsistent with `OnlineKmeansModel`. We probably need to discuss about whether it is necessary to output the model version.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionModel.java
```diff
@@ -0,0 +1,225 @@
...
+                    .getMetricGroup()
+                    .gauge(
+                            "MODEL_DATA_VERSION_GAUGE_KEY",
+                            (Gauge<String>) () -> Long.toString(modelDataVersion));
+        }
+
+        @Override
+        public void processElement1(StreamRecord<Row> streamRecord) throws Exception {
+            Row dataPoint = streamRecord.getValue();
+            while (coefficient == null) {
```
zhipeng93, line 140: nit: could be `if`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionModel.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * The main logic that predicts one input record.
+     *
+     * @param feature The input feature.
+     * @param coefficient The model parameters.
+     * @return The prediction label and the raw probabilities.
+     */
+    private static Tuple2<Double, DenseVector> predictRaw(Vector feature, DenseVector coefficient) {
```
zhipeng93, line 178: This logic is the same as that in `LogisticRegression`. Should we reuse the logic there?

zhipeng93 (commented):
@weibozhao Thanks for the PR! I left some comments below.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public interface OnlineLogisticRegressionParams<T>
+        extends HasLabelCol<T>,
+                HasBatchStrategy<T>,
+                HasGlobalBatchSize<T>,
+                HasL1<T>,
+                HasL2<T>,
+                OnlineLogisticRegressionModelParams<T> {
+
+    Param<Double> ALPHA =
```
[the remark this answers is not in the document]
weibozhao: OK，I will add this param.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegression.java
```diff
@@ -391,7 +391,8 @@ public void onIterationTerminated(Context context, Collector<double[]> collector
             feedbackBufferState.clear();
             if (getRuntimeContext().getIndexOfThisSubtask() == 0) {
                 updateModel();
-                context.output(modelDataOutputTag, new LogisticRegressionModelData(coefficient));
+                context.output(
+                        modelDataOutputTag, new LogisticRegressionModelData(coefficient, 0L));
```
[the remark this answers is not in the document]
weibozhao: This is not conflict with barrierModelData. This just add a version for every model produced by online algo or offline algo. then, the offline and online algo can share the same model data.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelData.java
```diff
@@ -37,17 +37,19 @@
...
- * Model data of {@link LogisticRegressionModel}.
+ * Model data of {@link LogisticRegressionModel} and {@link OnlineLogisticRegressionModel}.
  *
  * <p>This class also provides methods to convert model data from Table to Datastream, and classes
  * to save/load model data.
  */
 public class LogisticRegressionModelData {
 
     public DenseVector coefficient;
+    public long modelVersion;
```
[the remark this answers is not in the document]
weibozhao: see above comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasL2.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.DoubleParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared L2 param. */
+public interface HasL2<T> extends WithParams<T> {
+    Param<Double> L_2 = new DoubleParam("l2", "The l2 param.", 0.1, ParamValidators.gt(0.0));
```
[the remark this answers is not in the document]
weibozhao: L1 and L2 are the hyper parameters for ftrl. It has different meaning from regParam.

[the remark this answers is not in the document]
lindong28: @weibozhao Currently the class `LogisticRegression` pass the value of `getReg()` to `LogisticGradient::l2`, which suggests that getReg() actually represents the hyper-parameter l2 for `LogisticRegression`.

Could you explain why we can not get the value of `l2` from `getReg())` for `OnlineLogisticRegression`?

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        @Override
+        public void cancel() {
+            isRunning = false;
+        }
+    }
+
+    private Table getTable(
+            int timeInterval, int numSample, List<Row> data, int parallel, boolean isSparse) {
```
lindong28, line 429: Would it be more readable to rename `parallel` as `sourceParallelism`?

Would it be more readable to rename this method as `getSourceTable`?

When the parallelism is 2, the last 50% of elements of `data` will never be emitted by the source function. Is this expected?

Would you help explain the motivation to test source function with different parallelism? For example, would it be sufficient to only test `sourceParallelism = 2`?

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
```diff
@@ -0,0 +1,444 @@
...
+                        .setInitialModelData(initSparseModel)
+                        .setGlobalBatchSize(500)
+                        .setLabelCol(LABEL_COL)
+                        .fit(onlineTrainTable);
+        Table resultTable = model.setPredictionCol(PREDICT_COL).transform(onlinePredictTable)[0];
+        verifyPredictionResult(resultTable, 4);
+    }
+
+    @Test
+    public void testParam() {
```
lindong28, line 211: nits: could we order the methods in this class to be more consistent with test methods in other test classes, e.g. move the `testParam()` to be the first method in this class?

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
```diff
@@ -0,0 +1,444 @@
...
+        Table onlineTrainTable = getTable(2, 2000, TRAIN_DENSE_ROWS, 2, false);
+        Table onlinePredictTable = getTable(2, 3000, PREDICT_DENSE_ROWS, 2, false);
+        OnlineLogisticRegressionModel model =
+                new OnlineLogisticRegression()
+                        .setFeaturesCol(FEATURE_COL)
+                        .setInitialModelData(initDenseModel)
+                        .setGlobalBatchSize(500)
+                        .setLabelCol(LABEL_COL)
+                        .fit(onlineTrainTable);
+        Table resultTable = model.setPredictionCol(PREDICT_COL).transform(onlinePredictTable)[0];
```
lindong28, line 188: Since `setPredictionCol` is already tested in `testParam()`, would it be simpler to skip calling `setPredictionCol(PREDICT_COL)` in this test?

Same for other invocations of `setFeaturesCol(FEATURE_COL)` and `setLabelCol(LABEL_COL)` in this class.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
```diff
@@ -0,0 +1,444 @@
...
+        Table onlinePredictTable = getTable(5, 3000, PREDICT_SPARSE_ROWS, 4, true);
+        OnlineLogisticRegressionModel model =
+                new OnlineLogisticRegression()
+                        .setFeaturesCol(FEATURE_COL)
+                        .setInitialModelData(initSparseModel)
+                        .setGlobalBatchSize(500)
+                        .setLabelCol(LABEL_COL)
+                        .fit(onlineTrainTable);
+        Table resultTable = model.setPredictionCol(PREDICT_COL).transform(onlinePredictTable)[0];
+        verifyPredictionResult(resultTable, 4);
```
lindong28, line 207: This test would generate 2000 / 500 = 4 model data versions. And a given element in the `onlinePredictTable` might be transformed by any one of these 4 model data versions, right?

If so, the prediction result is non-deterministic and we won't be able to deterministically verify the numerical values of the prediction result. This might explain why this PR has test failures.

The `OnlineKMeansTest::testFitAndPredict` is able to address this problem using e.g. `waitModelDataUpdate()`. 

Could you try to follow the approach used in `OnlineKMeansTest` and make sure all tests added in this PR could succeed?

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasL2.java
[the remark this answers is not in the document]
lindong28: According to the Elastic net regularization wiki page [1], l1 and l2 could be theoretically translated to a regularization constant and the SVM regularization constant (i.e. the elastic net). Thus the regularization constant is different from l1 and l2 under this context

As far as the mathematical formula is concerned, is the existing `HasReg::REG` the same as L2 or the regularization constant explained in [1]? If it is the same as L2, we probably should rename this class to e.g. `HasL2`. Otherwise, we probably should rename the variable `LogisticGradient::l2` as appropriate.

[1] https://en.wikipedia.org/wiki/Elastic_net_regularization#:~:text=In%20statistics%20and%2C%20in%20particular,the%20lasso%20and%20ridge%20methods.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
weibozhao: For the first question, I don't think so. The data is random sampling from a fixed dataset.


The user set the total num samples, we need to calculate the num samples on every worker, then we need source parallel parameter.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
weibozhao: I will refine it later.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasL2.java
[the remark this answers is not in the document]
zhipeng93: After some offline discussion, @weibozhao  and I agree that we could reuse the exising `HasReg` and `HasElasticNet` param, without introducing `HasL1` and `HasL2`.

zhipeng93 (commented):
Thanks for the update. I found the implementation of distributed FTRL a bit confused. Left some comments below.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -182,4 +191,49 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
...
         }
     }
+
+    /**
+     * An operator that splits a global batch into evenly-sized local batches, and distributes them
+     * to downstream operator.
+     *
+     * @param <T> Data type of batch data.
+     */
+    public static class GlobalBatchSplitter<T> implements FlatMapFunction<T[], T[]> {
```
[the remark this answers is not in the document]
weibozhao: I think one function is better. I will refine it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,428 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An Estimator which implements the OnlineLogisticRegression algorithm.
+ *
+ * <p>See https://en.wikipedia.org/wiki/Online_machine_learning.
```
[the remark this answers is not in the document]
weibozhao: I have refine it.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasL2.java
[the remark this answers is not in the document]
weibozhao: done

yunfengzhou-hub (commented):
Thanks for the update. Here are some comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,434 @@
...
+                LogisticRegressionModelData.getModelDataStream(initModelDataTable);
+
+        DataStream<Row> points =
+                tEnv.toDataStream(inputs[0])
+                        .map(new FeaturesExtractor(getFeaturesCol(), getLabelCol()));
+
+        DataStream<DenseVector> initModelData =
+                modelDataStream.map(
+                        (MapFunction<LogisticRegressionModelData, DenseVector>)
+                                value -> value.coefficient);
```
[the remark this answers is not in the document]
weibozhao: I think it's not needed. The model version in this algorithm is from 1 and increasing every sink. The start version is 0 or 1 has no meaning.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
```diff
@@ -0,0 +1,661 @@
...
+            new Row[] {
+                Row.of(Vectors.sparse(10, new int[] {1, 3, 4}, ONE_ARRAY), 0.),
+                Row.of(Vectors.sparse(10, new int[] {0, 2, 3}, ONE_ARRAY), 0.),
+                Row.of(Vectors.sparse(10, new int[] {0, 3, 4}, ONE_ARRAY), 0.),
+                Row.of(Vectors.sparse(10, new int[] {2, 3, 4}, ONE_ARRAY), 0.),
+                Row.of(Vectors.sparse(10, new int[] {1, 3, 4}, ONE_ARRAY), 0.),
+                Row.of(Vectors.sparse(10, new int[] {6, 7, 8}, ONE_ARRAY), 1.),
+                Row.of(Vectors.sparse(10, new int[] {6, 8, 9}, ONE_ARRAY), 1.),
+                Row.of(Vectors.sparse(10, new int[] {5, 8, 9}, ONE_ARRAY), 1.),
+                Row.of(Vectors.sparse(10, new int[] {5, 6, 7}, ONE_ARRAY), 1.)
```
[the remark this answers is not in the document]
weibozhao: If we use the same data, some sparse properties may not covered by these ut.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    default T setBeta(Double value) {
+        return set(BETA, value);
+    }
+
+    Param<Integer> MODEL_SAVE_INTERVAL =
+            new IntParam(
+                    "modelSaveInterval",
+                    "The iteration steps between two output models.",
```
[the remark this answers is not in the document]
weibozhao: Maybe, we can discus it with @zhipeng93 this afternoon.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,434 @@
...
+                                    new FilterFunction<DenseVector>() {
+                                        private int step = 0;
+
+                                        @Override
+                                        public boolean filter(DenseVector denseVector) {
+                                            step++;
+                                            return step % modelSaveInterval == 0;
+                                        }
+                                    })
+                            .setParallelism(1)
```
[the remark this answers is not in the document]
yunfengzhou-hub: Got it.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. Thanks.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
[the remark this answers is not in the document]
yunfengzhou-hub: That sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,434 @@
...
+                    coefficient[i] =
+                            ((zParam[i] < 0 ? -1 : 1) * l1 - zParam[i])
+                                    / ((beta + Math.sqrt(nParam[i])) / alpha + l2);
+                }
+            }
+            output.collect(new StreamRecord<>(new DenseVector(coefficient)));
+        }
+    }
+
+    private static class CalculateLocalGradient extends AbstractStreamOperator<DenseVector[]>
```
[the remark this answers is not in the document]
weibozhao: add todo.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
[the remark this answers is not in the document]
weibozhao: I will remove this parameter just as online kmeans。

zhipeng93 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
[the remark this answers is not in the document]
weibozhao: OK, I will refine it.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.flink.ml.param.DoubleParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+
+/**
+ * Params of {@link OnlineLogisticRegression}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface OnlineLogisticRegressionParams<T>
```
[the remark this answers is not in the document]
weibozhao: I will add this param later.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelData.java
```diff
@@ -18,40 +18,85 @@
...
+    public LogisticRegressionModelData(DenseVector coefficient, long modelVersion) {
         this.coefficient = coefficient;
+        this.modelVersion = modelVersion;
     }
 
     public LogisticRegressionModelData() {}
 
+    public LogisticRegressionModelData(DenseVector coefficient) {
+        this.coefficient = coefficient;
+        this.modelVersion = 0L;
```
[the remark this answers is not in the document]
weibozhao: I think, leave a TODO is a better choice.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,434 @@
...
+                    weight[idx] += 1.0;
+                }
+            }
+
+            if (points.length > 0) {
+                output.collect(
+                        new StreamRecord<>(
+                                new DenseVector[] {
+                                    new DenseVector(gradient),
+                                    new DenseVector(weight),
```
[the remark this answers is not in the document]
weibozhao: The `weight`  will be renamed as `weightSum`. For as above comment I add weightCol param, this variable is just the sum of weight. 

I will add java doc for this code and I will refine the pattern of output model.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
```diff
@@ -0,0 +1,418 @@
...
+import org.apache.commons.collections.IteratorUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** An Estimator which implements the online logistic regression algorithm. */
```
lindong28, line 65: Should we provide reference/link to the original paper, so that users could know what is the algorithm and why it is useful?

Feel free to see KMeans.scala in Spark ML for example doc.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
```diff
@@ -0,0 +1,66 @@
...
+    default Double getAlpha() {
+        return get(ALPHA);
+    }
+
+    default T setAlpha(Double value) {
+        return set(ALPHA, value);
+    }
+
+    Param<Double> BETA =
+            new DoubleParam("alpha", "The parameter beta of ftrl.", 0.1, ParamValidators.gt(0.0));
```
lindong28, line 57: Hmm.. how would users know what is FTRL when they read this doc?

weibozhao: I have add ftrl doc in the doc of Online logistic regression.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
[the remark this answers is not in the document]
lindong28: I believe the model version does need to take into account the version of the input model data so that we can know the order of models across program restarts.

Here is an example scenario:
- Let's say the global batch size is 100 and there are 1000 records in the input datastream. The online training program should generate 10 model versions after processing these 1000 records.
- The training process finished processing 300 records and generated 3 model data with versions 1, 2, and 3. After successfully making a checkpoint, the process exited due to machine failure.
- The training process is restarted from the last successful checkpoint. It should continue to read input datastream starting from the 301th record. And it should read the latest model data generated before it is restarted.

Ideally, we should hide the machine failure from users, meaning that the sequence of model versions should be 1, 2, 3, 4, ...10 as if the failure has never happened. Therefore we have to set the initial model version to the model version from the input model data.

Does this make sense?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
```diff
@@ -0,0 +1,66 @@
...
+                HasGlobalBatchSize<T>,
+                HasReg<T>,
+                HasElasticNet<T>,
+                OnlineLogisticRegressionModelParams<T> {
+
+    Param<Double> ALPHA =
+            new DoubleParam("alpha", "The parameter alpha of ftrl.", 0.1, ParamValidators.gt(0.0));
+
+    Param<Double> BETA =
+            new DoubleParam("alpha", "The parameter beta of ftrl.", 0.1, ParamValidators.gt(0.0));
```
lindong28, line 49: alpha -> beta.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/OnlineLogisticRegressionTest.java
```diff
@@ -0,0 +1,687 @@
...
+        Assert.assertEquals(0.1, onlineLogisticRegression.getAlpha(), 1.0e-5);
+        Assert.assertEquals(0.1, onlineLogisticRegression.getBeta(), 1.0e-5);
+        Assert.assertEquals(32, onlineLogisticRegression.getGlobalBatchSize());
+
+        onlineLogisticRegression
+                .setFeaturesCol("test_feature")
+                .setLabelCol("test_label")
+                .setGlobalBatchSize(5)
+                .setReg(0.25)
+                .setElasticNet(0.25)
```
lindong28, line 369: Should we use different values for these parameters so that we can capture the bugs mentioned in the other comment?

Same for the python tests.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionModel.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static OnlineLogisticRegressionModel load(StreamTableEnvironment tEnv, String path)
+            throws IOException {
+        OnlineLogisticRegressionModel model = ReadWriteUtils.loadStageParam(path);
+        Table modelDataTable =
```
lindong28, line 180: Given that `OnlineLogisticRegressionModel::save(...)` does not write model data to disk, it seems inconsistent to have `OnlineLogisticRegressionModel::load(...)` read model data from the given path.

Should we follow the same behavior of `OnlineKMeansModel::load(...)` and only read metadata in this mehod?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegressionParams.java
```diff
@@ -0,0 +1,66 @@
...
+        extends HasLabelCol<T>,
+                HasWeightCol<T>,
+                HasBatchStrategy<T>,
+                HasGlobalBatchSize<T>,
+                HasReg<T>,
+                HasElasticNet<T>,
+                OnlineLogisticRegressionModelParams<T> {
+
+    Param<Double> ALPHA =
+            new DoubleParam("alpha", "The parameter alpha of ftrl.", 0.1, ParamValidators.gt(0.0));
```
lindong28, line 46: nits: Would it be more intuitive to say `The alpha parameter ...`?

Same for `The beta parameter ...`.

Same for the python parameter description.

lindong28 (commented):
Thanks for the update. The PR looks much better now. Left minor comments below.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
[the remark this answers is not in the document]
weibozhao: modelVersion has been write to state. If restarting happened, algorithm will read the model version from checkpoint. Just as your example, if restarting happened at record 301, the model version data (without model version) is read from checkpoint in CalculateLocalGradient() and the modelVersion (modelVersion=3) is read from checkpoint in the CreateLrModelData() before sink model out.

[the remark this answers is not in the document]
lindong28: Thanks for the explanation. Sounds good.

lindong28:
Thanks for the update! LGTM.

@zhipeng93 Do you want to take another look before merging this PR?

flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
```diff
@@ -74,6 +75,15 @@ public void testReduce() throws Exception {
         assertArrayEquals(new long[] {190L}, sum.stream().mapToLong(Long::longValue).toArray());
     }
 
+    @Test
+    public void testGenerateBatchData() throws Exception {
+        DataStream<Long> dataStream =
+                env.fromParallelCollection(new NumberSequenceIterator(0L, 19L), Types.LONG);
+        DataStream<Long[]> result = DataStreamUtils.generateBatchData(dataStream, 2, 4);
+        List<Long[]> batches = IteratorUtils.toList(result.executeAndCollect());
+        assertEquals(10, batches.size());
```
zhipeng93, line 84: Could you please check the size of each local batch?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelData.java
```diff
@@ -18,40 +18,86 @@
...
+     * @param tEnv The environment where to create the table.
+     * @param dim The size of generated coefficient.
+     * @param seed Random seed.
+     */
+    public static Table generateRandomModelData(StreamTableEnvironment tEnv, int dim, int seed) {
+        StreamExecutionEnvironment env = TableUtils.getExecutionEnvironment(tEnv);
+        return tEnv.fromDataStream(env.fromElements(1).map(new GenerateRandomModel(dim, seed)));
+    }
+
+    private static class GenerateRandomModel
```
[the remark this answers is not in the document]
zhipeng93: This comment is marked as resolved but seems not addressed. Did you forget about this?

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -182,4 +193,63 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
             }
         }
     }
+
+    /**
+     * A function that generate several data batches and distribute them to downstream operator.
+     *
+     * @param <T> Data type of batch data.
+     */
+    public static <T> DataStream<T[]> generateBatchData(
```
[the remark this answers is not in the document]
zhipeng93: This comment is not addressed nut marked as resolved. Did you forget to address it?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/OnlineLogisticRegression.java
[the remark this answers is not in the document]
zhipeng93: If the init model version is 5, then will you create a model with version smaller than 5?

zhipeng93 (commented):

(BTW, could you address the comment before resolve it? I noticed several comments resolved but not addressed...

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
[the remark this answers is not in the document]
weibozhao: I have addressed already.

PR #86, opened 2022-04-19
[FLINK-27294] Add Transformer for BinaryClassificationEvaluator

weibozhao:
- Add Transformer for BinaryClassificationEvaluator. 

## Brief change log
- Added BinaryClassificationEvaluator.
- Added unit tests for BinaryClassificationEvaluator.
- Added ParamValidators::isSubset.

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
@@ -28,22 +28,22 @@ public class ParamValidators {
         return (value) -> true;
     }
 
-    // Check if the parameter value is greater than lowerBound.
+    // Checks if the parameter value is greater than lowerBound.
     public static <T> ParamValidator<T> gt(double lowerBound) {
         return (value) -> value != null && ((Number) value).doubleValue() > lowerBound;
     }
 
-    // Check if the parameter value is greater than or equal to lowerBound.
+    // Checks if the parameter value is greater than or equal to lowerBound.
     public static <T> ParamValidator<T> gtEq(double lowerBound) {
         return (value) -> value != null && ((Number) value).doubleValue() >= lowerBound;
     }
 
-    // Check if the parameter value is less than upperBound.
+    // Checks if the parameter value is less than upperBound.
     public static <T> ParamValidator<T> lt(double upperBound) {
         return (value) -> value != null && ((Number) value).doubleValue() < upperBound;
     }
 
-    // Check if the parameter value is less than or equal to upperBound.
+    // Checks if the parameter value is less than or equal to upperBound.
     public static <T> ParamValidator<T> ltEq(double upperBound) {
         return (value) -> value != null && ((Number) value).doubleValue() <= upperBound;
     }
@@ -71,12 +71,12 @@ public class ParamValidators {
         };
     }
 
-    // Check if the parameter value is in the range [lowerBound, upperBound].
+    // Checks if the parameter value is in the range [lowerBound, upperBound].
     public static <T> ParamValidator<T> inRange(double lowerBound, double upperBound) {
         return inRange(lowerBound, upperBound, true, true);
     }
 
-    // Check if the parameter value is in the array of allowed values.
+    // Checks if the parameter value is in the array of allowed values.
     public static <T> ParamValidator<T> inArray(T... allowed) {
         return new ParamValidator<T>() {
             @Override
@@ -86,7 +86,7 @@ public class ParamValidators {
         };
     }
 
-    // Check if the parameter value is not null.
+    // Checks if the parameter value is not null.
     public static <T> ParamValidator<T> notNull() {
         return new ParamValidator<T>() {
             @Override
@@ -96,8 +96,27 @@ public class ParamValidators {
         };
     }
 
-    // Check if the parameter value array is not empty array.
+    // Checks if the parameter value array is not empty array.
     public static <T> ParamValidator<T[]> nonEmptyArray() {
         return value -> value != null && value.length > 0;
+    }
+
+    // Checks if every element in the array-typed parameter value is in the array of allowed values.
+    @SafeVarargs
+    public static <T> ParamValidator<T[]> isSubSet(T... allowed) {
+        return new ParamValidator<T[]>() {
+            @Override
+            public boolean validate(T[] value) {
+                if (value == null) {
+                    return false;
+                }
+                for (T t : value) {
+                    if (!ArrayUtils.contains(allowed, t)) {
+                        return false;
+                    }
+                }
+                return true;
+            }
+        };
     }
 }
--- a/flink-ml-core/src/test/java/org/apache/flink/ml/api/StageTest.java
+++ b/flink-ml-core/src/test/java/org/apache/flink/ml/api/StageTest.java
@@ -463,5 +463,11 @@ public class StageTest {
         Assert.assertTrue(nonEmptyArray.validate(new String[] {"1"}));
         Assert.assertFalse(nonEmptyArray.validate(null));
         Assert.assertFalse(nonEmptyArray.validate(new String[0]));
+
+        ParamValidator<String[]> isSubArray = ParamValidators.isSubSet("a", "b", "c");
+        Assert.assertFalse(isSubArray.validate(null));
+        Assert.assertFalse(isSubArray.validate(new String[] {"c", "v"}));
+        Assert.assertTrue(isSubArray.validate(new String[] {"a", "b"}));
+        Assert.assertFalse(isSubArray.validate(new String[] {"e", "v"}));
     }
 }
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
@@ -1,0 +1,723 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.evaluation.binaryclassfication;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.functions.MapPartitionFunction;
+import org.apache.flink.api.common.functions.ReduceFunction;
+import org.apache.flink.api.common.functions.RichFlatMapFunction;
+import org.apache.flink.api.common.functions.RichMapFunction;
+import org.apache.flink.api.common.functions.RichMapPartitionFunction;
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.tuple.Tuple3;
+import org.apache.flink.api.java.tuple.Tuple4;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.api.scala.typeutils.Types;
+import org.apache.flink.iteration.operator.OperatorStateUtils;
+import org.apache.flink.ml.api.AlgoOperator;
+import org.apache.flink.ml.common.broadcast.BroadcastUtils;
+import org.apache.flink.ml.common.datastream.DataStreamUtils;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.runtime.state.StateSnapshotContext;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
+import org.apache.flink.streaming.api.operators.BoundedOneInput;
+import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.Preconditions;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+/**
+ * An Estimator which calculates the evaluation metrics for binary classification. The input data
+ * has columns rawPrediction, label and an optional weight column. The rawPrediction can be of type
+ * double (binary 0/1 prediction, or probability of label 1) or of type vector (length-2 vector of
+ * raw predictions, scores, or label probabilities). The output may contain different metrics which
+ * will be defined by parameter MetricsNames. See {@link BinaryClassificationEvaluatorParams}.
+ */
+public class BinaryClassificationEvaluator
+        implements AlgoOperator<BinaryClassificationEvaluator>,
+                BinaryClassificationEvaluatorParams<BinaryClassificationEvaluator> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final int NUM_SAMPLE_FOR_RANGE_PARTITION = 100;
+    private static final Logger LOG = LoggerFactory.getLogger(BinaryClassificationEvaluator.class);
+
+    public BinaryClassificationEvaluator() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple3<Double, Boolean, Double>> evalData =
+                tEnv.toDataStream(inputs[0])
+                        .map(new ParseSample(getLabelCol(), getRawPredictionCol(), getWeightCol()));
+        final String boundaryRangeKey = "boundaryRange";
+        final String partitionSummariesKey = "partitionSummaries";
+
+        DataStream<Tuple4<Double, Boolean, Double, Integer>> evalDataWithTaskId =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(evalData),
+                        Collections.singletonMap(boundaryRangeKey, getBoundaryRange(evalData)),
+                        inputList -> {
+                            DataStream input = inputList.get(0);
+                            return input.map(new AppendTaskId(boundaryRangeKey));
+                        });
+
+        /* Repartition the evaluated data by range. */
+        evalDataWithTaskId =
+                evalDataWithTaskId.partitionCustom((chunkId, numPartitions) -> chunkId, x -> x.f3);
+
+        /* Sorts local data by score.*/
+        DataStream<Tuple3<Double, Boolean, Double>> sortEvalData =
+                DataStreamUtils.mapPartition(
+                        evalDataWithTaskId,
+                        new MapPartitionFunction<
+                                Tuple4<Double, Boolean, Double, Integer>,
+                                Tuple3<Double, Boolean, Double>>() {
+                            @Override
+                            public void mapPartition(
+                                    Iterable<Tuple4<Double, Boolean, Double, Integer>> values,
+                                    Collector<Tuple3<Double, Boolean, Double>> out) {
+                                List<Tuple3<Double, Boolean, Double>> bufferedData =
+                                        new LinkedList<>();
+                                for (Tuple4<Double, Boolean, Double, Integer> t4 : values) {
+                                    bufferedData.add(Tuple3.of(t4.f0, t4.f1, t4.f2));
+                                }
+                                bufferedData.sort(Comparator.comparingDouble(o -> -o.f0));
+                                for (Tuple3<Double, Boolean, Double> dataPoint : bufferedData) {
+                                    out.collect(dataPoint);
+                                }
+                            }
+                        });
+
+        /* Calculates the summary of local data. */
+        DataStream<BinarySummary> partitionSummaries =
+                sortEvalData.transform(
+                        "reduceInEachPartition",
+                        TypeInformation.of(BinarySummary.class),
+                        new PartitionSummaryOperator());
+
+        /* Sorts global data. Output Tuple4 : <score, order, isPositive, weight>. */
+        DataStream<Tuple4<Double, Long, Boolean, Double>> dataWithOrders =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(sortEvalData),
+                        Collections.singletonMap(partitionSummariesKey, partitionSummaries),
+                        inputList -> {
+                            DataStream input = inputList.get(0);
+                            return input.flatMap(new CalcSampleOrders(partitionSummariesKey));
+                        });
+
+        DataStream<double[]> localAreaUnderROCVariable =
+                dataWithOrders.transform(
+                        "AccumulateMultiScore",
+                        TypeInformation.of(double[].class),
+                        new AccumulateMultiScoreOperator());
+
+        DataStream<double[]> middleAreaUnderROC =
+                DataStreamUtils.reduce(
+                        localAreaUnderROCVariable,
+                        (ReduceFunction<double[]>)
+                                (t1, t2) -> {
+                                    t2[0] += t1[0];
+                                    t2[1] += t1[1];
+                                    t2[2] += t1[2];
+                                    return t2;
+                                });
+
+        DataStream<Double> areaUnderROC =
+                middleAreaUnderROC.map(
+                        (MapFunction<double[], Double>)
+                                value -> {
+                                    if (value[1] > 0 && value[2] > 0) {
+                                        return (value[0] - 1. * value[1] * (value[1] + 1) / 2)
+                                                / (value[1] * value[2]);
+                                    } else {
+                                        return Double.NaN;
+                                    }
+                                });
+
+        Map<String, DataStream<?>> broadcastMap = new HashMap<>();
+        broadcastMap.put(partitionSummariesKey, partitionSummaries);
+        broadcastMap.put(AREA_UNDER_ROC, areaUnderROC);
+        DataStream<BinaryMetrics> localMetrics =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(sortEvalData),
+                        broadcastMap,
+                        inputList -> {
+                            DataStream input = inputList.get(0);
+                            return DataStreamUtils.mapPartition(
+                                    input, new CalcBinaryMetrics(partitionSummariesKey));
+                        });
+
+        DataStream<Map<String, Double>> metrics =
+                DataStreamUtils.mapPartition(localMetrics, new MergeMetrics());
+        metrics.getTransformation().setParallelism(1);
+
+        final String[] metricsNames = getMetricsNames();
+        TypeInformation<?>[] metricTypes = new TypeInformation[metricsNames.length];
+        Arrays.fill(metricTypes, Types.DOUBLE());
+        RowTypeInfo outputTypeInfo = new RowTypeInfo(metricTypes, metricsNames);
+
+        DataStream<Row> evalResult =
+                metrics.map(
+                        (MapFunction<Map<String, Double>, Row>)
+                                value -> {
+                                    Row ret = new Row(metricsNames.length);
+                                    for (int i = 0; i < metricsNames.length; ++i) {
+                                        ret.setField(i, value.get(metricsNames[i]));
+                                    }
+                                    return ret;
+                                },
+                        outputTypeInfo);
+        return new Table[] {tEnv.fromDataStream(evalResult)};
+    }
+
+    /** Updates variables for calculating AreaUnderROC. */
+    private static class AccumulateMultiScoreOperator extends AbstractStreamOperator<double[]>
+            implements OneInputStreamOperator<Tuple4<Double, Long, Boolean, Double>, double[]>,
+                    BoundedOneInput {
+        private ListState<double[]> accValueState;
+        private ListState<Double> scoreState;
+
+        double[] accValue;
+        double score;
+
+        @Override
+        public void endInput() {
+            if (accValue != null) {
+                output.collect(
+                        new StreamRecord<>(
+                                new double[] {
+                                    accValue[0] / accValue[1] * accValue[2],
+                                    accValue[2],
+                                    accValue[3]
+                                }));
+            }
+        }
+
+        @Override
+        public void processElement(
+                StreamRecord<Tuple4<Double, Long, Boolean, Double>> streamRecord) {
+            Tuple4<Double, Long, Boolean, Double> t = streamRecord.getValue();
+            if (accValue == null) {
+                accValue = new double[4];
+                score = t.f0;
+            } else if (score != t.f0) {
+                output.collect(
+                        new StreamRecord<>(
+                                new double[] {
+                                    accValue[0] / accValue[1] * accValue[2],
+                                    accValue[2],
+                                    accValue[3]
+                                }));
+                Arrays.fill(accValue, 0.0);
+            }
+            accValue[0] += t.f1;
+            accValue[1] += 1.0;
+            if (t.f2) {
+                accValue[2] += t.f3;
+            } else {
+                accValue[3] += t.f3;
+            }
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            accValueState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "accValueState", TypeInformation.of(double[].class)));
+            accValue =
+                    OperatorStateUtils.getUniqueElement(accValueState, "accValueState")
+                            .orElse(null);
+
+            scoreState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "scoreState", TypeInformation.of(Double.class)));
+            score = OperatorStateUtils.getUniqueElement(scoreState, "scoreState").orElse(0.0);
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            accValueState.clear();
+            scoreState.clear();
+            if (accValue != null) {
+                accValueState.add(accValue);
+                scoreState.add(score);
+            }
+        }
+    }
+
+    private static class PartitionSummaryOperator extends AbstractStreamOperator<BinarySummary>
+            implements OneInputStreamOperator<Tuple3<Double, Boolean, Double>, BinarySummary>,
+                    BoundedOneInput {
+        private ListState<BinarySummary> summaryState;
+        private BinarySummary summary;
+
+        @Override
+        public void endInput() {
+            if (summary != null) {
+                output.collect(new StreamRecord<>(summary));
+            }
+        }
+
+        @Override
+        public void processElement(StreamRecord<Tuple3<Double, Boolean, Double>> streamRecord) {
+            updateBinarySummary(summary, streamRecord.getValue());
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            summaryState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "summaryState",
+                                            TypeInformation.of(BinarySummary.class)));
+            summary =
+                    OperatorStateUtils.getUniqueElement(summaryState, "summaryState")
+                            .orElse(
+                                    new BinarySummary(
+                                            getRuntimeContext().getIndexOfThisSubtask(),
+                                            -Double.MAX_VALUE,
+                                            0,
+                                            0));
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            summaryState.clear();
+            if (summary != null) {
+                summaryState.add(summary);
+            }
+        }
+    }
+
+    /** Merges the metrics calculated locally and output metrics data. */
+    private static class MergeMetrics
+            implements MapPartitionFunction<BinaryMetrics, Map<String, Double>> {
+        @Override
+        public void mapPartition(
+                Iterable<BinaryMetrics> values, Collector<Map<String, Double>> out) {
+            Iterator<BinaryMetrics> iter = values.iterator();
+            BinaryMetrics reduceMetrics = iter.next();
+            while (iter.hasNext()) {
+                reduceMetrics = reduceMetrics.merge(iter.next());
+            }
+            Map<String, Double> map = new HashMap<>();
+            map.put(AREA_UNDER_ROC, reduceMetrics.areaUnderROC);
+            map.put(AREA_UNDER_PR, reduceMetrics.areaUnderPR);
+            map.put(AREA_UNDER_LORENZ, reduceMetrics.areaUnderLorenz);
+            map.put(KS, reduceMetrics.ks);
+            out.collect(map);
+        }
+    }
+
+    private static class CalcBinaryMetrics
+            extends RichMapPartitionFunction<Tuple3<Double, Boolean, Double>, BinaryMetrics> {
+        private final String partitionSummariesKey;
+
+        public CalcBinaryMetrics(String partitionSummariesKey) {
+            this.partitionSummariesKey = partitionSummariesKey;
+        }
+
+        @Override
+        public void mapPartition(
+                Iterable<Tuple3<Double, Boolean, Double>> iterable,
+                Collector<BinaryMetrics> collector) {
+
+            List<BinarySummary> statistics =
+                    getRuntimeContext().getBroadcastVariable(partitionSummariesKey);
+            long[] countValues =
+                    reduceBinarySummary(statistics, getRuntimeContext().getIndexOfThisSubtask());
+
+            double areaUnderROC =
+                    getRuntimeContext().<Double>getBroadcastVariable(AREA_UNDER_ROC).get(0);
+            long totalTrue = countValues[2];
+            long totalFalse = countValues[3];
+            if (totalTrue == 0) {
+                LOG.warn("There is no positive sample in data!");
+            }
+            if (totalFalse == 0) {
+                LOG.warn("There is no negative sample in data!");
+            }
+
+            BinaryMetrics metrics = new BinaryMetrics(0L, areaUnderROC);
+            double[] tprFprPrecision = new double[4];
+            for (Tuple3<Double, Boolean, Double> t3 : iterable) {
+                updateBinaryMetrics(t3, metrics, countValues, tprFprPrecision);
+            }
+            collector.collect(metrics);
+        }
+    }
+
+    private static void updateBinaryMetrics(
+            Tuple3<Double, Boolean, Double> cur,
+            BinaryMetrics binaryMetrics,
+            long[] countValues,
+            double[] recordValues) {
+        if (binaryMetrics.count == 0) {
+            recordValues[0] = countValues[2] == 0 ? 1.0 : 1.0 * countValues[0] / countValues[2];
+            recordValues[1] = countValues[3] == 0 ? 1.0 : 1.0 * countValues[1] / countValues[3];
+            recordValues[2] =
+                    countValues[0] + countValues[1] == 0
+                            ? 1.0
+                            : 1.0 * countValues[0] / (countValues[0] + countValues[1]);
+            recordValues[3] =
+                    1.0 * (countValues[0] + countValues[1]) / (countValues[2] + countValues[3]);
+        }
+
+        binaryMetrics.count++;
+        if (cur.f1) {
+            countValues[0]++;
+        } else {
+            countValues[1]++;
+        }
+
+        double tpr = countValues[2] == 0 ? 1.0 : 1.0 * countValues[0] / countValues[2];
+        double fpr = countValues[3] == 0 ? 1.0 : 1.0 * countValues[1] / countValues[3];
+        double precision =
+                countValues[0] + countValues[1] == 0
+                        ? 1.0
+                        : 1.0 * countValues[0] / (countValues[0] + countValues[1]);
+        double positiveRate =
+                1.0 * (countValues[0] + countValues[1]) / (countValues[2] + countValues[3]);
+
+        binaryMetrics.areaUnderLorenz +=
+                ((positiveRate - recordValues[3]) * (tpr + recordValues[0]) / 2);
+        binaryMetrics.areaUnderPR += ((tpr - recordValues[0]) * (precision + recordValues[2]) / 2);
+        binaryMetrics.ks = Math.max(Math.abs(fpr - tpr), binaryMetrics.ks);
+
+        recordValues[0] = tpr;
+        recordValues[1] = fpr;
+        recordValues[2] = precision;
+        recordValues[3] = positiveRate;
+    }
+
+    /**
+     * For each sample, calculates its score order among all samples. The sample with minimum score
+     * has order 1, while the sample with maximum score has order samples.
+     *
+     * <p>Input is a dataset of tuple (score, is real positive, weight), output is a dataset of
+     * tuple (score, order, is real positive, weight).
+     */
+    private static class CalcSampleOrders
+            extends RichFlatMapFunction<
+                    Tuple3<Double, Boolean, Double>, Tuple4<Double, Long, Boolean, Double>> {
+        private long startIndex;
+        private long total = -1;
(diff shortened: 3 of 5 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorParams.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Params of BinaryClassificationEvaluator.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface BinaryClassificationEvaluatorParams<T>
+        extends HasLabelCol<T>, HasRawPredictionCol<T>, HasWeightCol<T> {
+    /**
+     * param for metric names in evaluation (supports 'areaUnderROC', 'areaUnderPR', 'KS' and
```
lindong28, line 36: Would it be useful to explain the definition of each metric here?

Note that Spark ML provides the following Java doc in `BinaryClassificationMetrics.scala` [1]:

areaUnderROC: the area under the receiver operating characteristic (ROC) curve.

areaUnderPR: the area under the precision-recall curve.

[1] https://spark.apache.org/docs/2.0.2/api/java/org/apache/spark/mllib/evaluation/BinaryClassificationMetrics.html

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple3<Double, Boolean, Double>> evalData =
+                tEnv.toDataStream(inputs[0])
+                        .map(new ParseSample(getLabelCol(), getRawPredictionCol(), getWeightCol()));
+
+        DataStream<Tuple4<Double, Boolean, Double, Integer>> evalDataWithTaskId =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(evalData),
+                        Collections.singletonMap(BOUNDARY_RANGE, getBoundaryRange(evalData)),
```
lindong28, line 111: nits: Would it be simpler to define `boundaryRange` as a local variable in this method, similar to how we define `broadcastModelKey` in `LogisticRegressionModel::transform(...)`?

Same for `partitionSummaries`.

The program is typically simpler and easier to read by minimizing the class private (or global) variables.

lindong28 (commented):

The implementation seems to be longer and more complex than Spark's implementation. I am wondering if we can refactor it to be simpler or more readable. Maybe @yunfengzhou-hub and @zhipeng93 can also comment on this.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+        @Override
+        public void endInput() {
+            if (summary != null) {
+                output.collect(new StreamRecord<>(summary));
+            }
+        }
+
+        @Override
+        public void processElement(StreamRecord<Tuple3<Double, Boolean, Double>> streamRecord) {
+            if (summary == null) {
```
yunfengzhou-hub, line 283: Shall we move this logic to the `orElse(null)` used in `initializeState`?

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+                localAucVariable
+                        .transform(
+                                "reduceInEachPartition",
+                                TypeInformation.of(double[].class),
+                                new CalcAucOperator())
+                        .transform(
+                                "reduceInFinalPartition",
+                                TypeInformation.of(double[].class),
+                                new CalcAucOperator())
+                        .setParallelism(1)
```
yunfengzhou-hub, line 213: I noticed that in several places we have firstly used an operation for each subtask or partition, followed by a global reduce/aggregation. I believe improving map-reduce's operations' efficiency in this way is globally applicable to all Flink operators, such that Flink should have implemented relevant mechanism for it, and we can directly use `reduce()`/`aggregate()` without implementing it on our own. Could you please help check it?

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+                                                    negativeSum += t.f3;
+                                                }
+                                            }
+                                            out.collect(
+                                                    new double[] {
+                                                        1. * sum / cnt * positiveSum,
+                                                        positiveSum,
+                                                        negativeSum
+                                                    });
+                                        })
```
yunfengzhou-hub, line 200: Shall we move anonymous functions like this into private static classes, give them meaningful class names and JavaDocs? That might help improve readability.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+            Tuple2<Boolean, long[]> t =
+                    reduceBinarySummary(statistics, getRuntimeContext().getIndexOfThisSubtask());
+            long[] countValues = t.f1;
+
+            double areaUnderROC =
+                    getRuntimeContext().<Double>getBroadcastVariable(AREA_UNDER_ROC).get(0);
+            long totalTrue = countValues[2];
+            long totalFalse = countValues[3];
+            if (totalTrue == 0) {
+                System.out.println("There is no positive sample in data!");
```
yunfengzhou-hub, line 409: It might be better to avoid using `println`, instead, we can use `LOG` to record such information.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+
+        binaryMetrics.areaUnderLorenz +=
+                ((positiveRate - recordValues[3]) * (tpr + recordValues[0]) / 2);
+        binaryMetrics.areaUnderPR += ((tpr - recordValues[0]) * (precision + recordValues[2]) / 2);
+        binaryMetrics.ks = Math.max(Math.abs(fpr - tpr), binaryMetrics.ks);
+
+        recordValues[0] = tpr;
+        recordValues[1] = fpr;
+        recordValues[2] = precision;
+        recordValues[3] = positiveRate;
```
yunfengzhou-hub, line 464: Shall we define a class to represent the metric results? It would bring better readability than remembering the meaning of each metric by their index in a double array.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+                        new AppendMaxWatermark(x -> x));
+
+        DataStream<double[]> localAucVariable =
+                dataWithOrders
+                        .keyBy(
+                                (KeySelector<Tuple4<Double, Long, Boolean, Double>, Double>)
+                                        value -> value.f0)
+                        .window(EndOfStreamWindows.get())
+                        .apply(
+                                (WindowFunction<
```
yunfengzhou-hub, line 174: Implementations like this brings the risk of OOM, as data allocated to a single subtask can still be larger than memory space. It might be unnecessary to use `apply()` here, and we can consider using `reduce()` or `aggregate()`.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,736 @@
...
+                                                    / (aucVariable[1] * aucVariable[2]);
+                                        } else {
+                                            return Double.NaN;
+                                        }
+                                    }
+                                });
+
+        Map<String, DataStream<?>> broadcastMap = new HashMap<>();
+        broadcastMap.put(PARTITION_SUMMARY, partitionSummaries);
+        broadcastMap.put(AREA_UNDER_ROC, areaUnderROC);
```
yunfengzhou-hub, line 233: I noticed that `areaUnderROC` is calculated separately from other metrics. Given that all metrics experience similarly to be computed, like a `mapPartition` or `window().apply()`, I have the sense that we may have just one window operator or map partition operator that gets all metrics out. That would greatly simplify the structure of this code. How do you like the idea to try in this way?

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,208 @@
...
+                new BinaryClassificationEvaluator()
+                        .setMetricsNames("areaUnderPR", "KS", "areaUnderROC")
+                        .setLabelCol("label")
+                        .setRawPredictionCol("raw");
+        Table evalResult = eval.transform(trainDataTable)[0];
+        DataStream<Row> dataStream = tEnv.toDataStream(evalResult);
+        List<Row> results = IteratorUtils.toList(dataStream.executeAndCollect());
+        Row result = results.get(0);
+        for (int i = 0; i < EXPECTED_DATA.length; ++i) {
+            assertEquals(EXPECTED_DATA[i], result.getFieldAs(i), 1.0e-5);
```
yunfengzhou-hub, line 175: Shall we also check the output column names, instead of just the value at each index?

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,208 @@
...
+                .setMetricsNames("areaUnderROC")
+                .setWeightCol("weight");
+        assertEquals("labelCol", binaryEval.getLabelCol());
+        assertEquals("weight", binaryEval.getWeightCol());
+        assertEquals("raw", binaryEval.getRawPredictionCol());
+        assertArrayEquals(new String[] {"areaUnderROC"}, binaryEval.getMetricsNames());
+    }
+
+    @Test
+    public void testSaveLoadAndTransform() throws Exception {
```
yunfengzhou-hub, line 146: The order of test cases in this class and the naming convention seems to be different from other existing operators. Shall we follow the test style in other classes?

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,208 @@
...
+public class BinaryClassificationEvaluatorTest {
+    @Rule public final TemporaryFolder tempFolder = new TemporaryFolder();
+    private StreamTableEnvironment tEnv;
+    private Table trainDataTable;
+    private Table trainDataTableWithMultiScore;
+    private Table trainDataTableWithWeight;
+
+    private static final List<Row> TRAIN_DATA =
+            new ArrayList<>(
+                    Arrays.asList(
```
yunfengzhou-hub, line 56: It seems that we do not need `new ArrayList<>()`.

yunfengzhou-hub (commented):
Thanks for the update. Some comments are as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorParams.java
```diff
@@ -0,0 +1,53 @@
...
+     * 'areaUnderLorenz').
+     */
+    Param<String[]> METRICS_NAMES =
+            new StringArrayParam(
+                    "metricsNames",
+                    "Names of output metrics. The array element must be 'areaUnderROC', 'areaUnderPR', 'KS' and 'areaUnderLorenz'",
+                    new String[] {"areaUnderROC", "areaUnderPR"},
+                    ParamValidators.nonEmptyArray());
```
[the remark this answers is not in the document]
weibozhao: I agree with it.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
weibozhao: I have refine the naming format. This algo is different from other algo, then the type also has some different properties.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
weibozhao: AreaUnderROC's calculation is different from others. It need keyed by score and do some calculation, other metrics just do some statistics of the input data.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
weibozhao: I have defined a class (BinaryMetrics) to represent the metrics results. This double array stores the middle values which is used to  calculate the metrics results.

flink-ml-core/src/test/java/org/apache/flink/ml/api/StageTest.java
```diff
@@ -463,5 +463,10 @@ public void testValidators() {
         Assert.assertTrue(nonEmptyArray.validate(new String[] {"1"}));
         Assert.assertFalse(nonEmptyArray.validate(null));
         Assert.assertFalse(nonEmptyArray.validate(new String[0]));
+
+        ParamValidator<String[]> isSubArray = ParamValidators.isSubArray("a", "b", "c");
+        Assert.assertFalse(isSubArray.validate(new String[] {"c", "v"}));
```
lindong28, line 468: nits: could we also check the case where the input value is null, to be consistent with the above tests?

`Assert.assertFalse(isSubArray.validate(null))`

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorParams.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface BinaryClassificationEvaluatorParams<T>
+        extends HasLabelCol<T>, HasRawPredictionCol<T>, HasWeightCol<T> {
+    /**
+     * param for metric names in evaluation (supports 'areaUnderROC', 'areaUnderPR', 'KS' and
+     * 'areaUnderLorenz').
+     *
+     * <p>areaUnderROC: the area under the receiver operating characteristic (ROC) curve.
```
lindong28, line 39: nits: could we update the Java doc here to follow the format used in `HasHandleInvalid.java`? It seems that the format in `HasHandleInvalid.java` is more readable.

Currently the doc for `KS` and `areaUnderLorenz` are on the same lines.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,216 @@
...
+                    Row.of(0.0, Vectors.dense(0.4, 0.6), 1.3),
+                    Row.of(1.0, Vectors.dense(0.1, 0.9), 1.5),
+                    Row.of(1.0, Vectors.dense(0.1, 0.9), 1.4),
+                    Row.of(0.0, Vectors.dense(0.6, 0.4), 0.3),
+                    Row.of(0.0, Vectors.dense(0.7, 0.3), 0.5),
+                    Row.of(1.0, Vectors.dense(0.1, 0.9), 1.9),
+                    Row.of(0.0, Vectors.dense(0.8, 0.2), 1.2),
+                    Row.of(1.0, Vectors.dense(0.9, 0.1), 1.0));
+
+    private static final double[] EXPECTED_DATA =
```
lindong28, line 98: Since we check double value equality using `delta=1.0e-5`, would it be simpler to reduce the precision of the expected values here accordingly?

Same for other expected values initialized in this file.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+/** Tests {@link BinaryClassificationEvaluator}. */
+public class BinaryClassificationEvaluatorTest {
+    @Rule public final TemporaryFolder tempFolder = new TemporaryFolder();
+    private StreamTableEnvironment tEnv;
+    private Table trainDataTable;
+    private Table trainDataTableWithMultiScore;
+    private Table trainDataTableWithWeight;
+
+    private static final List<Row> TRAIN_DATA =
```
lindong28, line 53: Hmm... since `BinaryClassification` does not do any training (i.e. fit), would it be more intuitive to rename this as `INPUT_DATA*`?

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,216 @@
...
+        assertEquals("raw", binaryEval.getRawPredictionCol());
+        assertArrayEquals(new String[] {"areaUnderROC"}, binaryEval.getMetricsNames());
+    }
+
+    @Test
+    public void testSaveLoadAndEvaluate() throws Exception {
+        BinaryClassificationEvaluator eval =
+                new BinaryClassificationEvaluator()
+                        .setMetricsNames("areaUnderPR", "KS", "areaUnderROC")
+                        .setLabelCol("label")
```
lindong28, line 146: Since the goal of this test is to test save/load/transform, would it be simpler to keep using the default label parameter value (i.e. `label`) without explicitly setting it? Similarly, can we keep using the default value for rawPredictionCol?

Same for other tests whose purpose is not to test the parameter set/get.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,216 @@
...
+                new String[] {"areaUnderPR", "KS", "areaUnderROC"},
+                evalResult.getResolvedSchema().getColumnNames().toArray());
+        Row result = results.get(0);
+        for (int i = 0; i < EXPECTED_DATA.length; ++i) {
+            assertEquals(EXPECTED_DATA[i], result.getFieldAs(i), 1.0e-5);
+        }
+    }
+
+    @Test
+    public void testEvaluateWithMultiScore() throws Exception {
```
lindong28, line 182: Could you help explain why we call this test `testEvaluateWithMultiScore` and call the test above `testEvaluate`? It seems that both tests use multiple evaluation metrics and their inputs have the same format.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,731 @@
...
+import java.util.Random;
+
+import static org.apache.flink.runtime.blob.BlobWriter.LOG;
+
+/**
+ * Calculates the evaluation metrics for binary classification. The input data has columns
+ * rawPrediction, label and an optional weight column. The output may contain different metrics
+ * which will be defined by parameter MetricsNames. See @BinaryClassificationEvaluatorParams.
+ */
+public class BinaryClassificationEvaluator
```
lindong28, line 81: According to Spark's Java doc, its `BinaryClassificationEvaluator` supports rawPrediction column of type double (binary 0/1 prediction, or probability of label 1), in addition to the rawPrediction column of type vector (length-2 vector of raw predictions, scores, or label probabilities).

Should we also support rawPrediction column of type double? If not, how do we handle those use-case that currently requires rawPrediction column of type double in Spark?

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,731 @@
...
+        implements AlgoOperator<BinaryClassificationEvaluator>,
+                BinaryClassificationEvaluatorParams<BinaryClassificationEvaluator> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final int NUM_SAMPLE_FOR_RANGE_PARTITION = 100;
+    private static final String AREA_UNDER_ROC = "areaUnderROC";
```
lindong28, line 86: Would it be better to move these variables to `BinaryClassificationEvaluatorParams` and make them `public`, so that users can set the parameter by referencing those variables instead of manually typing a string?

Manually typing a string is a bit more error-prone and IDE won't be able to give hint in this case.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
weibozhao: I have support these two format rawPrediction

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
weibozhao: `AccumulateMultiScore` will process the samples with same score, if the samples has same score more than one, then the action is different.

yunfengzhou-hub (commented):
Thanks for the update. Some comments are as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+                                Tuple4<Double, Boolean, Double, Integer>,
+                                Tuple3<Double, Boolean, Double>>() {
+                            @Override
+                            public void mapPartition(
+                                    Iterable<Tuple4<Double, Boolean, Double, Integer>> values,
+                                    Collector<Tuple3<Double, Boolean, Double>> out) {
+                                List<Tuple3<Double, Boolean, Double>> bufferedData =
+                                        new LinkedList<>();
+                                for (Tuple4<Double, Boolean, Double, Integer> t4 : values) {
+                                    bufferedData.add(Tuple3.of(t4.f0, t4.f1, t4.f2));
```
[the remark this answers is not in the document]
weibozhao: I think there is no way to avoid sorting action.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+ * rawPrediction, label and an optional weight column. The rawPrediction can be of type double
+ * (binary 0/1 prediction, or probability of label 1) or of type vector (length-2 vector of raw
+ * predictions, scores, or label probabilities). The output may contain different metrics which will
+ * be defined by parameter MetricsNames. See @BinaryClassificationEvaluatorParams.
+ */
+public class BinaryClassificationEvaluator
+        implements AlgoOperator<BinaryClassificationEvaluator>,
+                BinaryClassificationEvaluatorParams<BinaryClassificationEvaluator> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final int NUM_SAMPLE_FOR_RANGE_PARTITION = 100;
```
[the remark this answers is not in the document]
weibozhao: This variable used by partitionSort, just used to repartition the data to workers. But every worker does sorting action locally. 100 just a empirical value.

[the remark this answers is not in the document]
weibozhao: New test for this parameters is not need. This parameter is just a inner parameter for partitionSort.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+
+    public BinaryClassificationEvaluator() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Table[] transform(Table... inputs) {
```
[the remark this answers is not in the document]
weibozhao: The input data will be applied two actions: sort action and statistics action. Sort action need buffer data but statistics not need. I  don't think merging these to one may improving performance.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Tuple3<Double, Boolean, Double>> evalData =
+                tEnv.toDataStream(inputs[0])
+                        .map(new ParseSample(getLabelCol(), getRawPredictionCol(), getWeightCol()));
+        final String boundaryRangeKey = "boundaryRange";
+        final String partitionSummariesKey = "partitionSummaries";
```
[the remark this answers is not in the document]
weibozhao: I use a static final variable last version. But LinDong suggests use local variable just as LogisticRegerssion done.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+                        inputList -> {
+                            DataStream input = inputList.get(0);
+                            return input.flatMap(new CalcSampleOrders(partitionSummariesKey));
+                        });
+
+        dataWithOrders =
+                dataWithOrders.transform(
+                        "appendMaxWaterMark",
+                        dataWithOrders.getType(),
+                        new AppendMaxWatermark(x -> x));
```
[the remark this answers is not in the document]
weibozhao: If this code removed, algorithm will get an err. I do this as CongZhou told me.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+                                TypeInformation.of(double[].class),
+                                new AucOperator())
+                        .setParallelism(1);
+
+        DataStream<Double> areaUnderROC =
+                middleAreaUnderROC.map(
+                        (MapFunction<double[], Double>)
+                                value -> {
+                                    if (value[1] > 0 && value[2] > 0) {
+                                        return (value[0] - 1. * value[1] * (value[1] + 1) / 2)
```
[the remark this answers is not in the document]
weibozhao: some value in doubleArray has no meaning, just a middle value in calculating AUC. If you want to know the meaning of variables, you can read the code which calculate these variables.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+                                    for (int i = 0; i < metricsNames.length; ++i) {
+                                        ret.setField(i, value.get(metricsNames[i]));
+                                    }
+                                    return ret;
+                                },
+                        outputTypeInfo);
+        return new Table[] {tEnv.fromDataStream(evalResult)};
+    }
+
+    /** Updates variables for calculating Auc. */
```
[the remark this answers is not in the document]
weibozhao: code has removed

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,783 @@
...
+
+        @Override
+        public void mapPartition(
+                Iterable<BinaryMetrics> values, Collector<Map<String, Double>> out) {
+            Iterator<BinaryMetrics> iter = values.iterator();
+            BinaryMetrics reduceMetrics = iter.next();
+            while (iter.hasNext()) {
+                reduceMetrics = reduceMetrics.merge(iter.next());
+            }
+            Map<String, Double> map = new HashMap<>();
```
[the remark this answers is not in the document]
weibozhao: I don‘t agree with you.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+/** Tests {@link BinaryClassificationEvaluator}. */
+public class BinaryClassificationEvaluatorTest {
```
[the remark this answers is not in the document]
weibozhao: The evaluation of data which all labels are 1.0 or 0.0 has no meaning.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. Thanks.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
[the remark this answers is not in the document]
yunfengzhou-hub: I tried removing this code, but all `BinaryClassificationEvaluatorTest` cases still passed. If this code is irreplaceable, could you please add tests to verify its function?

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. Maybe we could do possible optimizations after this PR in the next phase.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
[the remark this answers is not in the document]
yunfengzhou-hub: Spark BinaryClassification algorithm have test cases for the following situations. Could you please add corresponding test cases?
- `rawPredictionCol`'s data type is double
- `label`'s data type is any possible numeric type, like int, shorts or decimal.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,211 @@
...
+        BinaryClassificationEvaluator binaryEval = new BinaryClassificationEvaluator();
+        assertEquals("label", binaryEval.getLabelCol());
+        assertNull(binaryEval.getWeightCol());
+        assertEquals("rawPrediction", binaryEval.getRawPredictionCol());
+        assertArrayEquals(
+                new String[] {"areaUnderROC", "areaUnderPR"}, binaryEval.getMetricsNames());
+        binaryEval
+                .setLabelCol("labelCol")
+                .setRawPredictionCol("raw")
+                .setMetricsNames("areaUnderROC")
```
yunfengzhou-hub, line 136: nit: It might be better to use the constants we have already defined. For example, `setMetricsNames(BinaryClassificationEvaluatorParams.AREA_UNDER_ROC)`

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,211 @@
...
+                new BinaryClassificationEvaluator()
+                        .setMetricsNames("areaUnderPR", "ks", "areaUnderROC");
+        BinaryClassificationEvaluator loadedEval =
+                StageTestUtils.saveAndReload(tEnv, eval, tempFolder.newFolder().getAbsolutePath());
+        Table evalResult = loadedEval.transform(inputDataTable)[0];
+        DataStream<Row> dataStream = tEnv.toDataStream(evalResult);
+        assertArrayEquals(
+                new String[] {"areaUnderPR", "ks", "areaUnderROC"},
+                evalResult.getResolvedSchema().getColumnNames().toArray());
+        List<Row> results = IteratorUtils.toList(dataStream.executeAndCollect());
```
yunfengzhou-hub, line 156: nit: `List<Row> results = IteratorUtils.toList(evalResult.execute().collect());` is enough. `tEnv.toDataStream` is unnecessary.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,211 @@
...
+        assertEquals("weight", binaryEval.getWeightCol());
+        assertEquals("raw", binaryEval.getRawPredictionCol());
+        assertArrayEquals(new String[] {"areaUnderROC"}, binaryEval.getMetricsNames());
+    }
+
+    @Test
+    public void testSaveLoadAndEvaluate() throws Exception {
+        BinaryClassificationEvaluator eval =
+                new BinaryClassificationEvaluator()
+                        .setMetricsNames("areaUnderPR", "ks", "areaUnderROC");
```
yunfengzhou-hub, line 148: `areaUnderLorenz` is not tested in these test cases yet.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassificationevaluator/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,730 @@
...
+            }
+        }
+
+        @Override
+        public void processElement(
+                StreamRecord<Tuple4<Double, Long, Boolean, Double>> streamRecord) {
+            Tuple4<Double, Long, Boolean, Double> t = streamRecord.getValue();
+            if (accValue == null) {
+                accValue = new double[4];
+                score = t.f0;
```
yunfengzhou-hub, line 253: score is not checkpointed or restored. When there is failover, stream records of different scores might be accumulated in the same `accValue`.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassificationevaluator/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,730 @@
...
+        recordValues[1] = fpr;
+        recordValues[2] = precision;
+        recordValues[3] = positiveRate;
+    }
+
+    /**
+     * For each sample, calculates its score order among all samples. The sample with minimum score
+     * has order 1, while the sample with maximum score has order samples.
+     *
+     * <p>Input is a dataset of tuple (score, is real positive, wight), output is a dataset of tuple
```
yunfengzhou-hub, line 451: "weight".

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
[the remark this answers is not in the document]
yunfengzhou-hub: Got it.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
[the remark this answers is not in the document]
yunfengzhou-hub: Maybe we can dive deeper into performance optimization chances later after this PR. Please feel free to close this conversation for now.

yunfengzhou-hub (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluatorTest.java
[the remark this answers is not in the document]
weibozhao: OK, I will add these test case.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluatorParams.java
```diff
@@ -0,0 +1,67 @@
...
+     *   <li>areaUnderPR: the area under the precision-recall curve.
+     *   <li>ks: Kolmogorov-Smirnov, measures the ability of the model to separate positive and
+     *       negative samples.
+     *   <li>areaUnderLorenz: the area under the lorenz curve.
+     * </ul>
+     */
+    Param<String[]> METRICS_NAMES =
+            new StringArrayParam(
+                    "metricsNames",
+                    "Names of output metrics, which may contains 'areaUnderROC', 'areaUnderPR', 'ks' or 'areaUnderLorenz'",
```
yunfengzhou-hub, line 55: nit: We don't need to present possible values in the description string. "Names of output metrics." is enough. Please check `HasHandleInvalid` as an example.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,245 @@
...
+            };
+    private static final double EXPECTED_DATA_W = 0.8911680911680911;
+    private static final double EPS = 1.0e-5;
+
+    @Before
+    public void before() {
+        Configuration config = new Configuration();
+        config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment(config);
+        env.setParallelism(3);
```
yunfengzhou-hub, line 129: Thanks for fixing the problem here. Could we also add a test to verify that the algorithm can still work even if parallelism is larger than number of data? `LogisticRegressionTest.testMoreSubtaskThanData` is a good example. Or reviewers would have to manually change this to `env.setParallelism(20)` and re run the tests.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,245 @@
...
+                new String[] {"areaUnderPR", "ks", "areaUnderROC"},
+                evalResult.getResolvedSchema().getColumnNames().toArray());
+        Row result = results.get(0);
+        for (int i = 0; i < EXPECTED_DATA.length; ++i) {
+            assertEquals(EXPECTED_DATA[i], result.getFieldAs(i), EPS);
+        }
+    }
+
+    @Test
+    public void testEvaluateWithDoubleRaw() throws Exception {
```
yunfengzhou-hub, line 201: Thanks for adding this test. In Spark's BinaryClassification algorithm, it also tests when label's data type is any possible numeric type, like int, shorts or decimal. Could we also add these tests?

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,245 @@
...
+        BinaryClassificationEvaluator binaryEval = new BinaryClassificationEvaluator();
+        assertEquals("label", binaryEval.getLabelCol());
+        assertNull(binaryEval.getWeightCol());
+        assertEquals("rawPrediction", binaryEval.getRawPredictionCol());
+        assertArrayEquals(
+                new String[] {"areaUnderROC", "areaUnderPR"}, binaryEval.getMetricsNames());
+        binaryEval
+                .setLabelCol("labelCol")
+                .setRawPredictionCol("raw")
+                .setMetricsNames(BinaryClassificationEvaluatorParams.AREA_UNDER_ROC)
```
yunfengzhou-hub, line 158: Thanks for making this change. There are some other places that have directly used the string values like ”"areaUnderRoc" in BinaryClassificationEvaluatorTest. Shall we replace all of them into constants like `AREA_UNDER_ROC`?

nit: Besides, statically importing the constants might make code looks better.
```java

import static org.apache.flink.ml.evaluation.binaryclassfication.BinaryClassificationEvaluatorParams.AREA_UNDER_ROC;
...
.setMetricsNames(AREA_UNDER_ROC)
```

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,245 @@
...
+                .setWeightCol("weight");
+        assertEquals("labelCol", binaryEval.getLabelCol());
+        assertEquals("weight", binaryEval.getWeightCol());
+        assertEquals("raw", binaryEval.getRawPredictionCol());
+        assertArrayEquals(new String[] {"areaUnderROC"}, binaryEval.getMetricsNames());
+    }
+
+    @Test
+    public void testSaveLoadAndEvaluate() throws Exception {
```
yunfengzhou-hub, line 167: nit: Shall we adjust the order of these test cases to follow other classes? For example, First is `testParam`, followed by `testOutputSchema`, `testEvaluate` and then save/load and other special cases.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,245 @@
...
+                new String[] {"areaUnderPR", "ks", "areaUnderROC"},
+                evalResult.getResolvedSchema().getColumnNames().toArray());
+        Row result = results.get(0);
+        for (int i = 0; i < EXPECTED_DATA.length; ++i) {
+            assertEquals(EXPECTED_DATA[i], result.getFieldAs(i), EPS);
+        }
+    }
+
+    @Test
+    public void testEvaluateWithMultiScore() throws Exception {
```
yunfengzhou-hub, line 217: nit: Shall we check and clean up warnings in `BinaryClassificationEvaluator` and `BinaryClassificationEvaluatorTest`? For example, Exception is never thrown in this method.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,742 @@
...
+            accValue =
+                    OperatorStateUtils.getUniqueElement(accValueState, "accValueState")
+                            .orElse(null);
+
+            scoreState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "scoreState", TypeInformation.of(Double.class)));
+            score = OperatorStateUtils.getUniqueElement(scoreState, "scoreState").orElse(0.0);
```
yunfengzhou-hub, line 293: score has not been saved in snapshot yet before restoring from operator state.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,742 @@
...
+import java.util.Random;
+
+import static org.apache.flink.runtime.blob.BlobWriter.LOG;
+
+/**
+ * An Estimator which calculates the evaluation metrics for binary classification. The input data
+ * has columns rawPrediction, label and an optional weight column. The rawPrediction can be of type
+ * double (binary 0/1 prediction, or probability of label 1) or of type vector (length-2 vector of
+ * raw predictions, scores, or label probabilities). The output may contain different metrics which
+ * will be defined by parameter MetricsNames. See @BinaryClassificationEvaluatorParams.
```
yunfengzhou-hub, line 78: nit: `{@link BinaryClassificationEvaluatorParams}`. `parameter MetricsNames` might be outdated.

yunfengzhou-hub (commented):
Thanks for the updates. Left some comments as below.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
weibozhao: I have add test for label‘s data type is int or double. I think it's not needed test all the types.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryeval/BinaryClassificationEvaluator.java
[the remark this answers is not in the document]
weibozhao: I have remove this code.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,721 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+import static org.apache.flink.runtime.blob.BlobWriter.LOG;
```
yunfengzhou-hub, line 69: We should use `org.slf4j.Logger` directly. `AbstractBroadcastWrapperOperator` or `Benchmark` could serve as an example.

yunfengzhou-hub (commented):
Thanks for the update. LGTM in general. Left one small comment.

lindong28:
Thanks for the update. LGTM. Will merge this PR after all comments are addressed.

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,724 @@
...
+ * double (binary 0/1 prediction, or probability of label 1) or of type vector (length-2 vector of
+ * raw predictions, scores, or label probabilities). The output may contain different metrics which
+ * will be defined by parameter MetricsNames. See {@link BinaryClassificationEvaluatorParams}.
+ */
+public class BinaryClassificationEvaluator
+        implements AlgoOperator<BinaryClassificationEvaluator>,
+                BinaryClassificationEvaluatorParams<BinaryClassificationEvaluator> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final int NUM_SAMPLE_FOR_RANGE_PARTITION = 100;
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractWrapperOperator.class);
```
zhipeng93, line 85: nit: should be BinaryClassificationEvaluator.class

flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
```diff
@@ -100,4 +100,22 @@ public boolean validate(T value) {
     public static <T> ParamValidator<T[]> nonEmptyArray() {
         return value -> value != null && value.length > 0;
     }
+
+    // Check if every element in the array-typed parameter value is in the array of allowed values.
```
zhipeng93, line 104: nit: Check -> Checks

zhipeng93, line 104: nit: Check -> Checks

flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
```diff
@@ -100,4 +100,22 @@ public boolean validate(T value) {
...
+    public static <T> ParamValidator<T[]> isSubArray(T... allowed) {
```
zhipeng93, line 105: nit: `subArray` seems a bit confusing here since `subArray` indicates that the order should also be the same with the `allowed`, which is not true. How about `isSubSet`?

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassfication/BinaryClassificationEvaluator.java
```diff
@@ -0,0 +1,724 @@
...
+                        });
+
+        /* Calculates the summary of local data. */
+        DataStream<BinarySummary> partitionSummaries =
+                sortEvalData.transform(
+                        "reduceInEachPartition",
+                        TypeInformation.of(BinarySummary.class),
+                        new PartitionSummaryOperator());
+
+        /* Sorts global data. Output Tuple4 : <score, order, isPositive, weight> */
```
zhipeng93, line 146: nit: The comment should ends with `.`.

flink-ml-lib/src/test/java/org/apache/flink/ml/evaluation/BinaryClassificationEvaluatorTest.java
```diff
@@ -0,0 +1,306 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+/** Tests {@link BinaryClassificationEvaluator}. */
+public class BinaryClassificationEvaluatorTest {
+    @Rule public final TemporaryFolder tempFolder = new TemporaryFolder();
```
zhipeng93, line 48: nit: public -> private

zhipeng93 (commented):
Sorry for the late review. 

Thanks for the PR @weibozhao ! LGTM overall. Only left some minor comments below.

PR #87, opened 2022-04-20
[FLINK-27096] Improve Benchmark Framework

yunfengzhou-hub:
- Add python scripts for batch generating & analyzing benchmark
- Enrich benchmark result file's content with input parameters
- Support saving benchmark results every time a benchmark is completed


flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -75,33 +75,43 @@ public static void printHelp() {
...
-        }
+            LOG.info("\n" + BenchmarkUtils.getResultsMapAsJson(result));
+
+            benchmarkResultsJson =
+                    BenchmarkUtils.getResultsMapAsJson(results.toArray(new BenchmarkResult[0]));
 
-        String benchmarkResultsJson =
-                BenchmarkUtils.getResultsMapAsJson(results.toArray(new BenchmarkResult[0]));
+            if (commandLine.hasOption(OUTPUT_FILE_OPTION.getLongOpt())) {
+                ReadWriteUtils.saveToFile(saveFile, benchmarkResultsJson, true);
```
lindong28, line 104: Instead of saving all benchmark results in the loop body, would it be more efficient to save benchmark results after the while loop is done?

We probably also need to record in the `results.json` the fact that a benchmark fails, probably with a string indicating the the cause of the failure. The `benchmark-results-visualize.py` would need to handle this information probably in the graph (e.g. with a red `X`) in the figure.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/BenchmarkResult.java
```diff
@@ -18,13 +18,25 @@
...
+import org.apache.flink.ml.benchmark.datagenerator.InputDataGenerator;
 import org.apache.flink.util.Preconditions;
 
 /** The result of executing a benchmark. */
 public class BenchmarkResult {
     /** The benchmark name. */
     public final String name;
 
+    /** The stage to execute benchmark on. */
+    public final Stage<?> stage;
```
lindong28, line 32: Instead of keeping the stage/inputDataGenerator/modelDataGenerator instances in `BenchmarkResult`, would it be simpler to just pass the original `Map<String, ?> params` (which is used to instantiate these instances) to `BenchmarkUtils.getResultsMapAsJson(...)` directly?

flink-ml-benchmark/README.md
```diff
@@ -194,3 +189,9 @@ stage. The stage is benchmarked against 10000 randomly generated vectors.
   }
 }
 ```
+
+### Benchmark Results Visualization
+
+`benchmark-results-visualize.py` is provided as a helper script to visualize
```
lindong28, line 195: Could we provide commands that users can copy and paste to visualize the `results.json` in this quickstart?

lindong28 (commented):

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
yunfengzhou-hub: According to offline discussion, we need to save the results in loop body to avoid losing all benchmark results when one of the benchmarks failed.

I have added the codes to save exception information in result json file, but I'm not sure whether and how to visualize exceptions on the plot. Given that they can be saved as exception for various reasons, like failing during the Flink job's execution and failing during parameter parsing, there seems no proper place for error information in a plot.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/BenchmarkResult.java
yunfengzhou-hub: I added a `toMap` function to `BenchmarkResult` and a `ExceptionToMap` static method to `BenchmarkUtils`. Please see if this solution is good enough.

yunfengzhou-hub:

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/BenchmarkResult.java
lindong28: Thanks for the update. The solution looks much better now.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -71,37 +70,57 @@ public static void printHelp() {
...
-            BenchmarkResult result =
-                    BenchmarkUtils.runBenchmark(
-                            tEnv, benchmark.getKey(), (Map<String, ?>) benchmark.getValue());
-
-            results.add(result);
-            LOG.info(BenchmarkUtils.getResultsMapAsJson(result));
+        Map<String, Map<String, Map<String, ?>>> results = new HashMap<>();
+        String benchmarkResultsJson = "{}";
+        int index = 0;
+        for (String benchmarkName : benchmarks.keySet()) {
```
lindong28, line 86: Since we need to get both the key and the value of this entry, would it be simpler to use `for (Map.Entry<String, Map<String, Map<String, ?>>> entry : benchmarks.entrySet())` here? 

Then we can use `entry.getKey()` and `entry.getValue()`, instead of repeatedly calling `results.get(benchmarkName)` and `benchmarks.get(benchmarkName)`, in the loop body.

flink-ml-dist/src/main/flink-ml-bin/bin/benchmark-results-visualize.py
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+################################################################################
+import argparse
+import json
+import matplotlib.pyplot as plt
+import re
+
+
+def benchmark_results_visualize(filename, name_pattern, x_field, y_field):
+    def get_field(json, fieldnames):
```
lindong28, line 25: Would it be simpler to move this method outside `benchmark_results_visualize`, to reduce the number of nesting?

flink-ml-dist/src/main/flink-ml-bin/bin/benchmark-results-visualize.py
```diff
@@ -0,0 +1,68 @@
...
+        field = json
```
lindong28, line 26: Would it be simpler to just rename the first function input parameter as `field`?

flink-ml-core/src/main/java/org/apache/flink/ml/util/ReadWriteUtils.java
```diff
@@ -71,9 +71,9 @@ private static <T> Object jsonEncodeHelper(Param<T> param, Object value) throws
         return param.jsonEncode((T) value);
     }
 
-    // Converts Map<Param<?>, Object> to Map<String, String> which maps the parameter name to the
-    // string-encoded parameter value.
-    private static Map<String, Object> jsonEncode(Map<Param<?>, Object> paramMap)
+    // Converts Map<Param<?>, Object> to Map<String, Object> which maps the parameter name to the
+    // json-supported parameter value.
+    public static Map<String, Object> jsonEncode(Map<Param<?>, Object> paramMap)
```
lindong28, line 76: Does this need to be `public`?

lindong28 (commented):

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
lindong28: Hmm.. I think we can still save all benchmark results outside the loop, when one benchmark fails, by catching the exception from the benchmark failure.

yunfengzhou-hub:

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -31,34 +31,33 @@
...
 
-    static final Options OPTIONS =
+    private static final Options OPTIONS =
             new Options().addOption(HELP_OPTION).addOption(OUTPUT_FILE_OPTION);
 
-    public static void printHelp() {
+    private static void printHelp() {
         HelpFormatter formatter = new HelpFormatter();
         formatter.setLeftPadding(5);
         formatter.setWidth(80);
```
lindong28, line 63: Should we update the line below to use the new script name?

And we can simplify the code a bit by using `\n` in `"./benchmark-run.sh <config-file-path> [OPTIONS]\n"` instead of using `System.out.println()`.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -71,46 +70,59 @@ public static void printHelp() {
...
+                benchmarkMap.put(
+                        "results",
+                        Collections.singletonMap(
+                                "exception",
+                                String.format(
+                                        "%s(%s:%s)",
+                                        e,
+                                        e.getStackTrace()[0].getFileName(),
+                                        e.getStackTrace()[0].getLineNumber())));
+                LOG.info(String.format("Benchmark %s failed.\n%s", benchmarkName, e));
```
lindong28, line 108: Should this be `LOG.error(...)`?

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -71,46 +70,59 @@ public static void printHelp() {
...
+        String benchmarkResultsJson =
+                ReadWriteUtils.OBJECT_MAPPER
+                        .writerWithDefaultPrettyPrinter()
+                        .writeValueAsString(benchmarks);
         if (commandLine.hasOption(OUTPUT_FILE_OPTION.getLongOpt())) {
-            String saveFile = commandLine.getOptionValue(OUTPUT_FILE_OPTION.getLongOpt());
             ReadWriteUtils.saveToFile(saveFile, benchmarkResultsJson, true);
             System.out.println("Benchmark results saved as json in " + saveFile + ".");
+        } else {
+            System.out.println("Benchmark results summary:");
```
lindong28, line 121: Would it be simpler and more consistent with the `LOG.info(...)` above by using the following code:

`System.out.println(String.format("Benchmark results summary:\n%s", benchmarkResultsJson))`

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -120,7 +132,10 @@ public static void main(String[] args) throws Exception {
         } else if (commandLine.getArgs().length == 1) {
             executeBenchmarks(commandLine);
         } else {
-            printInvalidError(args);
+            System.out.println("Invalid command line arguments " + Arrays.toString(args));
```
lindong28, line 135: Would it be simpler and more consistent with the LOG.info(...) above by using the following code:

`System.out.println(String.format("Invalid command line arguments %s \n", Arrays.toString(args)))`

flink-ml-dist/src/main/flink-ml-bin/bin/benchmark-results-visualize.py
```diff
@@ -0,0 +1,68 @@
...
+import re
+
+
+def get_nested_field_value(nested_fields, field_names):
+    for field_name in field_names:
+        nested_fields = nested_fields[field_name]
+    return nested_fields
+
+
+def benchmark_results_visualize(file_name, name_pattern, x_field, y_field):
```
lindong28, line 30: Would it be a bit more intuitive to name this method as `visualize_benchmark_results(...)`?

flink-ml-benchmark/README.md
```diff
@@ -63,59 +63,89 @@ Then in Flink ML's binary distribution's folder, execute the following command
 to run the default benchmarks.
 
 ```bash
-./bin/flink-ml-benchmark.sh ./conf/benchmark-conf.json --output-file results.json
+./bin/benchmark-run.sh ./conf/benchmark-conf.json --output-file results.json
```
lindong28, line 66: Would it be simpler and more readable to use the following command

```
./bin/benchmark-run.sh conf/benchmark-conf.json --output-file results.json
```

lindong28 (commented):
Thanks for the update. Looks good overall. Left just some minor comments.

PR #92, opened 2022-04-27
[hotfix] Update .asf.yaml file to disable the merge button and update collaborators etc.

lindong28:
This PR updates the `.asf.yaml` file to disable the merge button, update collaborators, and add project labels.

See https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features for an explanation of asf.yaml file and its entries' semantics.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,3 +1,17 @@
+github:
+  enabled_merge_buttons:
+    squash: true
+    merge: false
+    rebase: true
+  collaborators:
+    - flinkbot
+  labels:
+    - flink
+    - machine-learning
+    - ml
+    - big-data
+    - java
+    - python
 notifications:
   commits:      commits@flink.apache.org
   issues:       issues@flink.apache.org


lindong28:
@zhipeng93 could you review this PR?

.asf.yaml
```diff
@@ -1,3 +1,17 @@
...
+    squash: true
+    merge: false
+    rebase: true
+  labels:
+    - flink
+    - big-data
+    - java
+    - scala
+    - python
+    - sql
```
zhipeng93, line 12: Shall we update the labels here? For example, add `machine learning`, remove `scala` and `sql`?

zhipeng93 (commented):
Thanks for the PR. Just two minor comments: Should we avoid the last `dot` in the commit log?

.asf.yaml
lindong28: Thanks for the suggestion. Good catch.

I have updated the PR to remove the label additions. The reason is that apache/kafka and apache/flink-statefun do not have `labels` in their `.asf.yaml`. And I also could not find any concrete usage of these labels in apache/flink's pull requests.

The commit message has been updated as appropriate.

PR #97, opened 2022-05-10
[FLINK-27096] Improve DataCache and KMeans performance

yunfengzhou-hub:

This PR mainly improves `DataCacheReader` and `DataCacheWriter`'s performance with memory caches, and further optimizes KMeans algorithm with DataCache mechanism.

## Brief change log
- In DataCache mechanism, provides the option to cache records in memory in serialized format, instead of directly saving  them to file system, unless checkpoint is invoked or there is not enough memory.
- Changes `DataStreamUtils.mapParition()`'s implementation from using `ListState` to using `DataCacheWriter`.
- Utilizes the memory-caching functionality in `KMeans.SelectNearestCentroidOperator` to avoid OOM.
    - All other existing usages of `DataCacheWriter` is untouched, continuing to directly saving records to file system.
- Adds `DataStreamUtils.sample()` and uses it in `KMeans.selectRandomCentroids()` to reduce memory usage.

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
@@ -26,25 +26,36 @@ import org.apache.flink.api.common.functions.ReduceFunction;
 import org.apache.flink.api.common.state.ListState;
 import org.apache.flink.api.common.state.ListStateDescriptor;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeutils.base.IntSerializer;
+import org.apache.flink.api.dag.Transformation;
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.api.java.typeutils.TypeExtractor;
+import org.apache.flink.core.memory.ManagedMemoryUseCase;
+import org.apache.flink.iteration.datacache.nonkeyed.ListStateWithCache;
 import org.apache.flink.iteration.operator.OperatorStateUtils;
 import org.apache.flink.runtime.state.StateInitializationContext;
 import org.apache.flink.runtime.state.StateSnapshotContext;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.functions.windowing.AllWindowFunction;
+import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
 import org.apache.flink.streaming.api.operators.AbstractUdfStreamOperator;
 import org.apache.flink.streaming.api.operators.BoundedOneInput;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.api.operators.TimestampedCollector;
 import org.apache.flink.streaming.api.windowing.windows.GlobalWindow;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.table.api.TableException;
 import org.apache.flink.util.Collector;
 
 import org.apache.commons.collections.IteratorUtils;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
+import java.util.Random;
 
 /** Provides utility functions for {@link DataStream}. */
 @Internal
@@ -106,6 +117,56 @@ public class DataStreamUtils {
     }
 
     /**
+     * Performs a uniform sampling over the elements in a bounded data stream.
+     *
+     * <p>This method takes samples without replacement. If the number of elements in the stream is
+     * smaller than expected number of samples, all elements will be included in the sample.
+     *
+     * @param input The input data stream.
+     * @param numSamples The number of elements to be sampled.
+     * @param randomSeed The seed to randomly pick elements as sample.
+     * @return A data stream containing a list of the sampled elements.
+     */
+    public static <T> DataStream<T> sample(DataStream<T> input, int numSamples, long randomSeed) {
+        int inputParallelism = input.getParallelism();
+
+        return input.transform(
+                        "samplingOperator",
+                        input.getType(),
+                        new SamplingOperator<>(numSamples, randomSeed))
+                .setParallelism(inputParallelism)
+                .transform(
+                        "samplingOperator",
+                        input.getType(),
+                        new SamplingOperator<>(numSamples, randomSeed))
+                .setParallelism(1)
+                .map(x -> x, input.getType())
+                .setParallelism(inputParallelism);
+    }
+
+    /**
+     * Sets {Transformation#declareManagedMemoryUseCaseAtOperatorScope(ManagedMemoryUseCase, int)}
+     * using the given bytes for {@link ManagedMemoryUseCase#OPERATOR}.
+     *
+     * <p>This method is in reference to Flink's ExecNodeUtil.setManagedMemoryWeight. The provided
+     * bytes should be in the same scale as existing usage in Flink, for example,
+     * StreamExecWindowAggregate.WINDOW_AGG_MEMORY_RATIO.
+     */
+    public static <T> void setManagedMemoryWeight(
+            Transformation<T> transformation, long memoryBytes) {
+        if (memoryBytes > 0) {
+            final int weightInMebibyte = Math.max(1, (int) (memoryBytes >> 20));
+            final Optional<Integer> previousWeight =
+                    transformation.declareManagedMemoryUseCaseAtOperatorScope(
+                            ManagedMemoryUseCase.OPERATOR, weightInMebibyte);
+            if (previousWeight.isPresent()) {
+                throw new TableException(
+                        "Managed memory weight has been set, this should not happen.");
+            }
+        }
+    }
+
+    /**
      * A stream operator to apply {@link MapPartitionFunction} on each partition of the input
      * bounded data stream.
      */
@@ -113,7 +174,7 @@ public class DataStreamUtils {
             extends AbstractUdfStreamOperator<OUT, MapPartitionFunction<IN, OUT>>
             implements OneInputStreamOperator<IN, OUT>, BoundedOneInput {
 
-        private ListState<IN> valuesState;
+        private ListStateWithCache<IN> valuesState;
 
         public MapPartitionOperator(MapPartitionFunction<IN, OUT> mapPartitionFunc) {
             super(mapPartitionFunc);
@@ -122,23 +183,31 @@ public class DataStreamUtils {
         @Override
         public void initializeState(StateInitializationContext context) throws Exception {
             super.initializeState(context);
-            ListStateDescriptor<IN> descriptor =
-                    new ListStateDescriptor<>(
-                            "inputState",
-                            getOperatorConfig()
-                                    .getTypeSerializerIn(0, getClass().getClassLoader()));
-            valuesState = context.getOperatorStateStore().getListState(descriptor);
+
+            valuesState =
+                    new ListStateWithCache<>(
+                            getOperatorConfig().getTypeSerializerIn(0, getClass().getClassLoader()),
+                            getContainingTask(),
+                            getRuntimeContext(),
+                            context,
+                            config.getOperatorID());
+        }
+
+        @Override
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            valuesState.snapshotState(context);
+        }
+
+        @Override
+        public void processElement(StreamRecord<IN> input) throws Exception {
+            valuesState.add(input.getValue());
         }
 
         @Override
         public void endInput() throws Exception {
             userFunction.mapPartition(valuesState.get(), new TimestampedCollector<>(output));
             valuesState.clear();
-        }
-
-        @Override
-        public void processElement(StreamRecord<IN> input) throws Exception {
-            valuesState.add(input.getValue());
         }
     }
 
@@ -176,7 +245,7 @@ public class DataStreamUtils {
             state =
                     context.getOperatorStateStore()
                             .getListState(
-                                    new ListStateDescriptor<T>(
+                                    new ListStateDescriptor<>(
                                             "state",
                                             getOperatorConfig()
                                                     .getTypeSerializerIn(
@@ -253,6 +322,82 @@ public class DataStreamUtils {
             for (; i < downStreamParallelism; i++) {
                 collector.collect(Tuple2.of(i, Arrays.copyOfRange(values, offset, offset + size)));
                 offset += size;
+            }
+        }
+    }
+
+    /*
+     * A stream operator that takes a randomly sampled subset of elements in a bounded data stream.
+     */
+    private static class SamplingOperator<T> extends AbstractStreamOperator<T>
+            implements OneInputStreamOperator<T, T>, BoundedOneInput {
+        private final int numSamples;
+
+        private final Random random;
+
+        private ListState<T> samplesState;
+
+        private List<T> samples;
+
+        private ListState<Integer> countState;
+
+        private int count;
+
+        SamplingOperator(int numSamples, long randomSeed) {
+            this.numSamples = numSamples;
+            this.random = new Random(randomSeed);
+        }
+
+        @Override
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+
+            ListStateDescriptor<T> samplesDescriptor =
+                    new ListStateDescriptor<>(
+                            "samplesState",
+                            getOperatorConfig()
+                                    .getTypeSerializerIn(0, getClass().getClassLoader()));
+            samplesState = context.getOperatorStateStore().getListState(samplesDescriptor);
+            samples = new ArrayList<>(numSamples);
+            samplesState.get().forEach(samples::add);
+
+            ListStateDescriptor<Integer> countDescriptor =
+                    new ListStateDescriptor<>("countState", IntSerializer.INSTANCE);
+            countState = context.getOperatorStateStore().getListState(countDescriptor);
+            Iterator<Integer> countIterator = countState.get().iterator();
+            if (countIterator.hasNext()) {
+                count = countIterator.next();
+            } else {
+                count = 0;
+            }
+        }
+
+        @Override
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            samplesState.update(samples);
+            countState.update(Collections.singletonList(count));
+        }
+
+        @Override
+        public void processElement(StreamRecord<T> streamRecord) throws Exception {
+            T value = streamRecord.getValue();
+            count++;
+
+            if (samples.size() < numSamples) {
+                samples.add(value);
+            } else {
+                int index = random.nextInt(count);
+                if (index < numSamples) {
+                    samples.set(index, value);
+                }
+            }
+        }
+
+        @Override
+        public void endInput() throws Exception {
+            for (T sample : samples) {
+                output.collect(new StreamRecord<>(sample));
             }
         }
     }
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/typeinfo/DenseVectorSerializer.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/typeinfo/DenseVectorSerializer.java
@@ -20,27 +20,34 @@
 package org.apache.flink.ml.linalg.typeinfo;
 
 import org.apache.flink.api.common.typeutils.SimpleTypeSerializerSnapshot;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.api.common.typeutils.TypeSerializerSnapshot;
-import org.apache.flink.api.common.typeutils.base.TypeSerializerSingleton;
 import org.apache.flink.core.memory.DataInputView;
 import org.apache.flink.core.memory.DataOutputView;
 import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.util.Bits;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Objects;
 
 /** Specialized serializer for {@link DenseVector}. */
-public final class DenseVectorSerializer extends TypeSerializerSingleton<DenseVector> {
+public final class DenseVectorSerializer extends TypeSerializer<DenseVector> {
 
     private static final long serialVersionUID = 1L;
 
     private static final double[] EMPTY = new double[0];
 
-    public static final DenseVectorSerializer INSTANCE = new DenseVectorSerializer();
+    private final byte[] buf = new byte[1024];
 
     @Override
     public boolean isImmutableType() {
         return false;
+    }
+
+    @Override
+    public TypeSerializer<DenseVector> duplicate() {
+        return new DenseVectorSerializer();
     }
 
     @Override
@@ -75,9 +82,14 @@ public final class DenseVectorSerializer extends TypeSerializerSingleton<DenseVe
 
         final int len = vector.values.length;
         target.writeInt(len);
+
         for (int i = 0; i < len; i++) {
-            target.writeDouble(vector.get(i));
+            Bits.putDouble(buf, i << 3, vector.values[i]);
+            if ((i & 127) == 127) {
+                target.write(buf);
+            }
         }
+        target.write(buf, 0, (len & 127) << 3);
     }
 
     @Override
@@ -89,10 +101,17 @@ public final class DenseVectorSerializer extends TypeSerializerSingleton<DenseVe
     }
 
     // Reads `len` double values from `source` into `dst`.
-    private static void readDoubleArray(double[] dst, DataInputView source, int len)
-            throws IOException {
-        for (int i = 0; i < len; i++) {
-            dst[i] = source.readDouble();
+    private void readDoubleArray(double[] dst, DataInputView source, int len) throws IOException {
+        int index = 0;
+        for (int i = 0; i < (len >> 7); i++) {
+            source.read(buf, 0, 1024);
+            for (int j = 0; j < 128; j++) {
+                dst[index++] = Bits.getDouble(buf, j << 3);
+            }
+        }
+        source.read(buf, 0, (len << 3) & 1023);
+        for (int j = 0; j < (len & 127); j++) {
+            dst[index++] = Bits.getDouble(buf, j << 3);
         }
     }
 
@@ -116,6 +135,16 @@ public final class DenseVectorSerializer extends TypeSerializerSingleton<DenseVe
         target.write(source, len * 8);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof DenseVectorSerializer;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(DenseVectorSerializer.class);
+    }
+
     // ------------------------------------------------------------------------
 
     @Override
@@ -129,7 +158,7 @@ public final class DenseVectorSerializer extends TypeSerializerSingleton<DenseVe
             extends SimpleTypeSerializerSnapshot<DenseVector> {
 
         public DenseVectorSerializerSnapshot() {
-            super(() -> INSTANCE);
+            super(DenseVectorSerializer::new);
         }
     }
 }
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
@@ -20,10 +20,6 @@ package org.apache.flink.iteration.datacache.nonkeyed;
 
 import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.api.java.tuple.Tuple2;
-import org.apache.flink.core.fs.FSDataInputStream;
-import org.apache.flink.core.fs.FileSystem;
-import org.apache.flink.core.memory.DataInputView;
-import org.apache.flink.core.memory.DataInputViewStreamWrapper;
 
 import javax.annotation.Nullable;
 
@@ -31,37 +27,38 @@ import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
 
-/** Reads the cached data from a list of paths. */
+/** Reads the cached data from a list of segments. */
 public class DataCacheReader<T> implements Iterator<T> {
 
+    /** The tool to deserialize bytes into records. */
     private final TypeSerializer<T> serializer;
 
-    private final FileSystem fileSystem;
-
+    /** The segments where to read the records from. */
     private final List<Segment> segments;
 
-    @Nullable private SegmentReader currentSegmentReader;
+    /** The current reader for next records. */
+    @Nullable private SegmentReader<T> currentSegmentReader;
 
-    public DataCacheReader(
-            TypeSerializer<T> serializer, FileSystem fileSystem, List<Segment> segments)
-            throws IOException {
-        this(serializer, fileSystem, segments, new Tuple2<>(0, 0));
+    /** The index of the segment that current reader reads from. */
+    private int currentSegmentIndex;
+
+    /** The number of records that have been read through the current reader so far. */
+    private int currentSegmentCount;
+
+    public DataCacheReader(TypeSerializer<T> serializer, List<Segment> segments) {
+        this(serializer, segments, Tuple2.of(0, 0));
     }
 
     public DataCacheReader(
             TypeSerializer<T> serializer,
-            FileSystem fileSystem,
             List<Segment> segments,
-            Tuple2<Integer, Integer> readerPosition)
-            throws IOException {
-
+            Tuple2<Integer, Integer> readerPosition) {
         this.serializer = serializer;
-        this.fileSystem = fileSystem;
         this.segments = segments;
+        this.currentSegmentIndex = readerPosition.f0;
+        this.currentSegmentCount = readerPosition.f1;
 
-        if (readerPosition.f0 < segments.size()) {
-            this.currentSegmentReader = new SegmentReader(readerPosition.f0, readerPosition.f1);
-        }
+        createSegmentReader(readerPosition.f0, readerPosition.f1);
     }
 
     @Override
@@ -72,68 +69,42 @@ public class DataCacheReader<T> implements Iterator<T> {
     @Override
     public T next() {
         try {
-            T next = currentSegmentReader.next();
+            T record = currentSegmentReader.next();
 
+            currentSegmentCount++;
             if (!currentSegmentReader.hasNext()) {
                 currentSegmentReader.close();
-                if (currentSegmentReader.index < segments.size() - 1) {
-                    currentSegmentReader = new SegmentReader(currentSegmentReader.index + 1, 0);
-                } else {
-                    currentSegmentReader = null;
-                }
+                currentSegmentIndex++;
+                currentSegmentCount = 0;
+                createSegmentReader(currentSegmentIndex, currentSegmentCount);
             }
 
-            return next;
+            return record;
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
     }
 
     public Tuple2<Integer, Integer> getPosition() {
-        if (currentSegmentReader == null) {
-            return new Tuple2<>(segments.size(), 0);
-        }
-
-        return new Tuple2<>(currentSegmentReader.getIndex(), currentSegmentReader.getOffset());
+        return new Tuple2<>(currentSegmentIndex, currentSegmentCount);
     }
 
-    private class SegmentReader {
+    private void createSegmentReader(int index, int startOffset) {
+        try {
+            if (index >= segments.size()) {
+                currentSegmentReader = null;
+                return;
+            }
 
-        private final int index;
+            Segment segment = segments.get(currentSegmentIndex);
+            if (!segment.getCache().isEmpty()) {
+                currentSegmentReader = new MemorySegmentReader<>(serializer, segment, startOffset);
+            } else {
+                currentSegmentReader = new FileSegmentReader<>(serializer, segment, startOffset);
+            }
 
-        private final FSDataInputStream inputStream;
-
-        private final DataInputView inputView;
-
-        private int offset;
-
-        public SegmentReader(int index, int startOffset) throws IOException {
-            this.index = index;
-            this.inputStream = fileSystem.open(segments.get(index).getPath());
-            this.inputView = new DataInputViewStreamWrapper(inputStream);
-            this.offset = startOffset;
-        }
-
-        public boolean hasNext() {
-            return offset < segments.get(index).getCount();
-        }
-
-        public T next() throws IOException {
-            T next = serializer.deserialize(inputView);
-            offset++;
-            return next;
-        }
-
-        public void close() throws IOException {
-            inputStream.close();
-        }
-
-        public int getIndex() {
-            return index;
-        }
-
-        public int getOffset() {
-            return offset;
+        } catch (IOException e) {
+            throw new RuntimeException(e);
         }
     }
 }
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheSnapshot.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheSnapshot.java
@@ -25,10 +25,13 @@ import org.apache.flink.core.fs.FSDataOutputStream;
 import org.apache.flink.core.fs.FileSystem;
 import org.apache.flink.core.fs.Path;
 import org.apache.flink.core.memory.DataInputViewStreamWrapper;
+import org.apache.flink.runtime.memory.MemoryAllocationException;
 import org.apache.flink.runtime.util.NonClosingInputStreamDecorator;
 import org.apache.flink.runtime.util.NonClosingOutputStreamDecorator;
 import org.apache.flink.statefun.flink.core.feedback.FeedbackConsumer;
+import org.apache.flink.table.runtime.util.MemorySegmentPool;
 import org.apache.flink.util.IOUtils;
+import org.apache.flink.util.Preconditions;
 import org.apache.flink.util.function.SupplierWithException;
 
 import org.apache.commons.io.input.BoundedInputStream;
@@ -41,20 +44,26 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 import static org.apache.flink.util.Preconditions.checkState;
 
-/** The snapshot of a data cache. It could be written out or read from an external stream.O */
+/** The snapshot of a data cache. It could be written out or read from an external stream. */
 public class DataCacheSnapshot {
 
+    /** The version of DataCacheSnapshot format. */
     private static final int CURRENT_VERSION = 1;
 
+    /** The file system that contains the cache files. */
     private final FileSystem fileSystem;
 
+    /**
+     * An optional position of DataCacheReader. If not set, recovered readers would read the
+     * segments from the beginning.
+     */
     @Nullable private final Tuple2<Integer, Integer> readerPosition;
 
+    /** The segments to be snapshot or recovered from. */
     private final List<Segment> segments;
 
     public DataCacheSnapshot(
@@ -64,6 +73,9 @@ public class DataCacheSnapshot {
         this.fileSystem = fileSystem;
         this.readerPosition = readerPosition;
         this.segments = segments;
+        for (Segment segment : segments) {
+            Preconditions.checkArgument(segment.getFsSize() > 0);
+        }
     }
 
     public FileSystem getFileSystem() {
@@ -79,6 +91,7 @@ public class DataCacheSnapshot {
         return segments;
     }
 
+    /** Writes the information about this data cache to an output stream. */
     public void writeTo(OutputStream checkpointOutputStream) throws IOException {
         try (DataOutputStream dos =
                 new DataOutputStream(new NonClosingOutputStreamDecorator(checkpointOutputStream))) {
@@ -95,13 +108,10 @@ public class DataCacheSnapshot {
                 serializeSegments(segments, dos);
             } else {
                 // We have to copy the whole streams.
-                int totalRecords = segments.stream().mapToInt(Segment::getCount).sum();
-                long totalSize = segments.stream().mapToLong(Segment::getSize).sum();
-                checkState(totalRecords >= 0, "overflowed: " + totalRecords);
-                dos.writeInt(totalRecords);
-                dos.writeLong(totalSize);
-
+                dos.writeInt(segments.size());
                 for (Segment segment : segments) {
+                    dos.writeInt(segment.getCount());
+                    dos.writeLong(segment.getFsSize());
                     try (FSDataInputStream inputStream = fileSystem.open(segment.getPath())) {
                         IOUtils.copyBytes(inputStream, checkpointOutputStream, false);
                     }
@@ -110,10 +120,13 @@ public class DataCacheSnapshot {
         }
     }
 
+    /**
+     * Replays cached records in the data cache from the input stream into the target feedback
+     * consumer.
+     */
     public static <T> void replay(
             InputStream checkpointInputStream,
             TypeSerializer<T> serializer,
-            FileSystem fileSystem,
             FeedbackConsumer<T> feedbackConsumer)
             throws Exception {
         try (DataInputStream dis =
@@ -127,23 +140,27 @@ public class DataCacheSnapshot {
             boolean isDistributedFS = dis.readBoolean();
             if (isDistributedFS) {
                 List<Segment> segments = deserializeSegments(dis);
-                DataCacheReader<T> dataCacheReader =
-                        new DataCacheReader<>(serializer, fileSystem, segments);
+
+                DataCacheReader<T> dataCacheReader = new DataCacheReader<>(serializer, segments);
+
                 while (dataCacheReader.hasNext()) {
                     feedbackConsumer.processFeedback(dataCacheReader.next());
                 }
             } else {
                 DataInputViewStreamWrapper dataInputView = new DataInputViewStreamWrapper(dis);
-                int totalRecords = dis.readInt();
-                // Ignore the total size.
-                dis.readLong();
-                for (int i = 0; i < totalRecords; ++i) {
-                    feedbackConsumer.processFeedback(serializer.deserialize(dataInputView));
+                int segmentNum = dis.readInt();
+                for (int i = 0; i < segmentNum; i++) {
+                    int count = dis.readInt();
+                    dis.readLong();
+                    for (int j = 0; j < count; j++) {
+                        feedbackConsumer.processFeedback(serializer.deserialize(dataInputView));
+                    }
                 }
             }
         }
     }
 
+    /** Recovers a data cache instance from the input stream. */
     public static DataCacheSnapshot recover(
             InputStream checkpointInputStream,
             FileSystem fileSystem,
@@ -167,23 +184,66 @@ public class DataCacheSnapshot {
             if (isDistributedFS) {
                 segments = deserializeSegments(dis);
             } else {
-                int totalRecords = dis.readInt();
-                long totalSize = dis.readLong();
+                int segmentNum = dis.readInt();
+                segments = new ArrayList<>(segmentNum);
+                for (int i = 0; i < segmentNum; i++) {
+                    int count = dis.readInt();
+                    long fsSize = dis.readLong();
+                    Path path = pathGenerator.get();
+                    try (FSDataOutputStream outputStream =
+                            fileSystem.create(path, FileSystem.WriteMode.NO_OVERWRITE)) {
 
-                Path path = pathGenerator.get();
-                try (FSDataOutputStream outputStream =
-                        fileSystem.create(path, FileSystem.WriteMode.NO_OVERWRITE)) {
-
-                    BoundedInputStream inputStream =
-                            new BoundedInputStream(checkpointInputStream, totalSize);
-                    inputStream.setPropagateClose(false);
-                    IOUtils.copyBytes(inputStream, outputStream, false);
-                    inputStream.close();
+                        BoundedInputStream boundedInputStream =
+                                new BoundedInputStream(checkpointInputStream, fsSize);
+                        boundedInputStream.setPropagateClose(false);
+                        IOUtils.copyBytes(boundedInputStream, outputStream, false);
+                        boundedInputStream.close();
+                    }
+                    segments.add(new Segment(path, count, fsSize));
                 }
-                segments = Collections.singletonList(new Segment(path, totalRecords, totalSize));
             }
 
             return new DataCacheSnapshot(fileSystem, readerPosition, segments);
+        }
+    }
+
+    /**
+     * Attempts to cache the segments in memory.
+     *
+     * <p>The attempt is made at segment granularity, which means there might be only part of the
+     * segments are cached.
+     *
+     * <p>This method does not throw exceptions if there is not enough memory space for caching a
+     * segment.
+     */
+    public <T> void tryReadSegmentsToMemory(
+            TypeSerializer<T> serializer, MemorySegmentPool segmentPool) throws IOException {
+        boolean cacheSuccess;
+        for (Segment segment : segments) {
+            if (!segment.getCache().isEmpty()) {
+                continue;
+            }
+
+            SegmentReader<T> reader = new FileSegmentReader<>(serializer, segment, 0);
+            SegmentWriter<T> writer;
+            try {
+                writer =
+                        new MemorySegmentWriter<>(
+                                serializer, segment.getPath(), segmentPool, segment.getFsSize());
+            } catch (MemoryAllocationException e) {
+                break;
+            }
+
+            cacheSuccess = true;
+            while (cacheSuccess && reader.hasNext()) {
+                if (!writer.addRecord(reader.next())) {
+                    writer.finish().ifPresent(x -> segmentPool.returnAll(x.getCache()));
+                    cacheSuccess = false;
+                }
+            }
+            if (cacheSuccess) {
+                segment.setCache(writer.finish().get().getCache());
+            }
         }
     }
 
@@ -201,10 +261,10 @@ public class DataCacheSnapshot {
     private static void serializeSegments(List<Segment> segments, DataOutputStream dataOutputStream)
             throws IOException {
         dataOutputStream.writeInt(segments.size());
-        for (int i = 0; i < segments.size(); ++i) {
-            dataOutputStream.writeUTF(segments.get(i).getPath().toString());
-            dataOutputStream.writeInt(segments.get(i).getCount());
-            dataOutputStream.writeLong(segments.get(i).getSize());
+        for (Segment segment : segments) {
+            dataOutputStream.writeUTF(segment.getPath().toString());
+            dataOutputStream.writeInt(segment.getCount());
+            dataOutputStream.writeLong(segment.getFsSize());
         }
     }
 
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
@@ -19,38 +19,59 @@
 package org.apache.flink.iteration.datacache.nonkeyed;
 
 import org.apache.flink.api.common.typeutils.TypeSerializer;
-import org.apache.flink.core.fs.FSDataOutputStream;
 import org.apache.flink.core.fs.FileSystem;
 import org.apache.flink.core.fs.Path;
-import org.apache.flink.core.memory.DataOutputView;
-import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
+import org.apache.flink.runtime.memory.MemoryAllocationException;
+import org.apache.flink.table.runtime.util.MemorySegmentPool;
+import org.apache.flink.util.Preconditions;
 import org.apache.flink.util.function.SupplierWithException;
+
+import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
 
-/** Records the data received and replayed them on required. */
+/** Records the data received and replayed them when required. */
 public class DataCacheWriter<T> {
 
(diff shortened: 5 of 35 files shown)


flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,86 @@
...
+        this.memoryManager = memoryManager;
+    }
+
+    @Override
+    public boolean addRecord(T record) throws IOException {
+        if (!MemoryUtils.isMemoryEnoughForCache(memoryManager)) {
+            return false;
+        }
+
+        long recordSize = GraphLayout.parseInstance(record).totalSize();
```
lindong28, line 62: Does Spark ML also estimate the record size in this way? Given that we invoke this method for every record, we probably need to make sure its overhead is low enough even for small records.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,49 @@
...
+import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.core.fs.Path;
 
+import java.io.IOException;
 import java.io.Serializable;
+import java.util.List;
 import java.util.Objects;
 
-/** A segment represents a single file for the cache. */
+/** A segment contains the information about a cache unit. */
```
lindong28, line 29: After this PR, the Java doc says this class represents a cache unit. But the implementation shows that this class could represent a file, which seems confusing. Could you clarify it?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
```diff
@@ -20,120 +20,121 @@
...
+
+            if (!currentReader.hasNext()) {
+                currentReader.close();
+                if (cacheWriter != null) {
+                    cacheWriter
+                            .finish()
+                            .ifPresent(
+                                    x -> {
+                                        x.fsSize = segments.get(segmentIndex).fsSize;
+                                        segments.set(segmentIndex, x);
```
lindong28, line 120: This method will replace a file-only segment with in-memory segment. It is a bit counter-intuitive to make such changes in a XXXReader class.

lindong28 (commented):

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
yunfengzhou-hub: Spark uses its own classes, like `SizeTracker` and `SizeEstimator`, to calculate the size of each object. If we follow Spark's practice, we would also need to add such infra codes, which could add certain workload to this PR.

`GraphLayout` uses `java.lang.instrument.Instrumentation`, which is the default solution for Java programs to get object size, so I think its overhead should be acceptable.

The reason why I did not use `Instrumentation` is because that would require users add extra java options while executing flink ml programs, for example, 
```shell
java -javaagent:target/flink-ml-agent-2.1-SNAPSHOT.jar flink-ml-uber.jar <className>
```
The `-javaagent` option is additionally required. This more complex cmd requirement would decrease user experience while not bringing much performance improvement.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/FsSegmentWriter.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private int count;
+
+    public FsSegmentWriter(TypeSerializer<T> serializer, Path path) throws IOException {
+        this.serializer = serializer;
+        this.path = path;
+        this.fileSystem = path.getFileSystem();
+        this.outputStream = fileSystem.create(path, FileSystem.WriteMode.NO_OVERWRITE);
+        this.byteArrayOutputStream = new ByteArrayOutputStream();
+        this.objectOutputStream = new ObjectOutputStream(outputStream);
```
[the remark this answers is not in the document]
yunfengzhou-hub: My experiments shows that if we do `outputView = new DataOutputViewStreamWrapper(outputStream)`, the performance of disk IO would degrade obviously. One possible cause is that in `DataOutputViewStreamWrapper` or `FsDataOutputStream`'s subclass, the implementation flushes out bytes every time `write()` is invoked, as I can observe from the update frequency of `outputStream.getPos()`, but I have not dived further into this issue. Shall we leave this implementation as it is for now, and see if @zhipeng93 would offer a more simplified implementation with better performance? I'll leave a TODO here for now.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
yunfengzhou-hub: A reader would not modify data stored in file or metadata about that file. It can just create a cache of the data in a segment, and assign the cache metadata to that segment. I'll modify the implementation to better reflect this.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -89,57 +104,31 @@ public List<Segment> getFinishSegments() {
...
     public void cleanup() throws IOException {
-        finishCurrentSegment();
+        finish();
         for (Segment segment : finishSegments) {
-            fileSystem.delete(segment.getPath(), false);
+            if (segment.isOnDisk()) {
+                fileSystem.delete(segment.path, false);
+            }
+            if (segment.isInMemory()) {
+                memoryManager.releaseMemory(segment.path, segment.inMemorySize);
```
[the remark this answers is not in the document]
yunfengzhou-hub: Memory is reserved in `MemorySegmentWriter`. I had set the key to the writer object, rather than segment.path. I'll correct this.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemoryUtils.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.flink.iteration.datacache.nonkeyed;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.runtime.memory.MemoryManager;
+
+/** Utility variables and methods for memory operation. */
+@Internal
+class MemoryUtils {
+    // Cache is not suggested if over 80% of memory has been occupied.
+    private static final double CACHE_MEMORY_THRESHOLD = 0.2;
```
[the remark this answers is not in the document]
yunfengzhou-hub: Spark uses `spark.memory.storageFraction` to configure the proportion of memory used for storage, i.e., caching rdd results. I'll temporarily change this magic number's value to 0.5, the default value of `spark.memory.storageFraction`, to align with Spark's practice. Later we may consider make these magic numbers as Flink ML configuration parameters.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
yunfengzhou-hub: This class contains all information about a cache unit, no matter the cache is on disk, in memory, or both. I'll modify the doc and the api to improve its readability.

yunfengzhou-hub: This class contains all information about a cache unit, no matter the cache is on disk, in memory, or both. I'll modify the doc and the api to improve its readability.

zhipeng93 (commented):

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
```diff
@@ -20,120 +20,122 @@
...
-            if (!currentSegmentReader.hasNext()) {
-                currentSegmentReader.close();
-                if (currentSegmentReader.index < segments.size() - 1) {
-                    currentSegmentReader = new SegmentReader(currentSegmentReader.index + 1, 0);
-                } else {
-                    currentSegmentReader = null;
+            T record = currentReader.next();
+
+            if (cacheWriter != null) {
+                if (!cacheWriter.addRecord(record)) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: In cases when the flink job is restored from a snapshot, which means previous caches created by the segment writer in `DataCacheWriter` is non-existent, this cache writer helps to re-create the caches during reading process.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -94,6 +120,26 @@ public static <T> DataStream<T> reduce(DataStream<T> input, ReduceFunction<T> fu
...
+     * @param randomSeed The seed to randomly pick elements as sample.
+     * @return A data stream containing a list of the sampled elements.
+     */
+    public static <T> DataStream<List<T>> sample(
+            DataStream<T> input, int numSamples, long randomSeed) {
+        return input.transform(
+                        "samplingOperator",
+                        Types.LIST(input.getType()),
+                        new SamplingOperator<>(numSamples, randomSeed))
+                .setParallelism(1);
```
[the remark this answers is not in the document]
yunfengzhou-hub: The SampleOperator would return only one element, which is a List<T>, so retaining the parallelism of upstream operator seems to be meaningless. 

I agree that distributed sampling can bring better performance. I'll make the change this way.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,114 @@
...
+        this.memoryManager = memoryManager;
+
+        if (expectedSize > 0) {
+            memoryManager.reserveMemory(this.path, expectedSize);
+        }
+        this.reservedMemorySize = expectedSize;
+    }
+
+    @Override
+    public boolean addRecord(T record) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, I'll adopt the second choice to store records in Flink's managed memory in serialized format.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/FsSegmentWriter.java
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, I'll adopt buffered stream to improve disk io performance.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
yunfengzhou-hub: According to offline discussion, I'll store serialized format of the records in memory, so there is no need for size estimation now.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,49 @@
...
-    }
-
-    public Path getPath() {
-        return path;
+        this.fsSize = fsSize;
     }
 
-    public int getCount() {
-        return count;
+    boolean isOnDisk() throws IOException {
```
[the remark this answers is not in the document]
yunfengzhou-hub: We have the need to express a segment that is both cached in memory and persisted on disk. If we separate the class into MemorySegment and FsSegment, I'm not sure how to deal with such situation.

zhipeng93 (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -182,4 +307,79 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
...
+            implements OneInputStreamOperator<T, List<T>>, BoundedOneInput {
+        private final int numSamples;
+
+        private final Random random;
+
+        private ListState<T> samplesState;
+
+        private List<T> samples;
+
+        private ListState<Integer> countState;
```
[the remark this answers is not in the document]
yunfengzhou-hub: `count` and `countState` represents the number of elements that have been received by the subtask so far. It might be larger than or equal to `samples.size()`.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -182,4 +307,79 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
...
+            } else {
+                if (random.nextInt(count) < numSamples) {
+                    samples.set(random.nextInt(numSamples), sample);
+                }
+            }
+        }
+
+        @Override
+        public void endInput() throws Exception {
+            Collections.shuffle(samples, random);
```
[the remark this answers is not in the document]
yunfengzhou-hub: I think it is necessary. Otherwise the first arriving element, if sampled, will always be the first returning element, while elements that arrived after the first sampled `numSamples` elements will be returned in random order.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,170 @@
...
+
+        private void ensureCapacity(int capacity) throws MemoryAllocationException {
+            Preconditions.checkArgument(capacity > 0);
+            int requiredSegmentNum = (capacity - 1) / pageSize + 2 - segments.size();
+            if (requiredSegmentNum > 0) {
+                segments.addAll(memoryManager.allocatePages(getKey(), requiredSegmentNum));
+            }
+        }
+
+        private void writeRecursive(byte[] b, int off, int len) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Do you mean we should allow `len` to be zero or prohibit it? I think we should allow such usage so as to correspond to lower-level implementation, and the code has already achieved it.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,170 @@
...
+
+        public ManagedMemoryOutputStream(LimitedSizeMemoryManager memoryManager, long expectedSize)
+                throws MemoryAllocationException {
+            this.memoryManager = memoryManager;
+            this.pageSize = memoryManager.getPageSize();
+            this.segmentIndex = 0;
+            this.segmentOffset = 0;
+
+            Preconditions.checkArgument(expectedSize >= 0);
+            if (expectedSize > 0) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Do you mean we should allocate all segments from the beginning, or we should not do it but the current implementation has forced such practice?

In my opinion, we should not do it so as to improve memory utilization. And the current implementation has achieved it, since it will require memory as more records are added instead of occupying all memory space during initialization. `expectedSize` is only used when recreating cache during reading process, in which case we would have known the size to be occupied from the start.

According to other offline discussions I'll modify the code to use flink's `LazyMemorySegmentPool` instead of `LimitedSizeMemoryManager`, and does the re-caching process when recovering from snapshot instead of during reading. But the statements above remain valid.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,104 @@
...
+        this.memoryManager = memoryManager;
+        this.finishedSegments = new ArrayList<>(priorFinishedSegments);
+        this.currentWriter =
+                SegmentWriter.create(
+                        pathGenerator.get(), this.memoryManager, serializer, 0L, true, true);
     }
 
     public void addRecord(T record) throws IOException {
-        currentSegment.addRecord(record);
+        boolean success = currentWriter.addRecord(record);
```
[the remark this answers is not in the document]
yunfengzhou-hub: I believe such case is possible. For example, we have written size of a vector to the last bytes of a memory segment, and failed to continue writing the values of the vector because no more segments are available for this operator in memory manager. In this case the code would re-create an `FsSegmentWriter` and re-write the size of the vector to the file. 

Such case exists, but it will do no harm to the program, because the idling `size` value in the segment is not tracked so would not be accessed. It does not waste memory space either since we are allocating space at segments' granularity. It will be released afterwards along with those valid values in the segment.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheSnapshot.java
```diff
@@ -90,18 +90,18 @@ public void writeTo(OutputStream checkpointOutputStream) throws IOException {
...
                 // We have to copy the whole streams.
-                int totalRecords = segments.stream().mapToInt(Segment::getCount).sum();
-                long totalSize = segments.stream().mapToLong(Segment::getSize).sum();
-                checkState(totalRecords >= 0, "overflowed: " + totalRecords);
-                dos.writeInt(totalRecords);
-                dos.writeLong(totalSize);
-
+                dos.writeInt(segments.size());
                 for (Segment segment : segments) {
+                    dos.writeInt(segment.getCount());
```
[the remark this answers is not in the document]
yunfengzhou-hub: Because the max size of a segment is limited. For example, limited by the max allowed file size of the underlying filesystem. If we merge all segments into one during snapshot, errors due to such limits might be invoked. Treating each segment separately could help avoid such errors while not adding much overhead to the snapshot process.

yunfengzhou-hub:

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,61 +18,37 @@
...
+/** A segment contains the information about a cache unit. */
+@Internal
+class Segment {
 
-/** A segment represents a single file for the cache. */
-public class Segment implements Serializable {
+    private FileSegment fileSegment;
 
-    private final Path path;
+    private MemorySegment memorySegment;
```
lindong28, line 29: Instead of creating a new class `MemorySegment` whose name collide with a class in `org.apache.flink.core.memory.MemorySegment`, would it be simpler to just put `List<org.apache.flink.core.memory.MemorySegment>` inside `Segment`?

The `count` in `MemorySegment` and `FileSegment` can also be moved to `Segment` if we need it.

lindong28 (commented):

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCache.java
```diff
@@ -0,0 +1,351 @@
...
+        } catch (SegmentNoVacancyException e) {
+            currentWriter.finish().ifPresent(finishedSegments::add);
+            currentWriter = new FileSegmentWriter<>(serializer, pathGenerator.get());
+            currentWriter.addRecord(record);
+        }
+    }
+
+    /** Finishes adding records and closes resources occupied for adding records. */
+    public void finish() throws IOException {
+        if (currentWriter == null) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: Like described in the comment below, we might need to call `finish()` before reading from the data cache, which happens every time before we create a iterator. Thus finish() might be invoked multiple times. If we remove this statement, we would need to do like follows.
```java
    @Override
    public DataCacheIterator<T> iterator() {
       if (!finished){
            finish();
            finished=true;
      }
        return new DataCacheIterator<>(serializer, finishedSegments);
    }
```
Thus we would still need the `if` condition, and it might be better to directly place it inside `finish()`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCache.java
```diff
@@ -0,0 +1,351 @@
...
+                fileSystem.delete(segment.getFileSegment().getPath(), false);
+            }
+            if (segment.getMemorySegment() != null) {
+                segmentPool.returnAll(segment.getMemorySegment().getCache());
+            }
+        }
+        finishedSegments.clear();
+    }
+
+    private void finishCurrentSegmentIfAny() throws IOException {
```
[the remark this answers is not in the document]
yunfengzhou-hub: There is no such usage for now. I agree that we could put `finish()` an internal method and call it before reading.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheIterator.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+
+/** Reads the cached data from a list of segments. */
+@Internal
+public class DataCacheIterator<T> implements Iterator<T> {
```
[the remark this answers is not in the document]
yunfengzhou-hub: The original API is coupled together, so whenever the users use `DataCacheWriter`, they would also use `DataCacheSnapshot` and `DataCacheReader`. In this case I think it is better to merge them into one class.

Another advantage of the current API is that it better hides implementation details from developers.  If we still use the original API. we would have to preserve the `Segment` class in order to pass information between `DataCacheWriter`, `DataCacheSnapshot` and `DataCacheReader`, and developers need to learn the notion of `Segment` in order to use the data cache mechanism. While in previous offline discussion, we agreed that there might not be a unified `Segment` concept and might keep `FileSegment` and `MemorySegment` separately in internal implementation. We might also find that there is a better internal representation of the cache unit than `xxxSegment` in future, but it would be hard for us to make such improvement then because `Segment` has become a public API.

Since the `DataCache` holds most functions of previous `DataCacheWriter` and `DataCacheSnapshot`, naming it as `DataCacheWriter` might not be suitable now.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,61 +18,37 @@
...
-        Segment segment = (Segment) o;
-        return count == segment.count && size == segment.size && Objects.equals(path, segment.path);
+    FileSegment getFileSegment() {
+        return fileSegment;
     }
 
-    @Override
-    public int hashCode() {
-        return Objects.hash(path, count, size);
+    void setMemorySegment(MemorySegment memorySegment) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: It is used in `DataCache.tryCacheSegmentToMemory`. I just found that there is a bug in this method that it should call `segment.setMemorySegment(x.getMemorySegment())`. I'll make the modification.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,61 +18,37 @@
...
 
-import java.io.Serializable;
-import java.util.Objects;
+/** A segment contains the information about a cache unit. */
+@Internal
+class Segment {
 
-/** A segment represents a single file for the cache. */
-public class Segment implements Serializable {
+    private FileSegment fileSegment;
```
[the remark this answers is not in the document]
yunfengzhou-hub: We need to create the `Segment` instance before we persist the segment to disk and acquire enough information to create `FileSegment`. This means that the `fileSegment` field need to be set after the creation of `Segment`, and it cannot be final.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, I agree with it that we should not change parallelism to make it more generic. I'll make the change.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCache.java
```diff
@@ -0,0 +1,351 @@
...
+        if (currentWriter == null || currentWriter.getCount() == 0) {
```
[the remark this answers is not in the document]
yunfengzhou-hub: It is possible. For example in `ReplayOperator`, `finish()` is invoked at `onEpochWatermarkIncrement` while `finishCurrentSegmentIfAny()` is invoked at `snapshotState`. Since `snapshotState` can happen before or after `onEpochWatermarkIncrement`, `finish()` and `finishCurrentSegmentIfAny()` might be invoked repeatedly in any order, and the `if` condition if unavoidable. If we remove the `if` from DataCacheWriter, users still need to add it in their code.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/FileSegmentWriter.java
```diff
@@ -0,0 +1,92 @@
...
+        this.fileSystem = path.getFileSystem();
+        this.outputStream = fileSystem.create(path, FileSystem.WriteMode.NO_OVERWRITE);
+        this.bufferedOutputStream = new BufferedOutputStream(outputStream);
+        this.outputView = new DataOutputViewStreamWrapper(bufferedOutputStream);
+    }
+
+    @Override
+    public void addRecord(T record) throws IOException {
+        if (outputStream.getPos() >= FILE_SIZE_LIMIT) {
+            throw new SegmentNoVacancyException();
```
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, we still need to use Exception to pass relative information out from output stream, but in the wrapper method we should cache the exception and convert exceptions related to limited size to proper message, like a boolean denoting success or failure. I'll make the change this way.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheIterator.java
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, I'll retrieve the original `DataCacheWriter`, `DataCacheSnapshot` and `DataCacheReader`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, I'll make `path` a `final` variable in `Segment` as it always need to be allocated.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,158 @@
...
+            currentWriter = new FileSegmentWriter<>(serializer, pathGenerator.get());
+            Preconditions.checkState(currentWriter.addRecord(record));
+        }
     }
 
-    public FileSystem getFileSystem() {
-        return fileSystem;
-    }
+    /** Finishes current segment if records has ever been added to this segment. */
+    public void finishCurrentSegmentIfAny() throws IOException {
```
lindong28, line 112: It appears that this method is called when we increment epoch watermark, end input, snapshot state, or inside `processPendingElementsAndWatermarks()`.

The first four cases happens at most regularly. The segment will most likely contain data. And in the rare cases when it does not contain data, the performance impact of re-creating an empty segment should be negligible given the frequency of these cases. 

Due to the use of `hasPendingElements` inside `AbstractBroadcastWrapperOperator`, it appears that `processPendingElementsAndWatermarks()` is called only once for each `AbstractBroadcastWrapperOperator()` instance, which suggests that the overhead to re-create empty segment due to this method will also be pretty small.

How about we remove the optimization related to `getCount()` to simplify the code, and only add it if we have concrete reason to believe it will improve performance?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,158 @@
...
+        if (currentWriter == null || currentWriter.getCount() == 0) {
```
lindong28, line 113: It looks like we can make this method private and let `persistSegmentsToDisk/getFinishedSegments` call this method.

And we can rename `getFinishedSegments()` as `getSegments()`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,80 @@
...
+    /** The path to the file containing cached records. */
     private final Path path;
 
-    /** The count of the records in the file. */
+    /** The count of the records in the segment. */
     private final int count;
 
-    /** The total length of file. */
-    private final long size;
+    /** The total length of file containing cached records. */
```
lindong28, line 42: nits: the records in the file is not `cached`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,80 @@
...
 
+/** A segment contains the information about a cache unit. */
+@Internal
+public class Segment {
+
+    /** The path to the file containing cached records. */
     private final Path path;
 
-    /** The count of the records in the file. */
+    /** The count of the records in the segment. */
```
lindong28, line 39: By `in the segment`, do you mean `in the file` or `in the memory`? Could you explain this in the Java doc?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,158 @@
...
 
-        public SegmentWriter(Path path) throws IOException {
-            this.path = path;
-            this.outputStream = fileSystem.create(path, FileSystem.WriteMode.NO_OVERWRITE);
-            this.outputView = new DataOutputViewStreamWrapper(outputStream);
+    /** Cleans up all previously added records. */
+    public void cleanup() throws IOException {
+        finishCurrentSegmentIfAny();
+        for (Segment segment : finishedSegments) {
+            if (segment.isOnDisk()) {
```
lindong28, line 140: Suppose `fsSize == 0`, should we still invoke `fileSystem.delete(...)` to delete the empty file?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,80 @@
...
 
-    public int getCount() {
+    int getCount() {
         return count;
     }
 
-    public long getSize() {
-        return size;
+    long getFsSize() {
+        checkState(fsSize > 0);
```
lindong28, line 89: Suppose the input datastream does not have any data, we should still be able to snapshot/persist this data stream onto disk, and be able to reload this datastream after restarting the Flink job, right?

Then we will need to support a segment with `fsSize = 0`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,80 @@
...
+        this.cache = checkNotNull(cache);
     }
 
-    public Path getPath() {
+    void setDiskInfo(long fsSize) {
+        checkArgument(fsSize > 0);
+        this.fsSize = fsSize;
+    }
+
+    boolean isOnDisk() {
```
lindong28, line 72: Suppose `getFsSize()` can be used with `fsSize == 0`, would it be simpler to remove this method and let caller use `getFsSize() > 0`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,158 @@
...
+                fileSystem.delete(segment.getPath(), false);
+            }
+            if (segment.isCached()) {
+                assert segmentPool != null;
```
lindong28, line 144: After thinking about this more, I find it more consistent with other code to just remove the assert here. Note that the code will throw NullPointerException if segmentPool == null, which is good enough for us to investigate this bug.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,80 @@
...
 
-    public long getSize() {
-        return size;
+    long getFsSize() {
+        checkState(fsSize > 0);
+        return fsSize;
+    }
+
+    List<MemorySegment> getCache() {
+        return checkNotNull(cache);
```
lindong28, line 94: For `getXXX()` method, it is in general simpler and more intuitive to just return the fact (e.g. null or empty list) instead of throwing exception.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,191 @@
...
+                    segmentPool.returnAll(allocatedSegments);
+                    throw new MemoryAllocationException();
+                }
+                allocatedSegments.add(memorySegment);
+            }
+
+            segments.addAll(allocatedSegments);
+        }
+
+        private void writeRecursive(byte[] b, int off, int len) {
```
lindong28, line 176: Could we change this function to be iterative to improve performance?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,191 @@
...
+            Preconditions.checkArgument(capacity > 0);
+            int required =
+                    (int) (capacity % pageSize == 0 ? capacity / pageSize : capacity / pageSize + 1)
+                            - segments.size();
+
+            List<MemorySegment> allocatedSegments = new ArrayList<>();
+            for (int i = 0; i < required; i++) {
+                MemorySegment memorySegment = segmentPool.nextSegment();
+                if (memorySegment == null) {
+                    segmentPool.returnAll(allocatedSegments);
```
lindong28, line 167: Should we also return `segments` to the segmentPool? This could be useful when `write(...)` encounters limit.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,191 @@
...
+
+        @Override
+        public void write(int b) throws IOException {
+            write(new byte[] {(byte) b}, 0, 1);
+        }
+
+        @Override
+        public void write(@Nullable byte[] b, int off, int len) throws IOException {
+            try {
+                ensureCapacity(globalOffset + len);
```
lindong28, line 150: Instead of repeatedly allocate `List<MemorySegment>` instance and do divide operation for each write(...), would it be simpler to just maintain `allocatedBytes` and just check `globalOffset + len <= allocatedBytes` for must write() operation?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/FileSegmentWriter.java
```diff
@@ -0,0 +1,89 @@
...
+            return false;
+        }
+        serializer.serialize(record, outputView);
+        count++;
+        return true;
+    }
+
+    @Override
+    public Optional<Segment> finish() throws IOException {
+        this.bufferedOutputStream.flush();
```
lindong28, line 75: nits: we use `this` when the member variable name collide with the input parameter name. This typically happen in the constructor.  It looks like we don't need to use `this` here.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
```diff
@@ -20,120 +20,91 @@
...
         this.serializer = serializer;
-        this.fileSystem = fileSystem;
         this.segments = segments;
+        this.segmentIndex = readerPosition.f0;
+        this.segmentCount = readerPosition.f1;
 
-        if (readerPosition.f0 < segments.size()) {
-            this.currentSegmentReader = new SegmentReader(readerPosition.f0, readerPosition.f1);
-        }
+        createSegmentReader(readerPosition.f0, readerPosition.f1);
```
lindong28, line 61: The `readerPosition.f1` refers to the number of records that have been read from the `DataCache`, right?

But `createSegmentReader(...)` passes this position directly to the `MemorySegmentReader` constructor, which interprets this value as the number of records that should be skipped `within this segment`.

Should this inconsistency be fixed?

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/typeinfo/DenseVectorSerializer.java
```diff
@@ -20,29 +20,36 @@
...
 /** Specialized serializer for {@link DenseVector}. */
-public final class DenseVectorSerializer extends TypeSerializerSingleton<DenseVector> {
+public final class DenseVectorSerializer extends TypeSerializer<DenseVector> {
 
     private static final long serialVersionUID = 1L;
 
     private static final double[] EMPTY = new double[0];
 
-    public static final DenseVectorSerializer INSTANCE = new DenseVectorSerializer();
+    private final byte[] buf = new byte[1024];
```
lindong28, line 41: Instead of add buffer inside serializers and making those serializers non-singleton, would it be simpler to just let caller use `BufferedOutputStream`?

For example, `NaiveBayesModelData::ModelDataEncoder::encode()` can do this:

```
DataOutputViewStreamWrapper outputViewStreamWrapper =
        new DataOutputViewStreamWrapper(new BufferedOutputStream(outputStream));
```

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
```diff
@@ -20,120 +20,91 @@
...
     private final TypeSerializer<T> serializer;
 
-    private final FileSystem fileSystem;
-
+    /** The segments where to read the records from. */
     private final List<Segment> segments;
 
-    @Nullable private SegmentReader currentSegmentReader;
+    /** The current reader for next records. */
+    @Nullable private SegmentReader<T> currentReader;
```
lindong28, line 40: nits: since the class name is `DataCacheReader`, could we rename this parameter as `currentSegmentReader` to minimize confusion?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
```diff
@@ -20,120 +20,91 @@
...
 
-    public DataCacheReader(
-            TypeSerializer<T> serializer, FileSystem fileSystem, List<Segment> segments)
-            throws IOException {
-        this(serializer, fileSystem, segments, new Tuple2<>(0, 0));
+    /** The index of the segment that current reader reads from. */
+    private int segmentIndex;
```
lindong28, line 43: nits: how about renaming it as `currentSegmentIndex` to make the name more consistent with other variables?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheReader.java
```diff
@@ -20,120 +20,91 @@
...
+
+    /** The number of records that have been read through current reader so far. */
+    private int segmentCount;
```
lindong28, line 46: nits: to avoid confusing this count with the number of records that have been read by this data cache, how about renaming it as `curentSegmentCount`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,162 @@
...
-    private final List<Segment> finishSegments;
+    /** An optional pool that provide memory segments to hold cached records in memory. */
+    @Nullable private final MemorySegmentPool segmentPool;
+
+    /** The segments that contain previously added records. */
+    private final List<Segment> finishedSegments;
 
-    private SegmentWriter currentSegment;
+    /** The current writer for new records. */
+    @Nullable private SegmentWriter<T> currentWriter;
```
lindong28, line 58: nits: could we rename this variable as `currentSegmentWriter`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheWriter.java
```diff
@@ -19,127 +19,162 @@
...
         }
+        finishedSegments.clear();
+    }
 
-        public void addRecord(T record) throws IOException {
-            serializer.serialize(record, outputView);
-            currentSegmentCount += 1;
-        }
+    /** Persists the segments in this writer to disk. */
+    public void persistSegmentsToDisk() throws IOException {
```
lindong28, line 154: nits: given that we use `FileSegmentWriter`, would it be more consistent to rename this method as `writeSegmentsToFiles()`?

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -113,32 +162,119 @@ public static <T> DataStream<T> reduce(DataStream<T> input, ReduceFunction<T> fu
...
 
         @Override
         public void processElement(StreamRecord<IN> input) throws Exception {
-            valuesState.add(input.getValue());
+            dataCacheWriter.addRecord(input.getValue());
+        }
+
+        @Override
+        public void endInput() throws Exception {
+            List<Segment> pendingSegments = dataCacheWriter.getSegments();
```
lindong28, line 251: nits: It is not clear what `pending` means in this context. Would it be simpler to just name it `segments`?

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -113,32 +162,119 @@ public static <T> DataStream<T> reduce(DataStream<T> input, ReduceFunction<T> fu
...
             implements OneInputStreamOperator<IN, OUT>, BoundedOneInput {
 
-        private ListState<IN> valuesState;
+        private final TypeInformation<IN> inputType;
+
+        private Path basePath;
+
+        private StreamConfig config;
+
+        private StreamTask<?, ?> containingTask;
```
lindong28, line 171: It appears that we only need `containingTask.getExecutionConfig()`. Would it be simpler to just save the ExecutionConfig?

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -256,4 +392,81 @@ public void flatMap(T[] values, Collector<Tuple2<Integer, T[]>> collector) {
...
+        @Override
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            samplesState.update(samples);
+            countState.update(Collections.singletonList(count));
+        }
+
+        @Override
+        public void processElement(StreamRecord<T> streamRecord) throws Exception {
+            T sample = streamRecord.getValue();
```
lindong28, line 451: nits: sample means something that is chosen from a collection. Since it is not sure we will chose this value, would it be simpler to just name it `value`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentReader.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+        return count < totalCount;
+    }
+
+    @Override
+    public T next() throws IOException {
+        T ret = serializer.deserialize(inputView);
```
lindong28, line 67: nits: could we replace `ret` with `value` for consistency with `FileSegmentReader::next()`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentReader.java
```diff
@@ -0,0 +1,126 @@
...
+                segmentOffset = 0;
+            }
+            return ret;
+        }
+
+        @Override
+        public int read(byte[] b, int off, int len) throws IOException {
+            int readLen = 0;
+
+            while (len > 0 && segmentIndex < segments.size()) {
```
lindong28, line 108: nits: the following code is probably simpler since it does not need to explicitly write `break`.

```
while (len > 0) {
    int currentLen = Math.min(segments.get(segmentIndex).size() - segmentOffset, len);
    segments.get(segmentIndex).get(segmentOffset, b, off, currentLen);
    segmentOffset += currentLen;
    if (segmentOffset >= segments.get(segmentIndex).size()) {
        segmentIndex++;
        segmentOffset = 0;
    }

    readLen += currentLen;
    off += currentLen;
    len -= currentLen;
}
```

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,194 @@
...
+
+        /** The number of bytes that have been allocated so far. */
+        private long allocatedBytes;
+
+        public ManagedMemoryOutputStream(MemorySegmentPool segmentPool, long expectedSize)
+                throws MemoryAllocationException {
+            this.segmentPool = segmentPool;
+            this.pageSize = segmentPool.pageSize();
+
+            Preconditions.checkArgument(expectedSize >= 0);
```
lindong28, line 131: nits: It seems unnecessary to check this and use Math.max(..) below. Since it does not cause any harm to have expectedSize < 0, would it be simpler to remove this line?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,194 @@
...
+        public void write(int b) throws IOException {
+            write(new byte[] {(byte) b}, 0, 1);
+        }
+
+        @Override
+        public void write(@Nullable byte[] b, int off, int len) throws IOException {
+            try {
+                ensureCapacity(globalOffset + len);
+            } catch (MemoryAllocationException e) {
+                throw new IOException(e);
```
lindong28, line 153: Would it be simpler to make `MemoryAllocationException` subclass of RuntimeException and throw/catch `MemoryAllocationException` directly, instead of wrapping it inside `IOException`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/MemorySegmentWriter.java
```diff
@@ -0,0 +1,194 @@
...
+            }
+
+            while (len > 0) {
```
lindong28, line 156: It seems simpler to use the following code, which does not need to explicitly call `break`.

```
while (len > 0) {
    int currentLen = Math.min(len, pageSize - segmentOffset);
    segments.get(segmentIndex).put(segmentOffset, b, off, currentLen);
    segmentOffset += currentLen;
    globalOffset += currentLen;
    if (segmentOffset >= pageSize) {
        segmentIndex++;
        segmentOffset = 0;
    }
    off += currentLen;
    len -= currentLen;
}
```

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,73 @@
...
+    /**
+     * The count of records in the file at the path if the file size is not zero, otherwise the
+     * count of records in the cache.
+     */
     private final int count;
 
-    /** The total length of file. */
-    private final long size;
+    /** The total length of file containing persisted records. */
+    private long fsSize = -1L;
```
lindong28, line 45: Would it be more intuitive to initialize the fsSize to 0?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,73 @@
...
-    /** The count of the records in the file. */
+    /**
+     * The count of records in the file at the path if the file size is not zero, otherwise the
+     * count of records in the cache.
+     */
     private final int count;
 
-    /** The total length of file. */
-    private final long size;
+    /** The total length of file containing persisted records. */
```
lindong28, line 44: Could you update the the comment explaining that `fsSize is 0 iff the segment has not been written to the given path`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,73 @@
...
+    private long fsSize = -1L;
+
+    /** The memory segments containing cached records. */
```
lindong28, line 47: Could you update the code explaining that `the cache list is empty iff the segment has not been cached in memory`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -18,38 +18,73 @@
...
     }
 
-    public long getSize() {
-        return size;
+    long getFsSize() {
+        return fsSize;
+    }
+
+    List<MemorySegment> getCache() {
+        return cache;
```
lindong28, line 87: Instead of indicating `the segment has been read into the memory` by using `cache == 0`, would it be simpler to use `cache.isEmpty()`?

This would allow us to handle less special case (e.g. null). After all if the segment has been read into memory, it must have non-empty memory segment list.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
```diff
@@ -63,16 +98,18 @@ public boolean equals(Object o) {
         }
 
         Segment segment = (Segment) o;
-        return count == segment.count && size == segment.size && Objects.equals(path, segment.path);
+        return count == segment.count
+                && fsSize == segment.fsSize
```
lindong28, line 102: Should a segment be uniquely identified by its count and path only?

I suppose if we write segment data from memory to file, it should still represent the same segment, right?

Same for `hashCode()`.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/FileSegmentWriter.java
```diff
@@ -0,0 +1,89 @@
...
+            return false;
+        }
+        serializer.serialize(record, outputView);
+        count++;
+        return true;
+    }
+
+    @Override
+    public Optional<Segment> finish() throws IOException {
+        bufferedOutputStream.flush();
```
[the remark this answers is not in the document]
yunfengzhou-hub: `bufferedOutputStream` wraps around `outputStream` and is invisible to `outputStream`. When `outputStream.flush()` is invoked, there might be some bytes left in `bufferedOutputStream`. When I tried to remove this line, most data cache tests failed. Thus this line cannot be removed.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/FileSegmentWriter.java
```diff
@@ -0,0 +1,89 @@
...
+    private final TypeSerializer<T> serializer;
+
+    /** The path to the target file. */
+    private final Path path;
+
+    /** The output stream that writes to the target file. */
+    private final FSDataOutputStream outputStream;
+
+    /** A buffer that wraps the output stream to optimize performance. */
+    private final BufferedOutputStream bufferedOutputStream;
```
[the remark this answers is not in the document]
yunfengzhou-hub: Since we cannot remove the `bufferedOutputStream.flush()` in `finish()` according to the comment above, we cannot remove this variable.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/Segment.java
yunfengzhou-hub: Now the actual behavior is that the cache list is null rather than empty when not cached. I think it better satisfies Java's default values. How do you think we should treat null and empty list accordingly?

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
```diff
@@ -254,58 +272,150 @@ public Tuple3<Integer, DenseVector, Long> map(Tuple2<Integer, DenseVector> value
                             DenseVector, DenseVector[], Tuple2<Integer, DenseVector>>,
                     IterationListener<Tuple2<Integer, DenseVector>> {
         private final DistanceMeasure distanceMeasure;
-        private ListState<DenseVector> points;
-        private ListState<DenseVector[]> centroids;
+        private ListState<DenseVector[]> centroidsState;
+        private DenseVector[] centroids;
```
lindong28, line 276: Does this improve performance by using `centroids`? If not, it seems simpler to not adding this variable.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
```diff
@@ -254,58 +272,150 @@ public Tuple3<Integer, DenseVector, Long> map(Tuple2<Integer, DenseVector> value
...
+        @Override
+        public void setup(
+                StreamTask<?, ?> containingTask,
+                StreamConfig config,
+                Output<StreamRecord<Tuple2<Integer, DenseVector>>> output) {
+            super.setup(containingTask, config, output);
+
+            operatorID = config.getOperatorID();
+
+            MemoryManager memoryManager = getContainingTask().getEnvironment().getMemoryManager();
```
lindong28, line 297: Could you refactor the code so that algorithm developers won't need to handle so many details?

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/typeinfo/DenseVectorSerializer.java
yunfengzhou-hub: According to offline discussion, BufferedOutputStream cannot avoid it that there will be a lot of small byte array read/write, and still have performance issues. We'll still use the internal-byte-buffer practice.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -113,32 +151,44 @@ public static <T> DataStream<T> reduce(DataStream<T> input, ReduceFunction<T> fu
...
-        private ListState<IN> valuesState;
+        private final TypeInformation<IN> inputType;
 
-        public MapPartitionOperator(MapPartitionFunction<IN, OUT> mapPartitionFunc) {
+        private DataCacheListState<IN> dataCacheListState;
+
+        public MapPartitionOperator(
+                MapPartitionFunction<IN, OUT> mapPartitionFunc, TypeInformation<IN> inputType) {
             super(mapPartitionFunc);
+            this.inputType = inputType;
```
lindong28, line 161: Instead of passing inputType to create serializer, would it be simpler to keep the previous approach and create serializer using the following code?

```
getOperatorConfig().getTypeSerializerIn(0, getClass().getClassLoader())
```

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
```diff
@@ -253,59 +258,76 @@ public Tuple3<Integer, DenseVector, Long> map(Tuple2<Integer, DenseVector> value
             implements TwoInputStreamOperator<
                             DenseVector, DenseVector[], Tuple2<Integer, DenseVector>>,
                     IterationListener<Tuple2<Integer, DenseVector>> {
+
         private final DistanceMeasure distanceMeasure;
-        private ListState<DenseVector> points;
-        private ListState<DenseVector[]> centroids;
+
+        private ListState<DenseVector[]> centroidsState;
```
lindong28, line 264: nits: could we keep the original name `centroids` for simplicity? Same for points.

lindong28 (commented):
Thanks for the update! The PR looks pretty good now. Left some comments below.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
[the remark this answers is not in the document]
zhipeng93: > the first arriving element, if sampled, will always be the first returning element

What is the problem of situation? I think it is ok if the sampled element preserves the order in each worker. If we look at spark#sample [1], the order in each partition also preserves.

[1] https://github.com/apache/spark/blob/6026dd25748fd79caeedc083f99d5c954fb3a19f/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L554

flink-ml-iteration/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/DataCacheSnapshot.java
```diff
@@ -167,26 +183,69 @@ public static DataCacheSnapshot recover(
...
-                    inputStream.setPropagateClose(false);
-                    IOUtils.copyBytes(inputStream, outputStream, false);
-                    inputStream.close();
+                int segmentNum = dis.readInt();
+                segments = new ArrayList<>(segmentNum);
+                for (int i = 0; i < segmentNum; i++) {
+                    int count = dis.readInt();
+                    long fsSize = dis.readLong();
+                    Path path = pathGenerator.get();
+                    try (FSDataOutputStream outputStream =
```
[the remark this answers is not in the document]
yunfengzhou-hub: According to our offline discussion, the original purpose of merging segments into one is to reduce the number of files created. He also agrees that it might cause the size of the file merged from several other files to exceed the max allowed size of the file system, so our current practice to preserve the original segments is acceptable to him. The current use cases of DataCache segments would not generate lots of small segments, so the performance issue is not obvious for now.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
```diff
@@ -160,6 +162,9 @@ public IterationBodyResult process(
                                             BasicTypeInfo.INT_TYPE_INFO,
                                             DenseVectorTypeInfo.INSTANCE),
                                     new SelectNearestCentroidOperator(distanceMeasure));
+            centroidIdAndPoints
+                    .getTransformation()
+                    .declareManagedMemoryUseCaseAtOperatorScope(ManagedMemoryUseCase.OPERATOR, 64);
```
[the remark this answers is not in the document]
yunfengzhou-hub: The methods provided in the reference link allow us to specify the proportion of managed memory shared among all usages of `ManagedMemoryUseCase.OPERATOR`, while the `declareManagedMemoryUseCaseAtOperatorScope` method here further specifies the proportion of managed memory this operator can use among all operators that have stated that they need managed memory through invoking this method. These two are configurations at different scopes.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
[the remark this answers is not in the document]
yunfengzhou-hub: Got it. As we does not provide any guarantee to the order to the returned samples, it is also acceptable not to shuffle the elements. I'll remove this line.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, we would create a utility method similar to `ExecNodeUtil.setManagedMemoryWeight` in order to control the usage of `declareManagedMemoryUseCaseAtOperatorScope` in Flink ML. The weight values should reference existing values in Flink, like `StreamExecWindowAggregate.WINDOW_AGG_MEMORY_RATIO`.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -256,4 +325,79 @@ public void flatMap(T[] values, Collector<Tuple2<Integer, T[]>> collector) {
...
+
+        @Override
+        public void processElement(StreamRecord<T> streamRecord) throws Exception {
+            T value = streamRecord.getValue();
+            count++;
+
+            if (samples.size() < numSamples) {
+                samples.add(value);
+            } else {
+                if (random.nextInt(count) < numSamples) {
```
lindong28, line 390: nits: could we re-use `random.nextInt(count)` for simplicity?

lindong28 (commented):
Thanks for the update! LGTM. Left just one minor comment.

PR #100, opened 2022-05-17
[FLINK-27096] Add LabeledPointWithWeightGenerator and Benchmark Configuration for KMeans and NaiveBayes

yunfengzhou-hub:
This PR adds a `LabeledPointWithWeightGenerator` to `flink-ml-benchmark` module.

`LabeledPointWithWeightGenerator` is used to generate vector-typed features with double-typed label and weight. The generated tables can be used as the train or predict input data of some supervised algorithms. For example,
- Naive Bayes
- Logistic Regression
- Linear Regression
- Linear SVC
- Knn

### DIFF
--- a/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/InputDataGenerator.java
+++ b/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/InputDataGenerator.java
@@ -21,7 +21,7 @@ package org.apache.flink.ml.benchmark.datagenerator;
 import org.apache.flink.ml.param.LongParam;
 import org.apache.flink.ml.param.Param;
 import org.apache.flink.ml.param.ParamValidators;
-import org.apache.flink.ml.param.StringArrayParam;
+import org.apache.flink.ml.param.StringArrayArrayParam;
 
 /** Interface for generating data as input table arrays. */
 public interface InputDataGenerator<T extends InputDataGenerator<T>> extends DataGenerator<T> {
@@ -29,10 +29,10 @@ public interface InputDataGenerator<T extends InputDataGenerator<T>> extends Dat
             new LongParam(
                     "numValues", "Number of data to be generated.", 10L, ParamValidators.gt(0));
 
-    Param<String[]> COL_NAMES =
-            new StringArrayParam(
+    Param<String[][]> COL_NAMES =
+            new StringArrayArrayParam(
                     "colNames",
-                    "An array of common-separated strings representing field names of data tables.",
+                    "A 2D array of strings. Each string array element represents the column names of a data table.",
                     null);
 
     default long getNumValues() {
@@ -43,11 +43,11 @@ public interface InputDataGenerator<T extends InputDataGenerator<T>> extends Dat
         return set(NUM_VALUES, value);
     }
 
-    default String[] getColNames() {
+    default String[][] getColNames() {
         return get(COL_NAMES);
     }
 
-    default T setColNames(String... value) {
+    default T setColNames(String[]... value) {
         return set(COL_NAMES, value);
     }
 }
--- a/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
+++ b/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
@@ -1,0 +1,180 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.benchmark.datagenerator.common;
+
+import org.apache.flink.api.common.functions.RichMapFunction;
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeinfo.Types;
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.ml.benchmark.datagenerator.InputDataGenerator;
+import org.apache.flink.ml.benchmark.datagenerator.param.HasVectorDim;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.ml.linalg.typeinfo.DenseVectorTypeInfo;
+import org.apache.flink.ml.param.IntParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.NumberSequenceIterator;
+import org.apache.flink.util.Preconditions;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+
+/**
+ * A DataGenerator which creates a table of features, label and weight.
+ *
+ * <p>Users need to specify three column names as {@link #COL_NAMES}'s value in the following order:
+ *
+ * <ul>
+ *   <li>features column name
+ *   <li>label column name
+ *   <li>weight column name
+ * </ul>
+ */
+public class LabeledPointWithWeightGenerator
+        implements InputDataGenerator<LabeledPointWithWeightGenerator>,
+                HasVectorDim<LabeledPointWithWeightGenerator> {
+
+    public static final Param<Integer> FEATURE_ARITY =
+            new IntParam(
+                    "featureArity",
+                    "Arity of each feature. "
+                            + "If set to positive value, each feature would be an integer in range [0, arity - 1]. "
+                            + "If set to zero, each feature would be a continuous double in range [0, 1).",
+                    2,
+                    ParamValidators.gtEq(0));
+
+    public static final Param<Integer> LABEL_ARITY =
+            new IntParam(
+                    "labelArity",
+                    "Arity of label. "
+                            + "If set to positive value, the label would be an integer in range [0, arity - 1]. "
+                            + "If set to zero, the label would be a continuous double in range [0, 1).",
+                    2,
+                    ParamValidators.gtEq(0));
+
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public LabeledPointWithWeightGenerator() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    public int getFeatureArity() {
+        return get(FEATURE_ARITY);
+    }
+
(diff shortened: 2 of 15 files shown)


flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+
+    public int getLabelArity() {
+        return get(LABEL_ARITY);
+    }
+
+    public LabeledPointWithWeightGenerator setLabelArity(int value) {
+        return set(LABEL_ARITY, value);
+    }
+
+    public static final Param<Integer> FEATURES_ARITY =
```
lindong28, line 77: nits: can we order feature/label/weight consistently across the code? For example, RandomLabeledPointWithWeightGenerator(...) has these fields in the order of feature, label, and weight.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+            int index = getRuntimeContext().getIndexOfThisSubtask();
+            random = new Random(Tuple2.of(initSeed, index).hashCode());
+        }
+
+        @Override
+        public Row map(Long aLong) {
+            double[] values = new double[vectorDim];
+            if (featuresArity > 0) {
+                for (int i = 0; i < vectorDim; i++) {
+                    values[i] = random.nextInt(featuresArity);
```
lindong28, line 177: It is probably simpler to add the following method instead of repeating this code for values, label, and weight.

```
private double getValue(int arity, Random random) {
    if (arity > 0) {
        return random.nextInt(arity);
    }
    return random.nextDouble();
}
```

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+        @Override
+        public void open(Configuration parameters) throws Exception {
+            super.open(parameters);
+            int index = getRuntimeContext().getIndexOfThisSubtask();
+            random = new Random(Tuple2.of(initSeed, index).hashCode());
+        }
+
+        @Override
+        public Row map(Long aLong) {
+            double[] values = new double[vectorDim];
```
lindong28, line 174: Would it be better to name this variable as features, to be consistent with `featuresArity`?

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+        return get(LABEL_ARITY);
+    }
+
+    public LabeledPointWithWeightGenerator setLabelArity(int value) {
+        return set(LABEL_ARITY, value);
+    }
+
+    public static final Param<Integer> FEATURES_ARITY =
+            new IntParam(
+                    "featuresArity",
```
lindong28, line 79: `features` is a dense vector. `featuresArity` might be confused with the rank of this dense vector.

How about renaming this parameter as `featureArity`, since it is the arity of each individual feature value?

flink-ml-benchmark/src/main/resources/benchmark-conf-kmeans.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": 1,
```
lindong28, line 2: Would it be better to rename these two files as `kmeans-benchmark.json` and `naivebayes-benchmark.json`,  so that the file names are a bit more concise, and it is a bit easier to spot the key differentiator (i.e. algorithm name) between these files?

How about we rename `benchmark-conf.json` as `benchmark-demo.json` to clarify the difference between this file and other files?

flink-ml-benchmark/src/main/resources/benchmark-conf-kmeans.json
```diff
@@ -0,0 +1,23 @@
+{
```
lindong28, line 1: Can you add the Apache license header to this file and similar files in the repo? Feel free to see [1] for example.

[1] https://github.com/apache/kafka/blob/trunk/metadata/src/main/resources/common/metadata/TopicRecord.json

lindong28 (commented):

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,202 @@
...
+/** A DataGenerator which creates a table of features, label and weight. */
+public class LabeledPointWithWeightGenerator
+        implements InputDataGenerator<LabeledPointWithWeightGenerator>,
+                HasFeaturesCol<LabeledPointWithWeightGenerator>,
+                HasWeightCol<LabeledPointWithWeightGenerator>,
+                HasLabelCol<LabeledPointWithWeightGenerator>,
+                HasVectorDim<LabeledPointWithWeightGenerator> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public static final Param<Integer> LABEL_ARITY =
```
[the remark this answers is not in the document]
yunfengzhou-hub: OK. I'll reorder the variables and methods.

`StandardScalerParams` seems to have followed these conventions. I have modified `MinMaxScalerParams` for consistency.

yunfengzhou-hub:

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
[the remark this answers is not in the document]
lindong28: Thanks. Yes `MinMaxScalerParams` needs to be updated.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,194 @@
...
+            new IntParam(
+                    "labelArity",
+                    "Arity of label. "
+                            + "If set to positive value, the label would be an integer in range [0, arity - 1]. "
+                            + "If set to zero, the label would be a continuous double in range [0, 1).",
+                    2,
+                    ParamValidators.gtEq(0));
+
+    public static final Param<Integer> WEIGHT_ARITY =
```
lindong28, line 77: For benchmark, should we always generate weight as a random value in range [0, 1)? If not, what would be the other use-cases? If yes, maybe we can remove this parameter?

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,194 @@
...
+                                        getWeightArity()),
+                                new RowTypeInfo(
+                                        new TypeInformation[] {
+                                            DenseVectorTypeInfo.INSTANCE, Types.DOUBLE, Types.DOUBLE
+                                        },
+                                        new String[] {
+                                            getFeaturesCol(), getLabelCol(), getWeightCol()
+                                        }));
+
+        Table dataTable = tEnv.fromDataStream(dataStream);
```
lindong28, line 139: Should this class handle the `colNames` parameter?

Currently `DenseVectorArrayGenerator` checks that `getColNames().length > 0` when this parameter value is not null. Would it be better to verify `getColNames().length == 1` in this case? And we probably also need to take care of the case where the string contains `,` since this is a common-separated string.

Can you update `DenseVectorGenerator` as well for consistency?

I am also wondering if it would be more intuitive to update the description of `colNames` to be `An array of common-separated strings representing field names of the generated tables`.

flink-ml-benchmark/src/main/resources/naivebayes-benchmark.json
```diff
@@ -0,0 +1,44 @@
...
+        "predictionCol": "prediction",
+        "labelCol": "label",
+        "modelType": "multinomial"
+      }
+    },
+    "inputData": {
+      "className": "org.apache.flink.ml.benchmark.datagenerator.common.LabeledPointWithWeightGenerator",
+      "paramMap": {
+        "seed": 2,
+        "featuresCol": "features",
```
lindong28, line 33: Instead of asking users to specify featuresCol, weightCol and labelCol, would it be simpler to let user just specify `colNames = ["features,weight,label"]`?

flink-ml-benchmark/src/main/resources/naivebayes-benchmark.json
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+
+{
+  "version": 1,
+  "NaiveBayes": {
+    "stage": {
+      "className": "org.apache.flink.ml.classification.naivebayes.NaiveBayes",
+      "paramMap": {
+        "smoothing": 1.0,
+        "featuresCol": "features",
```
lindong28, line 23: Do you think it would make user experience better by skipping parameters whose value is the same as the default value in the config file?

lindong28 (commented):
Thanks for the update. Left some followup comments below.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
```diff
@@ -0,0 +1,194 @@
...
+        @Override
+        public Row map(Long ignored) {
+            double[] features = new double[vectorDim];
+            for (int i = 0; i < vectorDim; i++) {
+                features[i] = getValue(featureArity, 0);
+            }
+
+            double label = getValue(labelArity, 0);
+
+            double weight = getValue(weightArity, 1);
```
lindong28, line 182: By setting `discreteValueOffset = 1`, when weightArity = 0, the generated weight value would be in the range [1, 2). This appears to be inconsistent with the weightArity specification.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
yunfengzhou-hub: I agree. I'll make the change.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
yunfengzhou-hub: This comment is not applicable now.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/LabeledPointWithWeightGenerator.java
yunfengzhou-hub: I agree that `LabeledPointWithWeightGenerator` should handle colNames, instead of featuresCol, labelCol and weightCol. I also think it is a good idea to verify `getColNames().length` in these table generators.

It has not been Flink ML's practice to handle potential column names that contain commas. Existing parameters, like `HasInputCols`, distinct column names from that they are different strings in the array. If `colNames = ["features,weight,label"]`, there would be ambiguity in what each element in the array stands for.

Still, I agree that we should handle situations when there are non-common characters in column name strings. I think we can add a ParamValidator that checks for the grammar of these strings, and apply this validator to classes like `HasColNames` and `HasInputCols`. How do you like this idea?

yunfengzhou-hub:

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/InputDataGenerator.java
```diff
@@ -21,18 +21,18 @@
...
             new LongParam(
                     "numValues", "Number of data to be generated.", 10L, ParamValidators.gt(0));
 
-    Param<String[]> COL_NAMES =
-            new StringArrayParam(
+    Param<String[][]> COL_NAMES =
+            new StringArrayArrayParam(
                     "colNames",
-                    "An array of common-separated strings representing field names of data tables.",
+                    "A 2D array of strings. Each string array element represents the column names of a data table.",
```
lindong28, line 35: nits: `the column names` -> `a column name`

lindong28 (commented):
Thanks for update! LGTM overall. Left just one minor comment.

PR #102, opened 2022-05-20
[FLINK-27294] Add evaluation algorithm support for BinaryClassificationEvaluator in ML Python API

yunfengzhou-hub:


.gitignore
```diff
@@ -19,3 +19,4 @@ target
 *.pyc
 flink-ml-python/dist/
 flink-ml-python/apache_flink_ml.egg-info/
+flink-ml-python/venv/
```
HuangXingBo, line 22: IMO, we don't need to add this.  Putting venv in the source directory is not a standard way

HuangXingBo (approved):
@yunfengzhou-hub  Thanks a lot for the PR. I only have left one comment and will resolve it before merge.

PR #103, opened 2022-05-23
[FLINK-27742] Fix compatibility issue between stages

yunfengzhou-hub:
Modify Flink ML Stages to achieve the followings.

- If a stage accepts a double-typed column, modify it to accept any numerical type.
- If a stage accepts a DenseVector/DenseMatrix-typed column, modify it to accept any child class of Vector/Matrix.
- If a stage generates an int-typed label column, modify it to generate double-typed label column.


yunfengzhou-hub:
Hi @lindong28 , Could you please help review this PR? I have just changed three algorithms for now, and if we agree that the changes are proper, I'll further make the change on all algorithms.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/DenseVector.java
```diff
@@ -51,6 +51,35 @@ public double[] toArray() {
...
     }
 
+    @Override
+    public DenseVector toDense() {
+        return this;
+    }
+
+    @Override
+    public SparseVector toSparse() {
+        int nnz = 0;
```
lindong28, line 61: I am not sure `nnz` is an intuitive name. The general rule is to make variable names either self-explanatory or simple.

Could we change this to either `non_zero_count` or `count`?

Same for `ii` and `vv`.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/KnnTest.java
```diff
@@ -121,8 +124,9 @@ private static void verifyPredictionResult(Table output, String labelCol, String
                                     @Override
                                     public Tuple2<Double, Double> map(Row row) {
                                         return Tuple2.of(
-                                                (Double) row.getField(labelCol),
-                                                (Double) row.getField(predictionCol));
+                                                ((Number) row.getField(labelCol)).doubleValue(),
```
lindong28, line 127: The test code basically should mimic the user code's beahvior.

Since we have updated the stage to output double-typed label column, do we still need to do type conversion here using `Number::doubleValue()`?

Same for prediction column.

Same for `LinearSVCTest`.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/KnnTest.java
```diff
@@ -179,6 +183,42 @@ public void testFitAndPredict() throws Exception {
         verifyPredictionResult(output, knn.getLabelCol(), knn.getPredictionCol());
     }
 
+    @Test
+    public void testDataTypes() throws Exception {
```
lindong28, line 187: nits: would it be more intuitive to rename this test as `testInputTypeConversion()`?

Same for `LinearSVCTest`.

lindong28 (commented):
Thanks for the PR! Looks good overall. Left some minor comments.

lindong28:
@yunfengzhou-hub Could you help update the PR title and the commit message so that it follows the same style as existing titles/messages? For example, not every word needs to be capitalized.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/KnnTest.java
lindong28: Since the values of the output table's label column have the same type as the input table's label column specified by the user, the user should already know that this column has double values, and therefore would not need to explicitly convert this column to double type, right?

Would it be simpler to skip type-casting for the label column as well?

yunfengzhou-hub: According to offline discussion, label column is inherited from input table and remain untouched, which means it could be double or integer type. Thus we still need to use (Number)...doubleValue() in order to extract its value.

lindong28:
Thanks for the update! LGTM.

PR #107, opened 2022-06-08
[FLINK-26801] Support duplicate checkpoint aborted messages

gaoyunhaii:
This PR fixes the issue that when there are duplicate checkpoint abort message, the current checkState would directly throws exceptions, which is not expected.


lindong28:
Thanks for the fix! LGTM.

PR #108, opened 2022-06-09
[FLINK-27096] Add on-heap DataCache and optimize KMeans performance

yunfengzhou-hub:
This PR mainly adds a data cache option that allows to store cached records in heap memory in deserialized format, and further optimizes the performance of KMeans algorithm.

## Brief change log
- Add `DATA_CACHE_STRATEGY` and `DATA_CACHE_HEAP_MEMORY_FRACTION` in `IterationOptions` to control how to cache records and how much memory space can be used for caching.
- Add `OnHeapMemoryPool` to control the usage of heap memory.
- Add `OnHeapMemorySegmentReader/Writer` for the implementation  of on-heap cache.
- Modify `DataCacheWriter`, `DataCacheSnapshot` and `ListStateWithCache` to support on-heap cache with the introduced classes above.
- Add tests to verify the correctness of introduced classes.
- Refractor KMeans implementation to reduce the overhead of data transmission.


yunfengzhou-hub:
The on-heap option introduced in this PR seems not to bring obvious performance improvement during offline benchmark process while bringing in OOM risk, and thus it will be closed.

PR #109, opened 2022-06-13
[hotfix][flink-ml-lib][syntax]fix the forgotten static import method compareResultCollections from TestBaseUtils

coderappee:

  - Fix syntax error not import the static method compareResultCollections from TestBaseUtils

## Brief change log

  - *Just import the static method compareResultCollections from TestBaseUtils*

### DIFF
--- a/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/BucketizerTest.java
+++ b/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/BucketizerTest.java
@@ -43,6 +43,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


yunfengzhou-hub:
Hi @coderappee , thanks for creating this PR. Could you please illustrate why do we need to add these static imports? For now the test cases can still pass without these imports.

yunfengzhou-hub:
Got it. From the information provided in [FLINK-27454](https://issues.apache.org/jira/browse/FLINK-27454) it seems that AbstractTestBase would still extend TestBaseUtils until flink 1.16.0. I think we could temporarily close this PR for now and re-visit it when Flink ML depends on newer versions of flink.

PR #110, opened 2022-06-15
[FLINK-27096] Optimize KMeans performance

yunfengzhou-hub:
This PR optimizes KMeans Algorithm's performance with the following modifications.

1. reduce data transmission and per-round operator creation overhead
2. pre-compute the L2 norm of each dense vector and use this information to accelerate the process of finding the closest centroid id of each data point.

After adding the 1st optimization, the throughput of KMeans algorithm increased by around 10%. After introducing the 2nd optimization, the throughput increased by another 60%.

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
@@ -114,8 +114,19 @@ public class BLAS {
     }
 
     /** \sqrt(\sum_i x_i * x_i) . */
-    public static double norm2(DenseVector x) {
+    public static double norm2(Vector x) {
+        if (x instanceof DenseVector) {
+            return norm2((DenseVector) x);
+        }
+        return norm2((SparseVector) x);
+    }
+
+    private static double norm2(DenseVector x) {
         return JAVA_BLAS.dnrm2(x.size(), x.values, 1);
+    }
+
+    private static double norm2(SparseVector x) {
+        return JAVA_BLAS.dnrm2(x.values.length, x.values, 1);
     }
 
     /** x = x * a . */
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/VectorWithNorm.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/VectorWithNorm.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.linalg;
+
+import org.apache.flink.api.common.typeinfo.TypeInfo;
+import org.apache.flink.ml.linalg.typeinfo.VectorWithNormTypeInfoFactory;
+
+/** A vector with its norm. */
+@TypeInfo(VectorWithNormTypeInfoFactory.class)
+public class VectorWithNorm {
+    public final Vector vector;
+
+    public final double l2Norm;
+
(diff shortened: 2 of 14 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
```diff
@@ -153,39 +149,24 @@ public IterationBodyResult process(
...
-                                    "SelectNearestCentroid",
+                                    "CentroidsUpdateAccumulator",
                                     new TupleTypeInfo<>(
-                                            BasicTypeInfo.INT_TYPE_INFO,
-                                            DenseVectorTypeInfo.INSTANCE),
-                                    new SelectNearestCentroidOperator(distanceMeasure));
+                                            BasicArrayTypeInfo.INT_ARRAY_TYPE_INFO,
+                                            ObjectArrayTypeInfo.getInfoFor(
+                                                    DenseVectorTypeInfo.INSTANCE)),
+                                    new CentroidsUpdateAccumulator(distanceMeasure));
```
zhipeng93, line 160: One optimization from algorithm side could be pre-compute the norm of each data point and thus accelerate the computing of `findClosestCentroidId`.

A similar solution could be found [1] [2].

[1] https://github.com/alibaba/Alink/blob/9ae07329a9fb77e39c334c09c79e5016b62f59ee/core/src/main/java/com/alibaba/alink/operator/common/distance/FastDistance.java
[2] https://github.com/apache/spark/blob/2fa98a6a0971cafeb98a2148a962cf8d79381842/mllib/src/main/scala/org/apache/spark/mllib/clustering/KMeans.scala#L222

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/kmeans/KMeans.java
```diff
@@ -307,24 +266,41 @@ public void processElement2(StreamRecord<DenseVector[]> streamRecord) throws Exc
...
-                output.collect(new StreamRecord<>(Tuple2.of(closestCentroidId, point)));
+
+                BLAS.axpy(1.0, point, newCentroids[closestCentroidId]);
+                counts[closestCentroidId]++;
             }
 
+            output.collect(
+                    new StreamRecord<>(
+                            Tuple2.of(
+                                    Arrays.stream(counts).boxed().toArray(Integer[]::new),
```
lindong28, line 295: `Arrays.stream(...)` is generally much slower than an iterative approach as of now [1]. If the iterative approach is not too complex to write, how about we use the iterative approach here?

[1] https://stackoverflow.com/questions/27925954/is-arrays-streamarray-name-sum-slower-than-iterative-approach

lindong28 (commented):
Thanks for the PR!

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/VectorWithNorm.java
```diff
@@ -0,0 +1,48 @@
...
+    public VectorWithNorm(Vector vector) {
+        this(vector, BLAS.norm2(vector));
+    }
+
+    public VectorWithNorm(Vector vector, double l2Norm) {
+        this.vector = vector;
+        this.l2Norm = l2Norm;
+    }
+
+    public Vector getVector() {
```
lindong28, line 41: It seems simpler to make both `vector` and `l2Norm` public final and remove the accessor methods.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/VectorWithNorm.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.flink.api.common.typeinfo.TypeInfo;
+import org.apache.flink.ml.linalg.typeinfo.VectorWithNormTypeInfoFactory;
+
+/** A vector with its norm. */
+@TypeInfo(VectorWithNormTypeInfoFactory.class)
+public class VectorWithNorm {
+    private final Vector vector;
+
+    public double l2Norm;
```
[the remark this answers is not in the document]
yunfengzhou-hub: `l2Norm` needs to be re-written in `reuse` object in its serializer's `deserialize` and `copy` method. Thus it might not be final.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -113,11 +113,22 @@ private static double dot(SparseVector x, SparseVector y) {
...
+        }
+        return norm2((SparseVector) x);
+    }
+
     /** \sqrt(\sum_i x_i * x_i) . */
     public static double norm2(DenseVector x) {
         return JAVA_BLAS.dnrm2(x.size(), x.values, 1);
     }
 
+    public static double norm2(SparseVector x) {
```
zhipeng93, line 128: nits: norm2(SparseVector) and norm2(DenseVector) could be private.

zhipeng93 (commented):
Thanks for the updates. The PR LGTM overall. I left two minor comments.

PR #112, opened 2022-06-17
[FLINK-27096] Flush buffer at epoch watermark

yunfengzhou-hub:
This PR reduces Flink ML iteration's latency by enforcing flush at each iteration epoch watermark.

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/RecordWriterBroadcastOutput.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/broadcast/RecordWriterBroadcastOutput.java
@@ -18,6 +18,7 @@
 package org.apache.flink.iteration.broadcast;
 
 import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.iteration.IterationRecord;
 import org.apache.flink.runtime.io.network.api.writer.RecordWriter;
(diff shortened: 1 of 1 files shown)


lindong28:
@yunfengzhou-hub Thanks for the PR! LGTM. Could you explain the performance impact of this PR on the iteration benchmark with and without trivial UDF payload?

yunfengzhou-hub:
Hi @lindong28, Thanks for reviewing this PR. 

Before this PR is implemented, Flink ML stages with iterations, like KMeans, have to spend at least 100 ms for each iteration they perform, even if the payload of the UDF in the iteration body is trivial. This is because Flink ML has to wait until buffer timeout before it can forward the iteration epoch watermark to the next round.

With the flushing behavior introduced by this PR, the 100 ms overhead described above would be saved. Flink ML stages could progress to the next epoch as soon as it receives the iteration watermark, instead of being restricted by buffer timeout.

lindong28:
For the record, @yunfengzhou-hub has done benchmarks to verify the performance impact of this PR. Here are the results:

- Benchmark setup: an iterative data stream job with an almost-empty iteration body. Buffer timeout is 100 ms. Run 500 iterations.
- Before this PR, Flink's the average time per iteration is 182 ms.
- After this PR, Flink's the average time per iteration is 6.4 ms.
- Spark's average time per iteration is 11.7 ms.

lindong28:
Thanks for the information. LGTM.

PR #113, opened 2022-06-21
[FLINK-27096] Optimize OneHotEncoder performance

yunfengzhou-hub:
This PR optimizes the performance of one-hot encoder algorithm with the following modifications:

- Restructures the DAG of OneHotEncoder, so that the very first stream operator can pre-process input data with aggregation operations, so that the data transmission overhead is reduced.
- Avoids unnecessary `String.format()` operation when passing error message to `Precondition.checkArgument`.

These optimizations together reduces the net runtime of OneHotEncoder benchmark jobs to about 1/6.

- Adds example benchmark json file for OneHotEncoder
- Supports generating distinct double values in `DoubleGenerator`.
  - This modification has only slight influence on the performance of DoubleGenerator.


lindong28:
Thanks for the PR! LGTM.

PR #114, opened 2022-06-22
[FLINK-27096] Optimize VectorAssembler performance

zhipeng93:
This PR optimizes the performance of VectorAssembler. It increases the throughput by 30%-50%.

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
@@ -26,6 +26,7 @@ import org.apache.flink.ml.common.param.HasHandleInvalid;
 import org.apache.flink.ml.linalg.DenseVector;
 import org.apache.flink.ml.linalg.SparseVector;
 import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.Vectors;
 import org.apache.flink.ml.linalg.typeinfo.VectorTypeInfo;
 import org.apache.flink.ml.param.Param;
 import org.apache.flink.ml.util.ParamUtils;
@@ -42,7 +43,6 @@ import org.apache.commons.lang3.ArrayUtils;
 
 import java.io.IOException;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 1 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -80,24 +82,68 @@ public Table[] transform(Table... inputs) {
         return new Table[] {outputTable};
     }
 
-    private static class AssemblerFunc implements FlatMapFunction<Row, Row> {
+    private static class AssemblerFunc extends RichFlatMapFunction<Row, Row> {
         private final String[] inputCols;
         private final String handleInvalid;
 
+        /** The indices for assembling vectors. */
+        private transient IntArrayList indices;
```
yunfengzhou-hub, line 90: Why should we use `IntArrayList`, instead of `List<Integer>` and `Integer[]`? Would it bring performance improvement compared with these options?

How about implementing it in two loops? In the first loop, we only extract the vector/numbers, and calculate their total size. Then we allocate the integer and double arrays according to the calculated size, and assign the values in the second loop.

yunfengzhou-hub (commented):
Thanks for the PR. Left one comment as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -80,24 +82,68 @@ public Table[] transform(Table... inputs) {
...
+                        }
+                        offset += denseVector.size();
+                    } else {
+                        throw new IllegalArgumentException(
+                                "Input type has not been supported yet.");
+                    }
+                }
+
+                Vector assembledVec =
+                        new SparseVector(
```
lindong28, line 137: If the performance of using `it.unimi.dsi.fastutil.*` is not considerably faster than using a for loop to construct int[] and double[], it seems simpler to still use a for-loop instead of introducing an extra library dependency.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -80,24 +82,68 @@ public Table[] transform(Table... inputs) {
...
 
-    private static class AssemblerFunc implements FlatMapFunction<Row, Row> {
+    private static class AssemblerFunc extends RichFlatMapFunction<Row, Row> {
         private final String[] inputCols;
         private final String handleInvalid;
 
+        /** The indices for assembling vectors. */
+        private transient IntArrayList indices;
+        /** The values for assembling vectors. */
+        private transient DoubleArrayList values;
```
lindong28, line 92: It seems simpler to instantiate indices and values as local variables in the `flatMap()` method. Does re-creating these two variables for each call in the flapMap() have non-trivial performance impact?

lindong28 (commented):
Thanks for the PR! Left two minor comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
zhipeng93: Thanks for the comment. I have removed the usage of `IntArrayList` and re-implemented in two loops. There is no performance regression.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
zhipeng93: Marked resolved since I removed the use of `it.unimi.dsi.fastutil.*`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
zhipeng93: Marked resolved since I removed the use of `it.unimi.dsi.fastutil.*`.

zhipeng93:
@lindong28 @yunfengzhou-hub Thanks for the comments.  I have addressed the comments in the latest push

lindong28:
Thanks for the update. LGTM.

PR #118, opened 2022-06-27
[FLINK-28254] Add benchmark of VectorAssembler and MinMaxScaler

weibozhao:
Add benchmark of vectorAssembler and minmaxScaler

### DIFF
--- a/flink-ml-benchmark/src/main/resources/minmaxscaler-benchmark.json
+++ b/flink-ml-benchmark/src/main/resources/minmaxscaler-benchmark.json
@@ -1,0 +1,40 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+{
+  "version": 1,
+  "minmaxscaler10000000": {
+    "inputData": {
+      "className": "org.apache.flink.ml.benchmark.datagenerator.common.DenseVectorGenerator",
+      "paramMap": {
+        "vectorDim": 100,
+        "colNames": [
+          [
+            "features"
+          ]
+        ],
+        "seed": 2,
+        "numValues": 10000000
+      }
+    },
+    "stage": {
+      "className": "org.apache.flink.ml.feature.minmaxscaler.MinMaxScaler",
+      "paramMap": {
+        "inputCol": "features",
+        "outputCol": "outputCol"
+      }
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/ColsWithLabelGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+        public Row map(Long ignored) {
+            Row ret = new Row(colNames.length);
+            for (int i = 0; i < colNames.length - 1; ++i) {
+                ret.setField(i, random.nextDouble());
+            }
+            ret.setField(colNames.length - 1, random.nextDouble() > 0.5 ? 1.0 : 0.0);
+            return ret;
+        }
+    }
+}
```
yunfengzhou-hub, line 123: nit: Let's follow the common practice to add a blank line at the end of each file.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/ColsWithLabelGenerator.java
```diff
@@ -0,0 +1,123 @@
+/*
```
yunfengzhou-hub, line 1: Let's add `[FLINK-27096]` to the title and commit message of this PR. It has been the parent ticket for all PRs related to algorithm benchmark and performance improvement before 2.1 release.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/ColsWithLabelGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+                colNames[i] = getColNames()[0][i];
+                types[i] = Types.DOUBLE;
+            }
+            colNames[colNames.length - 1] = getColNames()[1][0];
+            types[colNames.length - 1] = Types.DOUBLE;
+        } else {
+            colNames = getColNames()[0];
+            types = new TypeInformation<?>[getColNames()[0].length];
+            for (int i = 0; i < colNames.length; ++i) {
+                types[i] = Types.DOUBLE;
```
yunfengzhou-hub, line 75: Do you think it would be better to reuse `DoubleGenerator` or `DenseVectorGenerator`? I can see that this data generator is used for vector assembler benchmark, so inputting multiple double/vector columns should be enough for current use cases. For now `DoubleGenerator` or `DenseVectorGenerator` only generates one double/vector column, and we can modify them to generate arbitrary number of double/vector columns, according to the number of columns acquired from `getColNames()`. The label column seems unnecessary for now.

yunfengzhou-hub (commented):
Thanks for this PR. Left some comments as below.

flink-ml-benchmark/src/main/resources/minmaxscaler-benchmark.json
```diff
@@ -0,0 +1,42 @@
...
+      "className": "org.apache.flink.ml.feature.minmaxscaler.MinMaxScaler",
+      "paramMap": {
+        "inputCol": "features",
+        "withMean": true, 
+        "withStd": true,
+        "outputCol": "outputCol"
+      }
+    }
+  }
+}
```
yunfengzhou-hub, line 42: It seems that this json file cannot be executed. Could you please check if its grammar is correct?

flink-ml-benchmark/src/main/resources/minmaxscaler-benchmark.json
```diff
@@ -0,0 +1,42 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
```
yunfengzhou-hub, line 1: Let's rebase the commits in this PR into one to improve its readability.

yunfengzhou-hub (commented):

PR #121, opened 2022-06-27
[FLINK-27715] Add pyflink examples

yunfengzhou-hub:
This PR is a continuation of #115 . It adds pyflink examples for Flink ML's algorithms.

- fixes an exception message in `IndexToStringModel.java`
- fixes a bug in `linalg.py`
- deduplicates test cases in `test_logisticregression.py`
- fixes comments and API in `stringindexer.py`
- adds test cases for `IndexToStringModel`
- adds python support for Linear Regression algorithm

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -41,6 +41,8 @@ jobs:
         run: python -m mypy --config=setup.cfg
       - name: Test the source code
         working-directory: flink-ml-python
-        run: pytest
+        run: |
+          pytest pyflink/ml
+          pytest pyflink/examples
 
 
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/stringindexer/IndexToStringModel.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/stringindexer/IndexToStringModel.java
@@ -147,15 +147,8 @@ public class IndexToStringModel
             }
 
             Row outputStrings = new Row(inputCols.length);
-            int stringId;
             for (int i = 0; i < inputCols.length; i++) {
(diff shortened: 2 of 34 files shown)


flink-ml-python/pyflink/examples/ml/__init__.py
```diff
@@ -15,3 +15,80 @@
...
+
+
+def create_url_class_loader(urls, parent_class_loader):
+    gateway = get_gateway()
+    url_class_loader = gateway.jvm.java.net.URLClassLoader(
+        to_jarray(gateway.jvm.java.net.URL, urls), parent_class_loader)
+    return url_class_loader
+
+
+java_utils.add_jars_to_context_class_loader = add_jars_to_context_class_loader
```
lindong28, line 94: Could you explain why we need to do this? For example, is this because we want to get around a bug in `pyflink.util.java_utils` before that bug is fixed in Flink?

lindong28 (commented):
Thanks for the PR! LGTM overall. Left one comment below.

flink-ml-python/pyflink/examples/ml/__init__.py
yunfengzhou-hub: Yes. It is a walk-around to a bug related to FLINK-15635 and FLINK-28002. I'll add a TODO to avoid overwriting `pyflink.util.java_utils` after this bug is fixed and released.

And according to offline discussion, I'll try to reuse the `add_jars_to_context_class_loader` method across `pyflink/ml/__init__.py` and `pyflink/examples/ml/__init__.py`.

lindong28:
Thanks for the update. LGTM.

PR #124, opened 2022-06-30
[FLINK-28224] Add document for algorithms introduced before 2.1 release

yunfengzhou-hub:
This PR adds documents for algorithms introduced before Flink ML 2.1 release.

- sets equal weight for all operator document pages so that they would be ordered alphabetically on the document website
- fixes grammar errors in previous python comments
- update the Java/Python examples of previous operators according to the example codes introduced in #115 and #121.


lindong28:
Thanks for the PR. LGTM.

PR #125, opened 2022-06-30
[FLINK-27084] Fix classloader for per-round operators

yunfengzhou-hub:
This PR provides the correct user code class loader for per-round operators so that Flink ML jobs would not throw exceptions described in FLINK-27084.


lindong28:
Thanks for the PR! LGTM.

PR #129, opened 2022-07-12
[FLINK-28500] Add Transformer for Tokenizer

zhipeng93:
- Add Transformer for Tokenizer in Flink ML. The parameters of Tokenizer are consistent with that in Spark ML.

## Brief change log
- Add Transformer for Tokenizer in Flink ML (Java & Python).
- Add unit test for Tokenizer (Java & Python)
- Add example code for Tokenizer (Java&Python)

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/tokenizer/Tokenizer.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/tokenizer/Tokenizer.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.tokenizer;
+
+import org.apache.flink.ml.api.Transformer;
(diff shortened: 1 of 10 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/tokenizer/Tokenizer.java
```diff
@@ -0,0 +1,80 @@
...
+    public Tokenizer() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+
+        Expression tokenizerUdf = Expressions.call(TokenizerUdf.class, $(getInputCol()));
+        Table output = inputs[0].addColumns(tokenizerUdf).as(getInputCol(), getOutputCol());
```
yunfengzhou-hub, line 54: Could you please verify that the current implementation preserves columns that are not `getInputCol()` after this transformation? I'm a little concerned that if there are more than two input columns for this transformer, exceptions would be thrown.

yunfengzhou-hub (commented):
Thanks for the PR. Left one comment as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/tokenizer/Tokenizer.java
zhipeng93: Good catch. I will update the code and the corresponding test here.

lindong28 (commented):
Thanks for the PR! LGTM overall. Left just one minor comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/tokenizer/Tokenizer.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.table.api.Expressions.$;
+
+/**
+ * A tokenizer is a transformer that converts the input string to lowercase and then splits it by
```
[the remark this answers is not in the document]
zhipeng93: Thanks for the comment. I have changed the Java doc of Tokenizer, VectorAssembler and also Bucketizer as well as the corresponding python files.

lindong28:
Thanks for the update. LGTM.

PR #131, opened 2022-07-15
[FLINK-28563] Add Transformer for VectorSlicer

weibozhao:
Add Transformer for VectorSlicer in Flink ML.
## Brief change log
Add Transformer for VectorSlicer in Flink ML (Java & Python).
Add unit test for VectorSlicer (Java & Python)
Add example code for VectorSlicer (Java&Python)

### DIFF
--- a/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/VectorSlicerExample.java
+++ b/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/VectorSlicerExample.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.examples.feature;
+
+import org.apache.flink.ml.feature.vectorslicer.VectorSlicer;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+
+/** Simple program that creates a VectorSlicer instance and uses it for feature engineering. */
+public class VectorSlicerExample {
+    public static void main(String[] args) {
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        StreamTableEnvironment tEnv = StreamTableEnvironment.create(env);
+
+        // Generates input data.
+        DataStream<Row> inputStream =
+                env.fromElements(
+                        Row.of(Vectors.dense(2.1, 3.1, 1.2, 3.1, 4.6)),
+                        Row.of(Vectors.dense(1.2, 3.1, 4.6, 2.1, 3.1)));
+        Table inputTable = tEnv.fromDataStream(inputStream).as("vec");
+
+        // Creates a VectorSlicer object and initializes its parameters.
+        VectorSlicer vectorSlicer =
+                new VectorSlicer().setInputCol("vec").setIndices(1, 2, 3).setOutputCol("slicedVec");
+
+        // Uses the VectorSlicer object for feature transformations.
+        Table outputTable = vectorSlicer.transform(inputTable)[0];
+
+        // Extracts and displays the results.
+        for (CloseableIterator<Row> it = outputTable.execute().collect(); it.hasNext(); ) {
+            Row row = it.next();
+
+            Vector inputValue = (Vector) row.getField(vectorSlicer.getInputCol());
+
+            Vector outputValue = (Vector) row.getField(vectorSlicer.getOutputCol());
+
+            System.out.printf("Input Value: %s \tOutput Value: %s\n", inputValue, outputValue);
+        }
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
@@ -1,0 +1,148 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.vectorslicer;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.typeinfo.VectorTypeInfo;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A Transformer that transforms a vector to a new feature, which is a sub-array of the original
+ * feature. It is useful for extracting features from a given vector.
+ *
+ * <p>Note that duplicate features are not allowed, so there can be no overlap between selected
+ * indices. If the max value of the indices is greater than the size of the input vector, it throws
+ * an IllegalArgumentException.
+ */
+public class VectorSlicer implements Transformer<VectorSlicer>, VectorSlicerParams<VectorSlicer> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public VectorSlicer() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), VectorTypeInfo.INSTANCE),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(new VectorSliceFunction(getIndices(), getInputCol()), outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static VectorSlicer load(StreamTableEnvironment env, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /**
+     * Vector slice function which transforms a vector to a new one with a sub-array of the original
+     * features.
+     */
+    private static class VectorSliceFunction implements MapFunction<Row, Row> {
+        private final Integer[] indices;
+        private final String inputCol;
+        private int maxIndex = -1;
+
+        public VectorSliceFunction(Integer[] indices, String inputCol) {
+            this.indices = indices;
+            for (Integer index : indices) {
+                maxIndex = Math.max(maxIndex, index);
+            }
+            this.inputCol = inputCol;
+        }
+
+        @Override
+        public Row map(Row row) throws Exception {
+            Vector inputVec = row.getFieldAs(inputCol);
+            Vector outputVec;
+            if (maxIndex >= inputVec.size()) {
+                throw new IllegalArgumentException(
+                        "Index value "
+                                + maxIndex
+                                + " is greater than vector size:"
+                                + inputVec.size());
+            }
+            if (inputVec instanceof DenseVector) {
+                double[] values = new double[indices.length];
+                for (int i = 0; i < indices.length; ++i) {
+                    values[i] = ((DenseVector) inputVec).values[indices[i]];
+                }
+                outputVec = new DenseVector(values);
+            } else {
+                int nnz = 0;
+                SparseVector vec = (SparseVector) inputVec;
+                int[] outputIndices = new int[indices.length];
+                double[] outputValues = new double[indices.length];
+                for (int i = 0; i < indices.length; i++) {
+                    double val = vec.get(indices[i]);
+                    if (val != 0) {
+                        outputIndices[nnz] = i;
+                        outputValues[nnz] = val;
+                        nnz++;
+                    }
+                }
+                if (nnz < outputIndices.length) {
+                    outputIndices = Arrays.copyOf(outputIndices, nnz);
+                    outputValues = Arrays.copyOf(outputValues, nnz);
+                }
+                outputVec = new SparseVector(indices.length, outputIndices, outputValues);
+            }
+            return Row.join(row, Row.of(outputVec));
+        }
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.vectorslicer;
+
+import org.apache.flink.ml.common.param.HasInputCol;
+import org.apache.flink.ml.common.param.HasOutputCol;
+import org.apache.flink.ml.param.IntArrayParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidator;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Params of {@link VectorSlicer}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface VectorSlicerParams<T> extends HasInputCol<T>, HasOutputCol<T> {
+    Param<Integer[]> INDICES =
+            new IntArrayParam(
+                    "indices",
+                    "An array of indices to select features from a vector column.",
+                    null,
+                    indicesValidator());
+
+    default Integer[] getIndices() {
+        return get(INDICES);
+    }
+
+    default T setIndices(Integer... value) {
+        return set(INDICES, value);
+    }
+
+    // Checks the indices parameter.
+    static ParamValidator<Integer[]> indicesValidator() {
+        return indices -> {
+            if (indices == null) {
+                return false;
+            }
+            for (Number ele : indices) {
+                if (ele.doubleValue() < 0) {
+                    return false;
+                }
+            }
+            Set<Integer> set = new HashSet<>(Arrays.asList(indices));
+            if (set.size() != indices.length) {
+                return false;
+            }
+            return set.size() != 0;
+        };
+    }
+}
--- a/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
+++ b/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
@@ -1,0 +1,225 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature;
+
+import org.apache.flink.api.common.restartstrategy.RestartStrategies;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.ml.feature.vectorslicer.VectorSlicer;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.ml.util.TestUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.test.util.AbstractTestBase;
+import org.apache.flink.types.Row;
+
+import org.apache.commons.collections.IteratorUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/** Tests {@link VectorSlicer}. */
+public class VectorSlicerTest extends AbstractTestBase {
+
+    private StreamTableEnvironment tEnv;
+    private Table inputDataTable;
+
+    private static final List<Row> INPUT_DATA =
+            Arrays.asList(
+                    Row.of(
+                            0,
+                            Vectors.dense(2.1, 3.1, 2.3, 3.4, 5.3, 5.1),
+                            Vectors.sparse(5, new int[] {1, 3, 4}, new double[] {0.1, 0.2, 0.3})),
+                    Row.of(
+                            1,
+                            Vectors.dense(2.3, 4.1, 1.3, 2.4, 5.1, 4.1),
+                            Vectors.sparse(5, new int[] {1, 2, 4}, new double[] {0.1, 0.2, 0.3})));
+
+    private static final DenseVector EXPECTED_OUTPUT_DATA_1 = Vectors.dense(2.1, 3.1, 2.3);
+    private static final DenseVector EXPECTED_OUTPUT_DATA_2 = Vectors.dense(2.3, 4.1, 1.3);
+
+    private static final SparseVector EXPECTED_OUTPUT_DATA_3 =
+            Vectors.sparse(3, new int[] {1}, new double[] {0.1});
+    private static final SparseVector EXPECTED_OUTPUT_DATA_4 =
+            Vectors.sparse(3, new int[] {1, 2}, new double[] {0.1, 0.2});
+
+    @Before
+    public void before() {
+        Configuration config = new Configuration();
+        config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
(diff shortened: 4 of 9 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface VectorSlicerParams<T> extends HasInputCol<T>, HasOutputCol<T> {
+    Param<Integer[]> INDICES =
+            new IntArrayParam(
+                    "indices",
+                    "An array of indices to select features from a vector column.",
+                    null,
+                    ParamValidators.nonEmptyArray());
```
yunfengzhou-hub, line 38: It might be better to add a custom `ParamValidator` to verify that the indices are all greater or equal to 0 and are non-duplicate, like `VectorSlicer.validIndices()` in spark.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A feature transformer that transforms a vector to a new one with a sub-array of the original
+ * features. It is useful for extracting features from a given vector.
+ */
```
yunfengzhou-hub, line 49: Let's add comments that describe what will happen if the indices acquired from `setIndices()` are not in order, and add corresponding test cases.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,137 @@
...
+                int nnz = 0;
+                int[] inputIndices = ((SparseVector) inputVec).indices;
+                double[] inputValues = ((SparseVector) inputVec).values;
+                int[] outputIndices = new int[indices.length];
+                double[] outputValues = new double[indices.length];
+                for (int i = 0; i < indices.length; i++) {
+                    if (indices[i] >= inputVec.size()) {
+                        throw new RuntimeException("Index is larger than vector size.");
+                    }
+                    int pos = Arrays.binarySearch(inputIndices, indices[i]);
```
yunfengzhou-hub, line 107: The `SparseVector.get()` method has already implemented the binary search. We can directly use that method without implementing these on our own.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,156 @@
...
+            } else {
+                assertNull(result.getField(outputCol));
+            }
+        }
+    }
+
+    @Test
+    public void testParam() {
+        VectorSlicer vectorSlicer = new VectorSlicer();
+        assertEquals("output", vectorSlicer.getOutputCol());
```
yunfengzhou-hub, line 111: Let's also check the inputCol and indices here.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,156 @@
...
+                    assertEquals(EXPECTED_OUTPUT_DATA_1, result.getField(outputCol));
+                }
+            } else if (result.getField(0) == (Object) 1) {
+                if (isSparse) {
+                    assertEquals(EXPECTED_OUTPUT_DATA_4, result.getField(outputCol));
+                } else {
+                    assertEquals(EXPECTED_OUTPUT_DATA_2, result.getField(outputCol));
+                }
+            } else {
+                assertNull(result.getField(outputCol));
```
yunfengzhou-hub, line 103: It seems that this line would not be reached. It might be better to remove this line or replace it with throwing an exception.

flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/VectorSlicerExample.java
```diff
@@ -0,0 +1,62 @@
...
+
+        // Extracts and displays the results.
+        for (CloseableIterator<Row> it = outputTable.execute().collect(); it.hasNext(); ) {
+            Row row = it.next();
+
+            Vector inputValue = (Vector) row.getField(vectorSlicer.getInputCol());
+
+            Vector outputValue = (Vector) row.getField(vectorSlicer.getOutputCol());
+
+            System.out.printf("Input Values: %s \tOutput Value: %s\n", inputValue, outputValue);
```
yunfengzhou-hub, line 59: nit: "Values" -> "Value"

flink-ml-python/pyflink/ml/lib/param.py
```diff
@@ -155,6 +155,27 @@ def input_cols(self) -> Tuple[str, ...]:
         return self.get_input_cols()
 
 
+class HasIndices(WithParams, ABC):
```
yunfengzhou-hub, line 158: Maybe we should implement it inside `vectorslicer.py`, instead of treating it as a common parameter. You can refer to the `K` parameter in `kmeans.py`.

flink-ml-python/pyflink/ml/lib/feature/tests/test_vectorslicer.py
```diff
@@ -0,0 +1,74 @@
...
+                    [Types.INT(), DenseVectorTypeInfo()])))
+
+        self.expected_output_data_1 = Vectors.dense(2.1, 3.1, 1.2)
+        self.expected_output_data_2 = Vectors.dense(2.3, 2.1, 1.3)
+
+    def test_param(self):
+        vector_slicer = VectorSlicer()
+
+        self.assertEqual('input', vector_slicer.input_col)
+        self.assertEqual('output', vector_slicer.output_col)
```
yunfengzhou-hub, line 46: Let's use the methods instead of accessing members, which means instead of 
```python
vector_slicer.input_col
```
Let's use
```python
vector_slicer.get_input_col()
```

Besides, let's add check for indices here as well.

flink-ml-python/pyflink/ml/lib/feature/tests/test_vectorslicer.py
```diff
@@ -0,0 +1,74 @@
...
+
+        path = os.path.join(self.temp_dir, 'test_save_load_transform_vector_slicer')
+        vector_slicer.save(path)
+        vector_slicer = VectorSlicer.load(self.t_env, path)
+
+        output_table = vector_slicer.transform(self.input_data_table)[0]
+        actual_outputs = [(result[0], result[2]) for result in
+                          self.t_env.to_data_stream(output_table).execute_and_collect()]
+
+        for actual_output in actual_outputs:
```
yunfengzhou-hub, line 70: It might be better to also check the length of the actual outputs.

flink-ml-python/pyflink/examples/ml/feature/vectorslicer_example.py
```diff
@@ -0,0 +1,64 @@
...
+            ['id', 'vec'],
+            [Types.INT(), DenseVectorTypeInfo()])))
+
+# create a vector slicer object and initialize its parameters
+vector_slicer = VectorSlicer() \
+    .set_input_col('vec') \
+    .set_indices(1, 2, 3) \
+    .set_output_col('sub_vec')
+
+# use the vector slicer model for feature engineering
```
yunfengzhou-hub, line 56: I just realized that vector slicer is not a Model. Could you please help fix this comment with something like `vector slicer instance` or `vector slicer object`, and fix other places like that in `vectorassembler_example.py`?

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
```diff
@@ -38,6 +38,22 @@ public static <T> ParamValidator<T> gtEq(double lowerBound) {
         return (value) -> value != null && ((Number) value).doubleValue() >= lowerBound;
     }
 
+    // Checks if all values in the parameter intArray is greater than or equal to lowerBound.
+    public static <T> ParamValidator<T> intArrayGtEq(double lowerBound) {
+        return value -> {
```
yunfengzhou-hub, line 43: Given that we are making this a public method, I think it might be better to make it more generic to be useful in other algorithms as well. For example, instead of just validating integer arrays, we can make it validate any numeric arrays. Or we can introduce a `forEach` method, so that in order to validate indices, we can use `ParamValidators.forEach(ParamValidators.gtEq(0))`. What do you think of these ideas? we can discuss offline about this topic.

yunfengzhou-hub (commented):
Thanks for the update. Left one comment as below. The rest of this PR looks good to me.

flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
weibozhao: I think making it more generic is better than introducing a forEach function. Introducing a forEach may make the user API more complicated. 

- ParamValidators.numericalArrayGtEq(0))
- ParamValidators.forEach(ParamValidators.gtEq(0))

yunfengzhou-hub: We agreed during an offline discussion that `ParamValidators.numericalArrayGtEq(0)` is a proper solution. For similar discussions we can directly refer to existing practices, like that in Alink or Spark.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+                }
+                outputVec = new DenseVector(values);
+            } else {
+                int nnz = 0;
+                SparseVector vec = (SparseVector) inputVec;
+                int[] outputIndices = new int[indices.length];
+                double[] outputValues = new double[indices.length];
+                for (int i = 0; i < indices.length; i++) {
+                    if (indices[i] >= inputVec.size()) {
+                        throw new RuntimeException("Index is larger than vector size.");
```
zhipeng93, line 106: Same as above.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+        public VectorSlice(Integer[] indices, String inputCol) {
+            this.indices = indices;
+            this.inputCol = inputCol;
+        }
+
+        @Override
+        public Row map(Row row) throws Exception {
+            Vector inputVec = row.getFieldAs(inputCol);
+            Vector outputVec;
+            Arrays.sort(indices);
```
zhipeng93, line 89: How about move `sort` to the constructor method for high efficiency?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(new VectorSlice(getIndices(), getInputCol()), outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    /** Vector slice function. */
+    public static class VectorSlice implements MapFunction<Row, Row> {
```
zhipeng93, line 76: nit: This method could be private and moved to the end of this java class.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,166 @@
...
+                    assertEquals(EXPECTED_OUTPUT_DATA_2, result.getField(outputCol));
+                }
+            }
+        }
+    }
+
+    @Test
+    public void testParam() {
+        VectorSlicer vectorSlicer = new VectorSlicer();
+        assertEquals("output", vectorSlicer.getOutputCol());
```
zhipeng93, line 109: Let's check the parameters in the same order, i.e., "inputCol, outputCol, indices" for consistency.

zhipeng93 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+    public static VectorSlicer load(StreamTableEnvironment env, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /** Vector slice function. */
```
zhipeng93, line 89: Could you update the java doc here and let it contains more information? `Vector slice function` contains zero information here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        @Override
+        public Row map(Row row) throws Exception {
+            Vector inputVec = row.getFieldAs(inputCol);
+            Vector outputVec;
+            if (inputVec instanceof DenseVector) {
+                double[] values = new double[indices.length];
+                for (int i = 0; i < indices.length; ++i) {
+                    if (indices[i] >= inputVec.size()) {
```
zhipeng93, line 107: The check at Line#107 and Line#119 could be moved to Line#102 since the indices is sorted for readbility and efficiency.

Also how about we update the error message as `Index value is greater than vector size:  + inputVec.size`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /** Vector slice function. */
+    private static class VectorSlice implements MapFunction<Row, Row> {
+        private final Integer[] indices;
+        private final String inputCol;
+
+        public VectorSlice(Integer[] indices, String inputCol) {
+            this.indices = indices;
+            Arrays.sort(this.indices);
```
zhipeng93, line 96: nit: we could sort the indices outside this function, i.e., before constructing the job graph before Line#68.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A transformer that transforms a vector to a new one with a sub-array of the original features. It
```
zhipeng93, line 47: nit: A Transformer...

flink-ml-python/pyflink/ml/lib/feature/vectorslicer.py
```diff
@@ -0,0 +1,70 @@
...
+        return self.get(self.INDICES)
+
+    @property
+    def indices(self) -> Tuple[int, ...]:
+        return self.get_indices()
+
+
+class VectorSlicer(JavaFeatureTransformer, _VectorSlicerParams):
+    """
+    A feature transformer that transforms a vector to a new one with a sub-array of the original
```
zhipeng93, line 57: nit: make the python doc consistent with java doc.

zhipeng93 (commented):
Thanks for the update. I left some comments below.

flink-ml-core/src/main/java/org/apache/flink/ml/param/ParamValidators.java
```diff
@@ -38,6 +38,22 @@ public static <T> ParamValidator<T> gtEq(double lowerBound) {
         return (value) -> value != null && ((Number) value).doubleValue() >= lowerBound;
     }
 
+    // Checks if all values in the parameter numericalArray is greater than or equal to lowerBound.
```
yunfengzhou-hub, line 41: nit: `numericalArray` is not a Proper Noun in this context, as we do not have any class with this name. Consider expressing it in other ways, like "array-typed parameter" or "numerical array parameter".

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /** Vector slice function. */
+    private static class VectorSlice implements MapFunction<Row, Row> {
```
yunfengzhou-hub, line 90: nit: Let's rename this class to `VectorSliceFunction`, as class names usually end with nouns.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.flink.ml.feature.vectorslicer;
+
+import org.apache.flink.ml.common.param.HasInputCol;
+import org.apache.flink.ml.common.param.HasOutputCol;
+import org.apache.flink.ml.param.IntArrayParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+
+/**
+ * Params of VectorSlicer.
```
yunfengzhou-hub, line 28: nit: `{@link VectorSlicer}`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/** Tests VectorSlicer. */
```
yunfengzhou-hub, line 46: nit: {@link VectorSlicer}`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,138 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A transformer that transforms a vector to a new one with a sub-array of the original features. It
+ * is useful for extracting features from a given vector. If the indices acquired from setIndices()
+ * are not in order, the indices of the result vector will be sorted.
+ */
```
yunfengzhou-hub, line 50: What will happen if the max indices are larger than the size of the input vector? It might be better to have a clear definition of the behavior in this situation and add descriptions in JavaDoc and test cases.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,166 @@
...
+                } else {
+                    assertEquals(EXPECTED_OUTPUT_DATA_1, result.getField(outputCol));
+                }
+            } else if (result.getField(0) == (Object) 1) {
+                if (isSparse) {
+                    assertEquals(EXPECTED_OUTPUT_DATA_4, result.getField(outputCol));
+                } else {
+                    assertEquals(EXPECTED_OUTPUT_DATA_2, result.getField(outputCol));
+                }
+            }
```
yunfengzhou-hub, line 102: nit: Let's add an `else` here to handle unexpected situations.

flink-ml-python/pyflink/ml/lib/feature/vectorslicer.py
```diff
@@ -0,0 +1,70 @@
...
+):
+    """
+    Params for :class:`VectorSlicer`.
+    """
+
+    INDICES: Param[Tuple[int, ...]] = IntArrayParam(
+        "indices",
+        "An array of indices to select features from a vector column.",
+        None,
+        ParamValidators.non_empty_array())
```
yunfengzhou-hub, line 39: Let's use the same validator for Java and Python implementation.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface VectorSlicerParams<T> extends HasInputCol<T>, HasOutputCol<T> {
+    Param<Integer[]> INDICES =
+            new IntArrayParam(
+                    "indices",
+                    "An array of indices to select features from a vector column.",
+                    null,
+                    ParamValidators.numericalArrayGtEq(0));
```
yunfengzhou-hub, line 38: I just found that in order to align the behavior of this algorithm with existing practices, we need to make sure that the input indices are not duplicated. Could you please improve the param validator with this function, add JavaDoc to explain the behavior when there are duplicated input indices, and add corresponding tests?

Given that we need to verify the input integers are non-negative and non-duplicate, do you think it would be better to add a unique private param validator in this class to verify this specified case, instead of trying to add a common validator?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,143 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * VectorSlicer is a transformer that transforms a vector to a new one with a sub-array of the
+ * original features. It is useful for extracting features from a given vector. If the indices
+ * acquired from setIndices() are not in order, the indices of the result vector will be sorted. If
```
yunfengzhou-hub, line 49: The existing practice in Spark seems not to re-order the given indices before slicing, which means if the provided indices are not in order, it will arrange the indices in the order given. Could you please modify the corresponding implementation, unit test and JavaDocs?

flink-ml-python/pyflink/ml/lib/feature/vectorslicer.py
```diff
@@ -0,0 +1,71 @@
...
+    @property
+    def indices(self) -> Tuple[int, ...]:
+        return self.get_indices()
+
+
+class VectorSlicer(JavaFeatureTransformer, _VectorSlicerParams):
+    """
+    VectorSlicer is a transformer that transforms a vector to a new one with a sub-array of the
+    original features. It is useful for extracting features from a given vector. If the indices
+    acquired from setIndices() are not in order, the indices of the result vector will be sorted.
```
yunfengzhou-hub, line 59: The comment here is inconsistent with that in JavaDoc.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,168 @@
...
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/** Tests {@link VectorSlicer}. */
+public class VectorSlicerTest extends AbstractTestBase {
```
yunfengzhou-hub, line 47: Could you please add unit tests for the following corner cases?
- The `indices` is an empty array.
- The length of the vectors is not equal, but the max value in the `indices` is smaller than the minimum length.
- The length of the vectors is not equal, and the max value in the `indices` is larger than the minimum length.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
weibozhao: I think one validator can not do all the checks of this parameter. We just do some checks. The duplicated check may do in the algorithm code. The common validator only do one kind of checks. We can do the other checks in some other place, just like the begin of the algorithm. 

What do you think about.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
weibozhao: I have add these ut.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicerParams.java
yunfengzhou-hub: I prefer to add an exclusive validator for VectorSlicer, as that is what spark `VectorSlicer.validIndices` has done.

If we add `checkIndices()`, it means that exceptions would be thrown in `transform()` rather than `setIndices()`. I think the latter one has better readability.

zhipeng93: +1 to add a validator for VectorSlicer. If we agree to do this, we probably should also remove the numericalArrayGtEq since it is not needed any more then.

weibozhao: OK

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,217 @@
...
+                    "Index value 10 is greater than vector size:6",
+                    e.getCause().getCause().getCause().getCause().getCause().getMessage());
+            assertEquals(
+                    IllegalArgumentException.class,
+                    e.getCause().getCause().getCause().getCause().getCause().getClass());
+        }
+    }
+
+    @Test
+    public void testIndicesSallerThanZero() {
```
yunfengzhou-hub, line 167: nit: smaller

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,217 @@
...
+            DataStream<Row> dataStream = tEnv.toDataStream(output);
+            IteratorUtils.toList(dataStream.executeAndCollect());
+            Assert.fail("Expected RuntimeException");
+        } catch (Exception e) {
+            assertEquals(
+                    "Index value 10 is greater than vector size:6",
+                    e.getCause().getCause().getCause().getCause().getCause().getMessage());
+            assertEquals(
+                    IllegalArgumentException.class,
+                    e.getCause().getCause().getCause().getCause().getCause().getClass());
```
yunfengzhou-hub, line 162: nit: `ExceptionUtils.getRootCause(e)`

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,217 @@
...
+                new VectorSlicer().setInputCol("vec").setOutputCol("sliceVec").setIndices(0, 1, 2);
+        VectorSlicer loadedVectorSlicer =
+                TestUtils.saveAndReload(
+                        tEnv, vectorSlicer, TEMPORARY_FOLDER.newFolder().getAbsolutePath());
+        Table output = loadedVectorSlicer.transform(inputDataTable)[0];
+        verifyOutputResult(output, loadedVectorSlicer.getOutputCol(), false);
+    }
+
+    @Test
+    public void testEmptyIndices() {
```
yunfengzhou-hub, line 132: It seems that spark `VectorSlicer` treats an empty array as a valid input for indices. Can we also follow this convention?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,217 @@
...
+            VectorSlicer vectorSlicer =
+                    new VectorSlicer()
+                            .setInputCol("vec")
+                            .setOutputCol("sliceVec")
+                            .setIndices(1, -2);
+            vectorSlicer.transform(inputDataTable);
+            Assert.fail("Expected IllegalArgumentException");
+        } catch (Exception e) {
+            assertEquals("Parameter indices is given an invalid value {1,-2}", e.getMessage());
+            assertEquals(IllegalArgumentException.class, ((Throwable) e).getClass());
```
yunfengzhou-hub, line 178: nit: we may not need to cast `e` to `Throwable` before invoking `getClass()`. Same for other places.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+    @Test
+    public void testDuplicateIndices() {
+        try {
+            VectorSlicer vectorSlicer =
+                    new VectorSlicer()
+                            .setInputCol("vec")
+                            .setOutputCol("sliceVec")
+                            .setIndices(1, 1, 3);
+            vectorSlicer.transform(inputDataTable);
```
yunfengzhou-hub, line 190: We can remove this line because an exception would be thrown before `transform()` is invoked. Same for other test cases.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,217 @@
...
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/** Tests {@link VectorSlicer}. */
+public class VectorSlicerTest extends AbstractTestBase {
```
yunfengzhou-hub, line 48: Let's also add a test case to verify when input indices are unordered.

flink-ml-python/pyflink/ml/core/param.py
```diff
@@ -228,6 +228,22 @@ def validate(self, value: Tuple[T]) -> bool:
 
         return NonEmptyArray()
 
+    @staticmethod
+    def numerical_array_gt_eq(lower_bound: int) -> ParamValidator[Tuple[int]]:
```
yunfengzhou-hub, line 232: This method can be removed now.

flink-ml-python/pyflink/ml/lib/feature/vectorslicer.py
```diff
@@ -0,0 +1,71 @@
...
+    @property
+    def indices(self) -> Tuple[int, ...]:
+        return self.get_indices()
+
+
+class VectorSlicer(JavaFeatureTransformer, _VectorSlicerParams):
+    """
+    VectorSlicer is a transformer that transforms a vector to a new one with a sub-array of the
+    original features. It is useful for extracting features from a given vector. If the max
+    indices are larger than the size of the input vector, it will throw an IllegalArgumentException.
```
yunfengzhou-hub, line 59: nit: This document is inconsistent with that in Java.

yunfengzhou-hub (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
weibozhao: Empty indices means extracting nothing and return a vector with no values, which has no means. 
I don't think all things that spark done make sence.

zhipeng93: I agree with @weibozhao . Using a empty array as the indices seems not a valid use case. 

If the indices is an empty array, the output should be a table with each data point as an array with zero elements. We can simply acheive this via a map.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorslicer/VectorSlicer.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
```
zhipeng93, line 46: How about we update the java doc as follows:

```
A Transformer that transforms a vector to a new feature, which is a sub-array of the original feature. It is useful for extracting features from a given vector. 

<p>Note that  duplicate features are not allowed, so there can be no overlap between selected indices. If the max value of the indices is greater than the size of the input vector, it throws an IllegalArgumentException.

```

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,229 @@
...
+            assertEquals("Parameter indices is given an invalid value {1,-2}", e.getMessage());
+            assertEquals(IllegalArgumentException.class, e.getClass());
+        }
+    }
+
+    @Test
+    public void testDuplicateIndices() {
+        try {
+            new VectorSlicer().setInputCol("vec").setOutputCol("sliceVec").setIndices(1, 1, 3);
+            Assert.fail("Expected IllegalArgumentException");
```
zhipeng93, line 180: what about replace this line with `fail();`? We are assuming that the code never gets here. And it is not a IllegalArgumentException.

Same for other `Assert.fails`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorSlicerTest.java
```diff
@@ -0,0 +1,229 @@
...
+        } catch (Exception e) {
+            assertEquals("Parameter indices is given an invalid value {1,1,3}", e.getMessage());
+            assertEquals(IllegalArgumentException.class, ((Throwable) e).getClass());
```
zhipeng93, line 183: Also the check for `exception class type` seems unnecessary since we already checked the error message.

Same for other checks.

zhipeng93 (commented):
Thanks for the update. LGTM overall. I only left two minor comments.

zhipeng93:
Thanks for the update. LGTM.

PR #133, opened 2022-07-19
[FLINK-28601] Add Transformer for FeatureHasher

weibozhao:
Add Transformer for FeatureHasher.
## Brief change log
Add Transformer for FeatureHasher in Flink ML (Java & Python).
Add unit test for FeatureHasher (Java & Python)
Add example code for FeatureHasher (Java&Python)

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasCategoricalCols.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasCategoricalCols.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.StringArrayParam;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared categoricalCols param. */
+public interface HasCategoricalCols<T> extends WithParams<T> {
+    Param<String[]> CATEGORICAL_COLS =
+            new StringArrayParam(
+                    "categoricalCols",
+                    "Categorical column names.",
+                    new String[] {},
+                    ParamValidators.notNull());
+
+    default String[] getCategoricalCols() {
+        return get(CATEGORICAL_COLS);
+    }
+
+    default T setCategoricalCols(String... value) {
+        return set(CATEGORICAL_COLS, value);
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasNumFeatures.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasNumFeatures.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.IntParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared num features param. */
+public interface HasNumFeatures<T> extends WithParams<T> {
+    Param<Integer> NUM_FEATURES =
+            new IntParam(
+                    "numFeatures",
+                    "The number of features. It will be the length of the output vector.",
+                    262144,
+                    ParamValidators.gt(0));
+
+    default int getNumFeatures() {
+        return get(NUM_FEATURES);
+    }
+
+    default T setNumFeatures(int value) {
+        set(NUM_FEATURES, value);
+        return (T) this;
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
@@ -1,0 +1,209 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.featurehasher;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.typeinfo.VectorTypeInfo;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.catalog.ResolvedSchema;
+import org.apache.flink.table.types.DataType;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import static org.apache.flink.shaded.guava30.com.google.common.hash.Hashing.murmur3_32;
+
+/**
+ * A Transformer that transforms a set of categorical or numerical features into a sparse vector of
+ * a specified dimension. The rules of hashing categorical columns and numerical columns are as
+ * follows:
+ *
+ * <ul>
+ *   <li>For numerical columns, the index of this feature in the output vector is the hash value of
+ *       the column name and its correponding value is the same as the input.
+ *   <li>For categorical columns, the index of this feature in the output vector is the hash value
+ *       of the string "column_name=value" and the corresponding value is 1.0.
+ * </ul>
+ *
+ * <p>If multiple features are projected into the same column, the output values are accumulated.
+ * For the hashing trick, see https://en.wikipedia.org/wiki/Feature_hashing for details.
+ */
+public class FeatureHasher
+        implements Transformer<FeatureHasher>, FeatureHasherParams<FeatureHasher> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final org.apache.flink.shaded.guava30.com.google.common.hash.HashFunction HASH =
+            murmur3_32(0);
+
+    public FeatureHasher() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        ResolvedSchema tableSchema = inputs[0].getResolvedSchema();
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(tableSchema);
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), VectorTypeInfo.INSTANCE),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                new HashFunction(
+                                        getInputCols(),
+                                        generateCategoricalCols(
+                                                tableSchema, getInputCols(), getCategoricalCols()),
+                                        getNumFeatures()),
+                                outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    /**
+     * The main logic for transforming the categorical and numerical features into a sparse vector.
+     * It uses MurMurHash3 to compute the transformed index in the output vector. If multiple
+     * features are projected to the same column, their values are accumulated.
+     */
+    private static class HashFunction implements MapFunction<Row, Row> {
+        private final String[] categoricalCols;
+        private final int numFeatures;
+        private final String[] numericCols;
+
+        public HashFunction(String[] inputCols, String[] categoricalCols, int numFeatures) {
+            this.categoricalCols = categoricalCols;
+            this.numFeatures = numFeatures;
+            this.numericCols = ArrayUtils.removeElements(inputCols, this.categoricalCols);
+        }
+
+        @Override
+        public Row map(Row row) {
+            TreeMap<Integer, Double> feature = new TreeMap<>();
+            for (String col : numericCols) {
+                if (null != row.getField(col)) {
+                    double value = ((Number) row.getFieldAs(col)).doubleValue();
+                    updateMap(col, value, feature, numFeatures);
+                }
+            }
+            for (String col : categoricalCols) {
+                if (null != row.getField(col)) {
+                    updateMap(col + "=" + row.getField(col), 1.0, feature, numFeatures);
+                }
+            }
+            int nnz = feature.size();
+            int[] indices = new int[nnz];
+            double[] values = new double[nnz];
+            int pos = 0;
+            for (Map.Entry<Integer, Double> entry : feature.entrySet()) {
+                indices[pos] = entry.getKey();
+                values[pos] = entry.getValue();
+                pos++;
+            }
+            return Row.join(row, Row.of(new SparseVector(numFeatures, indices, values)));
+        }
+    }
+
+    private String[] generateCategoricalCols(
+            ResolvedSchema tableSchema, String[] inputCols, String[] categoricalCols) {
+        if (null == inputCols) {
+            return categoricalCols;
+        }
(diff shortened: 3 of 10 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasherParams.java
```diff
@@ -0,0 +1,51 @@
...
+ * @param <T> The class type of this instance.
+ */
+public interface FeatureHasherParams<T>
+        extends HasInputCols<T>, HasOutputCol<T>, HasCategoricalCols<T> {
+
+    Param<Integer> NUM_FEATURES =
+            new IntParam(
+                    "numFeatures",
+                    "The number of features. It will be the length of the output vector.",
+                    1 << 18,
```
yunfengzhou-hub, line 40: The default value for `numFeatures` is `262144`. Is this value or `1<<18` has any special meaning?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
```diff
@@ -0,0 +1,143 @@
...
+        return new Table[] {outputTable};
+    }
+
+    private static class HashFunc implements MapFunction<Row, Row> {
+        private final String[] categoricalCols;
+        private final int numFeatures;
+        private final String[] numericCols;
+
+        public HashFunc(String[] inputCols, String[] handleInvalid, int numFeatures) {
+            this.categoricalCols = handleInvalid;
```
yunfengzhou-hub, line 88: It might be better to rename `handleInvalid` to `categoricalCols`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
```diff
@@ -0,0 +1,143 @@
...
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+    private static final HashFunction HASH = murmur3_32(0);
+
+    public FeatureHasher() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
```
yunfengzhou-hub, line 64: It might be better to check it whether `categoricalCols` are a subarray of `inputCols`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
```diff
@@ -0,0 +1,143 @@
...
+                    double value = ((Number) row.getFieldAs(col)).doubleValue();
+                    updateMap(col, value, feature, numFeatures);
+                }
+            }
+            for (String col : categoricalCols) {
+                if (null != row.getField(col)) {
+                    updateMap(col + "=" + row.getField(col), 1.0, feature, numFeatures);
+                }
+            }
+            return Row.join(row, Row.of(new SparseVector(numFeatures, feature)));
```
yunfengzhou-hub, line 107: It might be better to extract the indices and values from the TreeMap here and use the original constructor of `SparseVector`. I'm not sure the newly-introduced SparseVector that accepts a TreeMap would be widely used.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
```diff
@@ -0,0 +1,143 @@
...
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), VectorTypeInfo.INSTANCE),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                new HashFunc(
+                                        getInputCols(), getCategoricalCols(), getNumFeatures()),
```
yunfengzhou-hub, line 76: Users should be allowed not to set `categoricalCols`, if they want all input columns to b numerical. For now, if users don't set this parameter, the `getCategoricalCols()` here would throw an exception.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasherParams.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+
+/**
+ * Params of FeatureHasher.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface FeatureHasherParams<T>
+        extends HasInputCols<T>, HasOutputCol<T>, HasCategoricalCols<T> {
```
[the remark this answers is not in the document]
weibozhao: categoricalCol is a common parameter used by more than one algorithms, such as GBDT, DecisionTree, RandomForest , featureHash, and so on. 

I don't agree with your idea —— if a parameter only used by one algorithm, it should be inclusive parameter. 

HasDecayFactor, HasDistanceMeasure, HasMultiClass also used by one algorithm, but it as common parameter.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasherParams.java
weibozhao: I think 262144 is better.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
weibozhao: In Alink, this constructor with Map is used by ItemCF, Naive Bayes, FeatureHasher and vectorAssembler. 

SparseVector also called kv data structure. Then I add this constructor.

yunfengzhou-hub: Got it. Then I'm OK with adding this constructor.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/FeatureHasherTest.java
```diff
@@ -0,0 +1,147 @@
...
+            } else if (result.getField(0) == (Object) 1) {
+                assertEquals(EXPECTED_OUTPUT_DATA_2, result.getField(outputCol));
+            }
+        }
+    }
+
+    @Test
+    public void testParam() {
+        FeatureHasher featureHasher = new FeatureHasher();
+        assertEquals("output", featureHasher.getOutputCol());
```
yunfengzhou-hub, line 88: Let's add validations for the default value of other parameters here as well.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/FeatureHasherTest.java
```diff
@@ -0,0 +1,147 @@
...
+        featureHasher
+                .setInputCols("f0", "f1", "f2")
+                .setCategoricalCols("f0", "f2")
+                .setOutputCol("vec")
+                .setNumFeatures(1000);
+        assertArrayEquals(new String[] {"f0", "f1", "f2"}, featureHasher.getInputCols());
+        assertEquals(1000, featureHasher.getNumFeatures());
+        assertEquals("vec", featureHasher.getOutputCol());
```
yunfengzhou-hub, line 96: `categoricalCols` has not been verified yet.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/FeatureHasherTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/** Tests FeatureHasher. */
```
yunfengzhou-hub, line 45: nit: `{@link FeatureHasher}`

flink-ml-python/pyflink/ml/lib/param.py
```diff
@@ -155,6 +155,48 @@ def input_cols(self) -> Tuple[str, ...]:
...
 
+class HasCategoricalCols(WithParams, ABC):
+    """
+    Base class for the shared input cols param.
+    """
+    CATEGORICAL_COLS: Param[Tuple[str, ...]] = StringArrayParam(
+        "categorical_cols",
+        "Categorical column names.",
+        None,
+        ParamValidators.non_empty_array())
```
yunfengzhou-hub, line 166: It might be better to align this parameter with that in Java API. Java `HasCategoricalCols` does not have this validator.

flink-ml-python/pyflink/ml/lib/param.py
```diff
@@ -155,6 +155,48 @@ def input_cols(self) -> Tuple[str, ...]:
         return self.get_input_cols()
 
 
+class HasCategoricalCols(WithParams, ABC):
+    """
+    Base class for the shared input cols param.
```
yunfengzhou-hub, line 160: nit: categorical columns.

flink-ml-python/pyflink/ml/lib/feature/tests/test_feature_hasher.py
```diff
@@ -0,0 +1,78 @@
...
+
+from pyflink.ml.core.linalg import Vectors
+from pyflink.ml.lib.feature.featurehasher import FeatureHasher
+from pyflink.ml.tests.test_utils import PyFlinkMLTestCase
+
+
+class FeatureHasherTest(PyFlinkMLTestCase):
+    def setUp(self):
+        super(FeatureHasherTest, self).setUp()
+        # TODO: Add test for handling invalid values after FLINK-27797 is resolved.
```
yunfengzhou-hub, line 30: nit: FeatureHasher does not need to handle invalid values.

yunfengzhou-hub (commented):

flink-ml-core/src/test/java/org/apache/flink/ml/linalg/SparseVectorTest.java
```diff
@@ -47,6 +49,16 @@ public void testConstructor() {
...
+
+        Map<Integer, Double> map = new HashMap<>(4);
+        map.put(0, 0.1);
+        map.put(2, 0.3);
+        map.put(3, 0.4);
+        SparseVector mapVector = Vectors.sparse(n, map);
+        assertEquals(n, mapVector.n);
+        assertArrayEquals(indices, mapVector.indices);
+        assertArrayEquals(values, mapVector.values, 1e-5);
+        assertEquals("(4, [0, 2, 3], [0.1, 0.3, 0.4])", mapVector.toString());
```
yunfengzhou-hub, line 61: Can we also add test cases for the following situations?
- Indices are invalid. e.g., indices are negative or larger than the size of the vector.
- Indices contain duplicated values.
- Some values are 0.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/FeatureHasherTest.java
```diff
@@ -0,0 +1,150 @@
...
+    private void verifyOutputResult(Table output, String outputCol) throws Exception {
+        DataStream<Row> dataStream = tEnv.toDataStream(output);
+        List<Row> results = IteratorUtils.toList(dataStream.executeAndCollect());
+        assertEquals(2, results.size());
+        for (Row result : results) {
+            if (result.getField(0) == (Object) 0) {
+                assertEquals(EXPECTED_OUTPUT_DATA_1, result.getField(outputCol));
+            } else if (result.getField(0) == (Object) 1) {
+                assertEquals(EXPECTED_OUTPUT_DATA_2, result.getField(outputCol));
+            }
```
yunfengzhou-hub, line 81: nit: Let's add code like follows here to avoid invalid output values.
```java
else {
    throw new xxxException("unknown output value ...");
}
```

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/FeatureHasherTest.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    @Test
+    public void testParam() {
+        FeatureHasher featureHasher = new FeatureHasher();
+        assertEquals("output", featureHasher.getOutputCol());
+        assertArrayEquals(new String[] {}, featureHasher.getCategoricalCols());
+        assertEquals(262144, featureHasher.getNumFeatures());
```
yunfengzhou-hub, line 90: nit: Let's use the same order for parameters when getting initial params, set params, and getting configured params.

flink-ml-python/pyflink/ml/lib/param.py
```diff
@@ -155,6 +155,48 @@ def input_cols(self) -> Tuple[str, ...]:
...
+        return self.get(self.CATEGORICAL_COLS)
+
+    @property
+    def categorical_cols(self) -> Tuple[str, ...]:
+        return self.get_categorical_cols()
+
+
+class HasNumFeatures(WithParams, ABC):
+    """
+    Base class for the shared maxIter param.
```
yunfengzhou-hub, line 181: nit: maxIter -> numFeatures, or Number of features.

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/SparseVector.java
```diff
@@ -42,6 +44,23 @@ public SparseVector(int n, int[] indices, double[] values) {
...
+        this.values = new double[nnz];
+        int pos = 0;
+        for (Map.Entry<Integer, Double> entry : kv.entrySet()) {
+            indices[pos] = entry.getKey();
+            values[pos] = entry.getValue();
+            pos++;
+        }
+        if (!(kv instanceof TreeMap)) {
+            sortIndices();
+            validateSortedData();
```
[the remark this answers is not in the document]
weibozhao: OK, I will refine it.

flink-ml-core/src/test/java/org/apache/flink/ml/linalg/SparseVectorTest.java
weibozhao: In map constructor, indices are keys in map. There is no possible existing duplicated indices.

1 and 3  has refined.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
zhipeng93: I hold a bit different opinion. It is kind of wierd to me to let SparseVector accept a `map` as a parameter for the constructor. 

Essentially what we need is a utility function that transforms a Map<Integer, Double> to `int[] indices, double[] values`.  What do you think @yunfengzhou-hub @weibozhao ?

yunfengzhou-hub: I just noticed that we have already implemented NaiveBayes and VectorAssembler without this constructor. If the new constructor is only used in FeatureHasher and ItemCF, I would prefer not to add this constructor.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasCategoricalCols.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.StringArrayParam;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared categoricalCols param. */
+public interface HasCategoricalCols<T> extends WithParams<T> {
```
yunfengzhou-hub, line 27: Would it be better to merge this class into `FeatureHasherParams`? We can extract this as a common parameter when we implement other algorithms that use it.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasNumFeatures.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared num features param. */
+public interface HasNumFeatures<T> extends WithParams<T> {
+    Param<Integer> NUM_FEATURES =
+            new IntParam(
+                    "numFeatures",
+                    "The number of features. It will be the length of the output vector.",
+                    262144,
```
yunfengzhou-hub, line 32: Could you please help investigate why spark uses `262144` as the default value? It is proper for us to reference existing practice, but we hope to make it clear why it is reasonable.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import static org.apache.flink.shaded.guava30.com.google.common.hash.Hashing.murmur3_32;
+
+/**
+ * A transformer that projects a set of categorical or numerical features into a vector with fixed
+ * dimension. This is done using the hashing trick (https://en.wikipedia.org/wiki/Feature_hashing)
+ * to map features to indices in the feature vector.
```
yunfengzhou-hub, line 57: nit: There are some grammar errors in JavaDocs. For example, "with fixed dimension" would be better in plural form or has articles.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
```diff
@@ -0,0 +1,186 @@
...
+                                        getInputCols(),
+                                        generateCategoricalCols(
+                                                tableSchema, getInputCols(), getCategoricalCols()),
+                                        getNumFeatures()),
+                                outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    private static class HashFunc implements MapFunction<Row, Row> {
```
yunfengzhou-hub, line 92: Could you please add JavaDoc explaining the function of this class？

flink-ml-python/pyflink/ml/lib/param.py
```diff
@@ -155,6 +155,48 @@ def input_cols(self) -> Tuple[str, ...]:
...
 
 
+class HasCategoricalCols(WithParams, ABC):
+    """
+    Base class for the shared categorical cols param.
+    """
+    CATEGORICAL_COLS: Param[Tuple[str, ...]] = StringArrayParam(
+        "categorical_cols",
+        "Categorical column names.",
+        None,
```
yunfengzhou-hub, line 165: The default value is inconsistent with that in Java.

flink-ml-python/pyflink/ml/lib/feature/tests/test_feature_hasher.py
```diff
@@ -0,0 +1,77 @@
...
+
+        path = os.path.join(self.temp_dir, 'test_save_load_transform_feature_hasher')
+        feature_hasher.save(path)
+        feature_hasher = FeatureHasher.load(self.t_env, path)
+
+        output_table = feature_hasher.transform(self.input_data_table)[0]
+        actual_outputs = [(result[0], result[4]) for result in
+                          self.t_env.to_data_stream(output_table).execute_and_collect()]
+
+        for actual_output in actual_outputs:
```
yunfengzhou-hub, line 73: nit: when `actual_output` is empty`, this test case still passes. Let's avoid such situations.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/featurehasher/FeatureHasher.java
weibozhao: I agree with you @zhipeng93, I will just add the transformation code in my FeatureHasher algorithm, if some other algorithm use this transform, then make it a common function.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasNumFeatures.java
weibozhao: I try to find the reason why the default value is 262144, but it's no result. Maybe, it is just an empirical value.

flink-ml-python/pyflink/ml/lib/feature/tests/test_feature_hasher.py
weibozhao: If actual_output is empty, `actual_output[0] == 0` will out of boundary.

yunfengzhou-hub: Sorry for my ambiguity. I mean when `actual_outputs` is empty.

weibozhao: I think before `actual_outputs` is empty, some other error or exception has happened already.
Then, `actual_outputs` is empty will never happen.

yunfengzhou-hub: There are cases when no exception or error would happen in advance. Let me illustrate with an example about `VectorAssembler`.
Suppose we forgot to add the following code:
https://github.com/weibozhao/flink-ml/blob/FLINK-28601/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java#L134
And we do not check for the size of the outputs, which means removing the following line:
https://github.com/weibozhao/flink-ml/blob/FLINK-28601/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java#L93
You will find that all test cases in `VectorAssemblerTest` can still pass, without throwing any exception or error.
Thus I think it is necessary to add checks about the output size.

weibozhao: OK. I get your idea, I will add this check.

zhipeng93 (commented):
Thanks for the PR. It LGTM. I left some comments about the java docs.

zhipeng93:
Thanks for the update @weibozhao. LGTM.

PR #135, opened 2022-07-21
[FLINK-28611] Add Transformer for ElementwiseProduct

weibozhao:
Add Transformer for ElementwiseProduct

## Brief change log
Add Transformer for ElementwiseProduct in Flink ML (Java & Python).
Add unit test for ElementwiseProduct (Java & Python)
Add example code for ElementwiseProduct (Java&Python)

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.param;
+
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vector;
+
+import java.util.List;
+import java.util.Map;
+
+/** Class for the Vector parameter. */
+public class VectorParam extends Param<Vector> {
+
+    public VectorParam(
+            String name,
+            String description,
+            Vector defaultValue,
+            ParamValidator<Vector> validator) {
+        super(name, Vector.class, description, defaultValue, validator);
+    }
+
+    public VectorParam(String name, String description, Vector defaultValue) {
+        this(name, description, defaultValue, ParamValidators.alwaysTrue());
+    }
+
+    @Override
+    public Vector jsonDecode(Object object) {
+        Map<String, Object> vecValues = (Map) object;
+        if (vecValues.size() == 1) {
+            List<Double> list = (List<Double>) vecValues.get("values");
+            double[] values = new double[list.size()];
+            for (int i = 0; i < values.length; ++i) {
+                values[i] = list.get(i);
+            }
+            return new DenseVector(values);
+        } else if (vecValues.size() == 3) {
+            List<Double> valuesList = (List<Double>) vecValues.get("values");
+            List<Integer> indicesList = (List<Integer>) vecValues.get("indices");
+            int n = (int) vecValues.get("n");
+            double[] values = new double[valuesList.size()];
+            int[] indices = new int[indicesList.size()];
+            for (int i = 0; i < values.length; ++i) {
+                values[i] = valuesList.get(i);
+                indices[i] = indicesList.get(i);
+            }
+            return new SparseVector(n, indices, values);
+        } else {
+            throw new UnsupportedOperationException("Vector parameter is invalid.");
+        }
+    }
+}
--- a/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/ElementwiseProductExample.java
+++ b/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/ElementwiseProductExample.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.examples.feature;
+
+import org.apache.flink.ml.feature.elementwiseproduct.ElementwiseProduct;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+
+/**
+ * Simple program that creates a ElementwiseProduct instance and uses it for feature engineering.
+ */
+public class ElementwiseProductExample {
+    public static void main(String[] args) {
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        StreamTableEnvironment tEnv = StreamTableEnvironment.create(env);
+
+        // Generates input data.
+        DataStream<Row> inputStream =
+                env.fromElements(
+                        Row.of(0, Vectors.dense(1.1, 3.2)), Row.of(1, Vectors.dense(2.1, 3.1)));
+
+        Table inputTable = tEnv.fromDataStream(inputStream).as("id", "vec");
+
+        // Creates an ElementwiseProduct object and initializes its parameters.
+        ElementwiseProduct elementwiseProduct =
+                new ElementwiseProduct()
+                        .setInputCol("vec")
+                        .setOutputCol("outputVec")
+                        .setScalingVec(Vectors.dense(1.1, 1.1));
+
+        // Transforms input data.
+        Table outputTable = elementwiseProduct.transform(inputTable)[0];
+
+        // Extracts and displays the results.
+        for (CloseableIterator<Row> it = outputTable.execute().collect(); it.hasNext(); ) {
+            Row row = it.next();
+            Vector inputValue = (Vector) row.getField(elementwiseProduct.getInputCol());
+            Vector outputValue = (Vector) row.getField(elementwiseProduct.getOutputCol());
+            System.out.printf("Input Value: %s \tOutput Value: %s\n", inputValue, outputValue);
+        }
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.elementwiseproduct;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.BLAS;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.typeinfo.VectorTypeInfo;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A Transformer that multiplies each input vector with a given scaling vector using Hadamard
+ * product.
+ *
+ * <p>If the size of the input vector does not equal the size of the scaling vector, the transformer
+ * will throw {@link IllegalArgumentException}.
+ */
+public class ElementwiseProduct
+        implements Transformer<ElementwiseProduct>, ElementwiseProductParams<ElementwiseProduct> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public ElementwiseProduct() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), VectorTypeInfo.INSTANCE),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                new ElementwiseProductFunction(getInputCol(), getScalingVec()),
+                                outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    private static class ElementwiseProductFunction implements MapFunction<Row, Row> {
+        private final String inputCol;
+        private final Vector scalingVec;
+
+        public ElementwiseProductFunction(String inputCol, Vector scalingVec) {
+            this.inputCol = inputCol;
+            this.scalingVec = scalingVec;
+        }
+
+        @Override
+        public Row map(Row value) {
+            Vector inputVec = value.getFieldAs(inputCol);
+            if (inputVec != null) {
+                if (scalingVec.size() != inputVec.size()) {
+                    throw new IllegalArgumentException(
+                            "The scaling vector size is "
+                                    + scalingVec.size()
+                                    + ", which is not equal input vector size("
+                                    + inputVec.size()
+                                    + ").");
+                }
+                Vector retVec = inputVec.clone();
+                BLAS.hDot(scalingVec, retVec);
+                return Row.join(value, Row.of(retVec));
+            } else {
+                return Row.join(value, Row.of((Object) null));
+            }
+        }
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static ElementwiseProduct load(StreamTableEnvironment env, String path)
+            throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+}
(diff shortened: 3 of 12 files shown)


zhipeng93:
Thanks for the PR. Could you please add a JIRA ticket for this? Also for the title of the PR, could you follow the existing convention, i.e., [FLINK-XXX] Add Transformer for ...

flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vector;
+
+import java.util.List;
+import java.util.Map;
+
+/** Class for the Vector parameter. */
+public class VectorParam extends Param<Vector> {
```
yunfengzhou-hub, line 29: Could you please add corresponding tests in `StageTest` to verify this class?

flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
```diff
@@ -0,0 +1,68 @@
...
+            int n = (int) vecValues.get("n");
+            double[] values = new double[valuesList.size()];
+            int[] indices = new int[indicesList.size()];
+            for (int i = 0; i < values.length; ++i) {
+                values[i] = valuesList.get(i);
+                indices[i] = indicesList.get(i);
+            }
+            return new SparseVector(n, indices, values);
+        } else {
+            throw new RuntimeException("Parameter scalingVec is invalid.");
```
yunfengzhou-hub, line 65: Let's throw `IllegalxxxException` or `UnsupportedxxxException` here.

flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
```diff
@@ -0,0 +1,68 @@
...
+            double[] values = new double[list.size()];
+            for (int i = 0; i < values.length; ++i) {
+                values[i] = list.get(i);
+            }
+            return new DenseVector(values);
+        } else if (vecValues.size() == 3) {
+            List<Double> valuesList = (List<Double>) vecValues.get("values");
+            List<Integer> indicesList = (List<Integer>) vecValues.get("indices");
+            int n = (int) vecValues.get("n");
+            double[] values = new double[valuesList.size()];
```
yunfengzhou-hub, line 57: nit: `List.toArray()` can simplify the logic here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProductParams.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.apache.flink.ml.common.param.HasInputCol;
+import org.apache.flink.ml.common.param.HasOutputCol;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.VectorParam;
+
+/**
+ * Params of ElementwiseProduct.
```
yunfengzhou-hub, line 29: nit: `{@link }`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
```diff
@@ -0,0 +1,121 @@
...
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * ElementwiseProduct is a transformer that multiplies each input vector by a given scaling vector,
+ * using element-wise multiplication.
```
yunfengzhou-hub, line 47: `Hadamard product` seems to be a more precise description.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProductParams.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface ElementwiseProductParams<T> extends HasInputCol<T>, HasOutputCol<T> {
+
+    Param<Vector> SCALING_VEC =
+            new VectorParam(
+                    "scalingVec",
+                    "the vector to multiply with input vectors.",
```
yunfengzhou-hub, line 38: `Hadamard product` seems to be a more precise description.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
```diff
@@ -0,0 +1,121 @@
...
+            this.scalingVec = scalingVec.toDense();
+        }
+
+        @Override
+        public Row map(Row value) {
+            Vector inputVec = value.getFieldAs(inputCol);
+            if (null == inputVec) {
+                return null;
+            }
+            if (inputVec instanceof DenseVector) {
```
yunfengzhou-hub, line 89: We can reuse `BLAS` operations here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
```diff
@@ -0,0 +1,121 @@
...
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), VectorTypeInfo.INSTANCE),
+                        ArrayUtils.addAll(inputTypeInfo.getFieldNames(), getOutputCol()));
+        DataStream<Row> output =
+                tEnv.toDataStream(inputs[0])
+                        .map(new Elementwise(getInputCol(), getScalingVec()), outputTypeInfo);
+        Table outputTable = tEnv.fromDataStream(output);
+        return new Table[] {outputTable};
+    }
+
+    private static class Elementwise implements MapFunction<Row, Row> {
```
yunfengzhou-hub, line 74: nit :`ElementwiseProductFunction`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
yunfengzhou-hub, line 47: Let's add more descriptions and test cases for the following situations:
- the input vector is null
- the size of the input vector does not match the scaling vector

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
```diff
@@ -0,0 +1,121 @@
...
+        private final String inputCol;
+        private final DenseVector scalingVec;
+
+        public Elementwise(String inputCol, Vector scalingVec) {
+            this.inputCol = inputCol;
+            this.scalingVec = scalingVec.toDense();
```
yunfengzhou-hub, line 80: This conversion seems unnecessary and might cause performance issues.

flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
yunfengzhou-hub, line 65: the exception message here seems improper.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
```diff
@@ -0,0 +1,146 @@
...
+                            Vectors.dense(2.1, 3.1),
+                            Vectors.sparse(5, new int[] {3}, new double[] {1.0})),
+                    Row.of(
+                            1,
+                            Vectors.dense(1.1, 3.3),
+                            Vectors.sparse(
+                                    5, new int[] {4, 2, 3, 1}, new double[] {4.0, 2.0, 3.0, 1.0})));
+
+    private static final DenseVector EXPECTED_OUTPUT_DATA_1 =
+            Vectors.dense(2.3100000000000005, 3.4100000000000006);
```
yunfengzhou-hub, line 64: It might be better to use `2.31` here and set a tolerance threshold to ignore the `0.0000000000000005` difference.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
```diff
@@ -0,0 +1,146 @@
...
+        for (Row result : results) {
+            if (result.getField(0) == (Object) 0) {
+                assertEquals(
+                        isSparse ? EXPECTED_OUTPUT_DATA_3 : EXPECTED_OUTPUT_DATA_1,
+                        result.getField(outputCol));
+            } else if (result.getField(0) == (Object) 1) {
+                assertEquals(
+                        isSparse ? EXPECTED_OUTPUT_DATA_4 : EXPECTED_OUTPUT_DATA_2,
+                        result.getField(outputCol));
+            }
```
yunfengzhou-hub, line 99: nit: Let's add an `else` here to handle unexpected output results.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.commons.collections.IteratorUtils;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/** Tests ElementwiseProduct. */
```
yunfengzhou-hub, line 45: nit: `{@link }`

flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/ElementwiseProductExample.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+
+/**
+ * Simple program that creates a ElementwiseProduct instance and uses it for feature engineering.
```
yunfengzhou-hub, line 32: nit: "an ElementwiseProduct instance". Same for comments below.

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/param/VectorParam.java
weibozhao: List<Double> toArray is Double[], but we need double[] in the constructor of SparseVector. Then I use for loop instead of List.toArray

flink-ml-python/pyflink/ml/lib/feature/elementwiseproduct.py
```diff
@@ -0,0 +1,73 @@
...
+    HasInputCol,
+    HasOutputCol
+):
+    """
+    Params for :class:`ElementwiseProduct`.
+    """
+
+    SCALING_VEC: Param[Vector] = VectorParam(
+        "scaling_vec",
+        "The scaling vector of the product op.",
```
yunfengzhou-hub, line 37: Let's keep the description the same as that in Java.

flink-ml-python/pyflink/ml/lib/feature/elementwiseproduct.py
```diff
@@ -0,0 +1,73 @@
...
+        return self.get_scaling_vec()
+
+
+class ElementwiseProduct(JavaFeatureTransformer, _ElementwiseProductParams):
+    """
+    A transformer that combines a given list of input columns into a vector column. Types of
+    input columns must be either vector or numerical value.
+
+    The `keep` option of :class:HasHandleInvalid means that we output bad rows with output column
+    set to null.
```
yunfengzhou-hub, line 61: Let's keep the description the same as that in Java.

flink-ml-python/pyflink/ml/lib/feature/tests/test_elementwiseproduct.py
```diff
@@ -0,0 +1,76 @@
...
+                 Vectors.dense(2.1, 3.1)),
+                (1,
+                 Vectors.dense(1.1, 3.3)),
+            ],
+                type_info=Types.ROW_NAMED(
+                    ['id', 'vec'],
+                    [Types.INT(), DenseVectorTypeInfo()])))
+
+        self.expected_output_data_1 = Vectors.dense(2.3100000000000005, 3.4100000000000006)
+        self.expected_output_data_2 = Vectors.dense(1.2100000000000002, 3.63)
```
yunfengzhou-hub, line 42: Let's apply the changes in Java to Python test cases as well.

flink-ml-python/pyflink/ml/lib/feature/tests/test_elementwiseproduct.py
```diff
@@ -0,0 +1,76 @@
...
+
+        path = os.path.join(self.temp_dir, 'test_save_load_transform_elementwise_product')
+        elementwise_product.save(path)
+        elementwise_product = ElementwiseProduct.load(self.t_env, path)
+
+        output_table = elementwise_product.transform(self.input_data_table)[0]
+        actual_outputs = [(result[0], result[2]) for result in
+                          self.t_env.to_data_stream(output_table).execute_and_collect()]
+
+        for actual_output in actual_outputs:
```
yunfengzhou-hub, line 72: Let's check the size of `actual_outputs` here.

flink-ml-python/pyflink/ml/core/param.py
```diff
@@ -333,6 +334,17 @@ def __init__(self, name: str, description: str, default_value: Optional[str],
         super(StringParam, self).__init__(name, str, "str", description, default_value, validator)
 
 
+class VectorParam(Param[Vector]):
```
yunfengzhou-hub, line 337: Let's add test cases for this parameter in python as well. Relevant test cases seem to be in `test_stage.py`.

flink-ml-python/pyflink/examples/ml/feature/elementwiseproduct_example.py
```diff
@@ -0,0 +1,64 @@
...
+input_data_table = t_env.from_data_stream(
+    env.from_collection([
+        (1, Vectors.dense(2.1, 3.1)),
+        (2, Vectors.dense(1.1, 3.3))
+    ],
+        type_info=Types.ROW_NAMED(
+            ['id', 'vec'],
+            [Types.INT(), DenseVectorTypeInfo()])))
+
+# create a elementwise product object and initialize its parameters
```
yunfengzhou-hub, line 50: nit: `an`.

yunfengzhou-hub (commented):

flink-ml-python/pyflink/ml/core/tests/test_stage.py
```diff
@@ -21,14 +21,16 @@
...
 from pyflink.ml.core.param import ParamValidators, Param, BooleanParam, IntParam, \
-    FloatParam, StringParam, IntArrayParam, FloatArrayParam, StringArrayParam
+    FloatParam, StringParam, VectorParam, IntArrayParam, FloatArrayParam, StringArrayParam
 from pyflink.ml.tests.test_utils import PyFlinkMLTestCase
 
 BOOLEAN_PARAM = BooleanParam("boolean_param", "Description", False)
 INT_PARAM = IntParam("int_param", "Description", 1, ParamValidators.lt(100))
 FLOAT_PARAM = FloatParam("float_param", "Description", 3.0, ParamValidators.lt(100))
 STRING_PARAM = StringParam('string_param', "Description", "5")
+VECTOR_PARAM = VectorParam('vector_param', "Description", Vectors.dense(1, 2, 3))
```
yunfengzhou-hub, line 33: It might be unable to verify the correctness of `VectorParam` by just creating the `VECTOR_PARAM`. Could you please refer to tests for other parameters, like `IntParam`, and add corresponding test cases about `VectorParam`?

yunfengzhou-hub (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
```diff
@@ -0,0 +1,207 @@
...
+                    assertArrayEquals(EXPECTED_OUTPUT_SPARSE_VEC_INDICES_1, sparseVector.indices);
+                    assertArrayEquals(
+                            EXPECTED_OUTPUT_SPARSE_VEC_VALUES_1, sparseVector.values, 1.0e-5);
+                } else {
+                    assertArrayEquals(
+                            EXPECTED_OUTPUT_DENSE_VEC_ARRAY_1,
+                            ((DenseVector) result.getField(outputCol)).values,
+                            1.0e-5);
+                }
+            } else if (result.getField(0) == (Object) 1) {
```
yunfengzhou-hub, line 111: Do you think the following code could be simpler? Take the sparse vector case as an example:
```java
    private static final List<Vector> EXPECTED_OUTPUT_SPARSE_VEC =
            Arrays.asList(
                    Vectors.sparse(5, new int[] {3}, new double[] {0.0}),
                    Vectors.sparse(5, new int[] {1, 2, 3, 4}, new double[] {1.1, 0.0, 0.0, 0.0}),
                    null
            );

...

    private void verifyOutputResult(Table output, String outputCol, boolean isSparse)
            throws Exception {
        List<Row> results = IteratorUtils.toList(output.execute().collect());
        results.sort(Comparator.comparingInt(x -> (Integer) x.getField(0)));
        List<Vector> vectors = results.stream().map(x -> (Vector) x.getField(outputCol)).collect(Collectors.toList());
        assertEquals(vectors, EXPECTED_OUTPUT_SPARSE_VEC);
    }
```

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
```diff
@@ -0,0 +1,207 @@
...
+    private static final double[] EXPECTED_OUTPUT_DENSE_VEC_ARRAY_2 = new double[] {1.21, 3.63};
+
+    private static final int EXPECTED_OUTPUT_SPARSE_VEC_SIZE_1 = 5;
+    private static final int[] EXPECTED_OUTPUT_SPARSE_VEC_INDICES_1 = new int[] {3};
+    private static final double[] EXPECTED_OUTPUT_SPARSE_VEC_VALUES_1 = new double[] {0.0};
+
+    private static final int EXPECTED_OUTPUT_SPARSE_VEC_SIZE_2 = 5;
+    private static final int[] EXPECTED_OUTPUT_SPARSE_VEC_INDICES_2 = new int[] {1, 2, 3, 4};
+    private static final double[] EXPECTED_OUTPUT_SPARSE_VEC_VALUES_2 =
+            new double[] {1.1, 0.0, 0.0, 0.0};
```
yunfengzhou-hub, line 77: It might be uncommon for sparse vectors to store zero values. Can we improve it to 
```java
    private static final int[] EXPECTED_OUTPUT_SPARSE_VEC_INDICES_2 = new int[] {1};
    private static final double[] EXPECTED_OUTPUT_SPARSE_VEC_VALUES_2 =
            new double[] {1.1};
```
by modifying `ElementwiseProduct` or `BLAS.hDot`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
```diff
@@ -0,0 +1,112 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An transformer that multiplies each input vector with a given scaling vector using Hadamard
+ * product.
+ *
+ * <p>If the input vector is null, then the transformer will return null. If the input vector size
+ * not equals scaling vector size then the transformer will throw IllegalArgumentException.
```
yunfengzhou-hub, line 49: nit: "If the size of the input vector does not equal the size of the scaling vector, the transformer will throw {@link IllegalArgumentException}`."

flink-ml-python/pyflink/ml/lib/feature/tests/test_elementwiseproduct.py
```diff
@@ -0,0 +1,83 @@
...
+        output_table = elementwise_product.transform(self.input_data_table)[0]
+        actual_outputs = [(result[0], result[2]) for result in
+                          self.t_env.to_data_stream(output_table).execute_and_collect()]
+
+        self.assertEqual(2, len(actual_outputs))
+        for actual_output in actual_outputs:
+            if actual_output[0] == 0:
+                self.assertAlmostEqual(self.expected_output_data_1.get(0),
+                                       actual_output[1].get(0), delta=1e-7)
+                self.assertAlmostEqual(self.expected_output_data_1.get(1),
```
yunfengzhou-hub, line 77: Can we compare the two vectors or two arrays here, instead of comparing values at each index? If it is difficult, let's at least use a for loop here, and also check for the size of the vectors.

flink-ml-python/pyflink/ml/core/param.py
```diff
@@ -86,6 +87,8 @@ def get_param_map(self) -> Dict['Param[Any]', Any]:
     @staticmethod
     def _is_compatible_type(param: 'Param[V]', value: V) -> bool:
         if value is not None and param.type != type(value):
+            if type(value).__name__ == 'DenseVector' or type(value).__name__ == 'SparseVector':
+                return issubclass(type(value), param.type)
```
yunfengzhou-hub, line 91: Could you please illustrate why we need to add this code here, and why we don't need to add corresponding codes in the Java implementation?

yunfengzhou-hub (commented):
Thanks for the update. Left some comments as below. The rest of this PR LGTM.

flink-ml-python/pyflink/ml/core/param.py
weibozhao: In the Java code, using `param.clazz.isAssignableFrom(value.getClass())` to check the compatible. 
But in the python code, type and class are different.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
weibozhao: If we use the above code, the epsilon cannot be added. When there exists a small error, just like 1.0e-6, the assert will fail. Assertion of double value must use epsilon.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ElementwiseProductTest.java
weibozhao: I think there is no need to do this. For doing this is too expensive, we need to do the below things:
1. Loop all the elements in the sparse vector and judge whether they equal zero(double) or not. If an element is 1.0e-12, It will give us some confusion.
2. New two arrays to store new indices and new values.

flink-ml-python/pyflink/ml/lib/feature/elementwiseproduct.py
```diff
@@ -0,0 +1,73 @@
...
+        return self.get(self.SCALING_VEC)
+
+    @property
+    def scaling_vec(self) -> Vector:
+        return self.get_scaling_vec()
+
+
+class ElementwiseProduct(JavaFeatureTransformer, _ElementwiseProductParams):
+    """
+    ElementwiseProduct is a transformer that multiplies each input vector with a
```
zhipeng93, line 57: nit: make the python doc consistent with java docs.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/elementwiseproduct/ElementwiseProduct.java
```diff
@@ -0,0 +1,112 @@
...
+
+        public ElementwiseProductFunction(String inputCol, Vector scalingVec) {
+            this.inputCol = inputCol;
+            this.scalingVec = scalingVec;
+        }
+
+        @Override
+        public Row map(Row value) {
+            Vector inputVec = value.getFieldAs(inputCol);
+            Vector retVec = (null != inputVec) ? inputVec.clone() : null;
```
zhipeng93, line 90: How about we check the size of the inputVec and the scaling vec before conducting `hdot`? Then we can throw an illegalArgumentException here and the exception seems more clear to me.

zhipeng93 (commented):

zhipeng93:
Thanks for the update. LGTM.

PR #137, opened 2022-07-26
[FLINK-28684] Fix OneHotEncoder NPE during checkpoint

yunfengzhou-hub:
This PR fixes the NullPointerException that might be generated during OneHotEncoder's checkpoint process.


zhipeng93:
Thanks for the PR. LGTM.

PR #139, opened 2022-08-09
[FLINK-28803] Add Transformer and Estimator for KBinsDiscretizer

zhipeng93:
- Add Transformer and Estimator for KBinsDiscretizer[1] in Flink ML. 

## Brief change log
- Added Transformer and Estimator for KBinsDiscretizer.
- Added java test/example  for Transformer and Estimator for KBinsDiscretizer.
- Added python source/test/example for KBinsDiscretizer.
- Comparing with Sklearn [1], we made the following changes:
     - We remove `encode` parameter since we can achieve the same goal with a onehot-encoder following KBinsDiscretizer.
     - We remove `random_state` since it is non-trivial to support reproduciable sampling in a distributed setting, while in sklearn (single-node setting), it is much easier. If the users ask for this feature, we can add it later.
     - We remove `dtype` since we can achieve the same goal with a simple type conversion following KBinsDiscretizer.
     - We support `sampling` for both `quantile` and `kmeans`, but sklearn supports `sampling` only for `quantile`. [2]
     - We hardcode the threshold for width of bins as zero, while Sklearn uses 1e-8. [3]
     - We do not support `numSamples` as None while sklearn supports. [4]

[1] https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.KBinsDiscretizer.html
[2] https://github.com/apache/flink-ml/pull/139#discussion_r943058613
[3] https://github.com/apache/flink-ml/pull/139#discussion_r943052952
[4] https://github.com/apache/flink-ml/pull/139#discussion_r943053798

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
@@ -1,0 +1,341 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.kbinsdiscretizer;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.functions.MapPartitionFunction;
+import org.apache.flink.ml.api.Estimator;
+import org.apache.flink.ml.common.datastream.DataStreamUtils;
+import org.apache.flink.ml.feature.minmaxscaler.MinMaxScaler.MinMaxReduceFunctionOperator;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.Preconditions;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An Estimator which implements discretization (also known as quantization or binning) to transform
+ * continuous features into discrete ones. The output values are in [0, numBins).
+ *
+ * <p>KBinsDiscretizer implements three different binning strategies, and it can be set by {@link
+ * KBinsDiscretizerParams#STRATEGY}. If the strategy is set as {@link KBinsDiscretizerParams#KMEANS}
+ * or {@link KBinsDiscretizerParams#QUANTILE}, users should further set {@link
+ * KBinsDiscretizerParams#SUB_SAMPLES} for better performance.
+ *
+ * <p>There are several corner cases for different inputs as listed below:
+ *
+ * <ul>
+ *   <li>When the input values of one column are all the same, then they should be mapped to the
+ *       same bin (i.e., the zero-th bin). Thus the corresponding bin edges are `{Double.MIN_VALUE,
+ *       Double.MAX_VALUE}`.
+ *   <li>When the number of distinct values of one column is less than the specified number of bins
+ *       and the {@link KBinsDiscretizerParams#STRATEGY} is set as {@link
+ *       KBinsDiscretizerParams#KMEANS}, we switch to {@link KBinsDiscretizerParams#UNIFORM}.
+ *   <li>When the width of one output bin is zero, i.e., the left edge equals to the right edge of
+ *       the bin, we remove it.
+ * </ul>
+ */
+public class KBinsDiscretizer
+        implements Estimator<KBinsDiscretizer, KBinsDiscretizerModel>,
+                KBinsDiscretizerParams<KBinsDiscretizer> {
+    private static final Logger LOG = LoggerFactory.getLogger(KBinsDiscretizer.class);
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public KBinsDiscretizer() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public KBinsDiscretizerModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+
+        String inputCol = getInputCol();
+        String strategy = getStrategy();
+        int numBins = getNumBins();
+
+        DataStream<DenseVector> inputData =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                (MapFunction<Row, DenseVector>)
+                                        value -> ((Vector) value.getField(inputCol)).toDense());
+
+        DataStream<DenseVector> preprocessedData;
+        if (strategy.equals(UNIFORM)) {
+            preprocessedData =
+                    inputData
+                            .transform(
+                                    "reduceInEachPartition",
+                                    inputData.getType(),
+                                    new MinMaxReduceFunctionOperator())
+                            .transform(
+                                    "reduceInFinalPartition",
+                                    inputData.getType(),
+                                    new MinMaxReduceFunctionOperator())
+                            .setParallelism(1);
+        } else {
+            preprocessedData =
+                    DataStreamUtils.sample(
+                            inputData, getSubSamples(), getClass().getName().hashCode());
+        }
+
+        DataStream<KBinsDiscretizerModelData> modelData =
+                DataStreamUtils.mapPartition(
+                        preprocessedData,
+                        new MapPartitionFunction<DenseVector, KBinsDiscretizerModelData>() {
+                            @Override
+                            public void mapPartition(
+                                    Iterable<DenseVector> iterable,
+                                    Collector<KBinsDiscretizerModelData> collector) {
+                                List<DenseVector> list = new ArrayList<>();
+                                iterable.iterator().forEachRemaining(list::add);
+
+                                if (list.size() == 0) {
+                                    throw new RuntimeException("The training set is empty.");
+                                }
+
+                                double[][] binEdges;
+                                switch (strategy) {
+                                    case UNIFORM:
+                                        binEdges = findBinEdgesWithUniformStrategy(list, numBins);
+                                        break;
+                                    case QUANTILE:
+                                        binEdges = findBinEdgesWithQuantileStrategy(list, numBins);
+                                        break;
+                                    case KMEANS:
+                                        binEdges = findBinEdgesWithKMeansStrategy(list, numBins);
+                                        break;
+                                    default:
+                                        throw new UnsupportedOperationException(
+                                                "Unsupported "
+                                                        + STRATEGY
+                                                        + " type: "
+                                                        + strategy
+                                                        + ".");
+                                }
+
+                                collector.collect(new KBinsDiscretizerModelData(binEdges));
+                            }
+                        });
+        modelData.getTransformation().setParallelism(1);
+
+        KBinsDiscretizerModel model =
+                new KBinsDiscretizerModel().setModelData(tEnv.fromDataStream(modelData));
+        ReadWriteUtils.updateExistingParams(model, getParamMap());
+        return model;
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static KBinsDiscretizer load(StreamTableEnvironment tEnv, String path)
+            throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    private static double[][] findBinEdgesWithUniformStrategy(
+            List<DenseVector> input, int numBins) {
+        DenseVector minVector = input.get(0);
+        DenseVector maxVector = input.get(1);
+        int numColumns = minVector.size();
+        double[][] binEdges = new double[numColumns][];
+
+        for (int columnId = 0; columnId < numColumns; columnId++) {
+            double min = minVector.get(columnId);
+            double max = maxVector.get(columnId);
+            if (min == max) {
+                LOG.warn("Feature " + columnId + " is constant and the output will all be zero.");
+                binEdges[columnId] = new double[] {Double.MIN_VALUE, Double.MAX_VALUE};
+            } else {
+                double width = (max - min) / numBins;
+                binEdges[columnId] = new double[numBins + 1];
+                binEdges[columnId][0] = min;
+                for (int edgeId = 1; edgeId < numBins + 1; edgeId++) {
+                    binEdges[columnId][edgeId] = binEdges[columnId][edgeId - 1] + width;
+                }
+            }
+        }
+        return binEdges;
+    }
+
+    private static double[][] findBinEdgesWithQuantileStrategy(
+            List<DenseVector> input, int numBins) {
+        int numColumns = input.get(0).size();
+        int numData = input.size();
+        double[][] binEdges = new double[numColumns][];
+        double[] features = new double[numData];
+
+        for (int columnId = 0; columnId < numColumns; columnId++) {
+            for (int i = 0; i < numData; i++) {
+                features[i] = input.get(i).get(columnId);
+            }
+            Arrays.sort(features);
+
+            if (features[0] == features[numData - 1]) {
+                LOG.warn("Feature " + columnId + " is constant and the output will all be zero.");
+                binEdges[columnId] = new double[] {Double.MIN_VALUE, Double.MAX_VALUE};
+            } else {
+                double width = 1.0 * features.length / numBins;
+                double[] tempBinEdges = new double[numBins + 1];
+
+                for (int binEdgeId = 0; binEdgeId < numBins; binEdgeId++) {
+                    tempBinEdges[binEdgeId] = features[(int) (binEdgeId * width)];
+                }
+                tempBinEdges[numBins] = features[numData - 1];
+
+                // Removes bins that are empty, i.e., the left edge equals to the right edge.
+                Set<Double> edges = new HashSet<>(numBins);
+                for (double edge : tempBinEdges) {
+                    edges.add(edge);
+                }
+
+                binEdges[columnId] = edges.stream().mapToDouble(Double::doubleValue).toArray();
+                Arrays.sort(binEdges[columnId]);
+            }
+        }
+        return binEdges;
+    }
+
+    private static double[][] findBinEdgesWithKMeansStrategy(List<DenseVector> input, int numBins) {
+        int numColumns = input.get(0).size();
+        int numData = input.size();
+        double[][] binEdges = new double[numColumns][numBins + 1];
+        double[] features = new double[numData];
+
+        double[] kMeansCentroids = new double[numBins];
+        double[] sumByCluster = new double[numBins];
+
+        for (int columnId = 0; columnId < numColumns; columnId++) {
+            for (int i = 0; i < numData; i++) {
+                features[i] = input.get(i).get(columnId);
+            }
+            Arrays.sort(features);
+
+            if (features[0] == features[numData - 1]) {
+                LOG.warn("Feature " + columnId + " is constant and the output will all be zero.");
+                binEdges[columnId] = new double[] {Double.MIN_VALUE, Double.MAX_VALUE};
+            } else {
+                // Checks whether there are more than {numBins} distinct feature values in each
+                // column.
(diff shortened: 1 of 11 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+/**
+ * An Estimator which implements discretization (also known as quantization or binning), which
+ * transforms continuous features into discrete ones. The output values are in [0, numBins).
+ *
+ * <p>KBinsDiscretizer implements three different binning strategies, and it can be set by {@link
+ * KBinsDiscretizerParams#STRATEGY}. If the strategy is set as {@link KBinsDiscretizerParams#KMEANS}
+ * or {@link KBinsDiscretizerParams#QUANTILE}, users should further set {@link
+ * KBinsDiscretizerParams#SUB_SAMPLES} for better performance.
+ *
+ * <p>There are several cornel cases for different inputs as listed below:
```
yunfengzhou-hub, line 60: nit: "corner"

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * <ul>
+ *   <li>When the input values of one column are all the same, then they should be mapped to the
+ *       same bin (i.e., the zero-th bin). Thus the corresponding bin edges are {Double.MIN_VALUE,
+ *       Double.MAX_VALUE}.
+ *   <li>When the number of distinct values of one column is less than the specified number of bins
+ *       and the {@link KBinsDiscretizerParams#STRATEGY} is set as {@link
+ *       KBinsDiscretizerParams#KMEANS}, we switch to {@link KBinsDiscretizerParams#UNIFORM}.
+ *   <li>When the width of one output bin is zero, i.e., the left edge equals to the right edge of
```
yunfengzhou-hub, line 69: SKlearn removes bins whose width is less than `1e-8`[1]. Shall we adopt this value as well, or introduce a `HasTol` parameter to allow users to specify the width of the bins to be ignored?

[1] https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.KBinsDiscretizer.html

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
```diff
@@ -0,0 +1,85 @@
...
+
+    Param<Integer> NUM_BINS =
+            new IntParam("numBins", "Number of bins to produce.", 5, ParamValidators.gtEq(2));
+
+    Param<Integer> SUB_SAMPLES =
+            new IntParam(
+                    "subSamples",
+                    "Maximum number of samples used to fit the model.",
+                    200000,
+                    ParamValidators.gt(0));
```
yunfengzhou-hub, line 60: sklearn allows this value to be set to `None`, in which case all the training samples are used when computing the quantiles that determine the binning thresholds. Shall we support this value as well?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.StringParam;
+
+/**
+ * Params for {@link KBinsDiscretizer}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface KBinsDiscretizerParams<T> extends KBinsDiscretizerModelParams<T> {
```
yunfengzhou-hub, line 31: Would it be better to inherit `HasSeed` in this class to provide the random seed for sub sampling?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerModel.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A Model which transforms continuous features into discrete features using the model data computed
+ * by {@link KBinsDiscretizer}.
+ *
+ * <p>A feature value {v} should be mapped to a bin with edges as {left, right} if {v} is in [left,
```
yunfengzhou-hub, line 51: Is usages like `{v}` a common way to represent variables? I have not seen other usages like this before.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+                                    "reduceInEachPartition",
+                                    inputData.getType(),
+                                    new MinMaxReduceFunctionOperator())
+                            .transform(
+                                    "reduceInFinalPartition",
+                                    inputData.getType(),
+                                    new MinMaxReduceFunctionOperator())
+                            .setParallelism(1);
+        } else {
+            preprocessedData = DataStreamUtils.sample(inputData, getSubSamples(), 2022L);
```
yunfengzhou-hub, line 113: According to sklearn's document, the subsample parameter is only used when `strategy="quantile"`[1]. Here, it seems that the sampling is also applied when `strategy="kmeans"`.

[1] https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.KBinsDiscretizer.html

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+
+            if (features[0] == features[numData - 1]) {
+                LOG.warn("Feature " + columnId + " is constant and the output will all be zero.");
+                binEdges[columnId] = new double[] {Double.MIN_VALUE, Double.MAX_VALUE};
+            } else {
+                // Checks whether there are more than {numBins} distinct feature values in each
+                // column.
+                // If the number of distinct values is less than {numBins + 1}, then we do not need
+                // to conduct KMeans. Instead, we switch to use {@link
+                // KBinsDiscretizerParams#UNIFORM} for bining.
```
yunfengzhou-hub, line 263: nit: "binning".

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+                    // Conducts KMeans here.
+                    double width = 1.0 * features.length / numBins;
+                    for (int clusterId = 0; clusterId < numBins; clusterId++) {
+                        kMeansCentroids[clusterId] = features[(int) (clusterId * width)];
+                    }
+
+                    // Default values for KMeans following sklearn:
+                    // https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.KBinsDiscretizer.html#sklearn-preprocessing-kbinsdiscretizer.
+                    final double tolerance = 1e-4;
+                    final int maxIterations = 300;
```
yunfengzhou-hub, line 291: I was unable to find information like `maxIterations = 300` in the link provided above. Do you know where I can find them?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+                    double max = features[features.length - 1];
+                    double width = (max - min) / numBins;
+                    binEdges[columnId] = new double[numBins + 1];
+                    binEdges[columnId][0] = min;
+                    for (int edgeId = 1; edgeId < numBins + 1; edgeId++) {
+                        binEdges[columnId][edgeId] = binEdges[columnId][edgeId - 1] + width;
+                    }
+                    continue;
+                } else {
+                    // Conducts KMeans here.
```
yunfengzhou-hub, line 282: Would it be better to reuse the logic in the kmeans or online kmeans stages?

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
zhipeng93: Removing bins whose width is less than `1e-8` is ambiguous to users --- Consider the following case:
Suppose we have four consecutive bins with width {6e-9, 6e-9, 6e-9, 6e-9} and the minimum bin width is `1e-8`, we may get different results: {1.2e-8, 1.2e-8} and {2.4e-8}, both of which are reasonable.

Setting minimum bin width as zero does not incur the above problem. If the users found the bin is too small, they may decrease number of bins.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
zhipeng93: Setting the value as `None` is useless here. If users want to use all of the samples, they can set the value as `Integer.MAX_VALUE`.

In our implementation, if `numSamples` is greater than number of input data, all of the data are used in quantile/kmeans.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
zhipeng93: Setting `HasSeed` has two possible indications:
- Using same seed can have reproduciable outputs
- Using different seed can have different outputs.

However, our implementation does not provide reproduciable outputs due to the complexity of reproduciable distributed sampling --- We always provide different outputs in each run. As a result, it seems meaningless to provide `HasSeed` here.

However, we can add `HasSeed` later if we can provide reproduciable distributed sampling.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
zhipeng93: It uses the default setting in Kmeans [1][2].

[1] https://github.com/scikit-learn/scikit-learn/blob/17df37aee774720212c27dbc34e6f1feef0e2482/sklearn/preprocessing/_discretization.py#L279
[2] https://github.com/scikit-learn/scikit-learn/blob/17df37aee774720212c27dbc34e6f1feef0e2482/sklearn/cluster/_kmeans.py#L1162

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
zhipeng93: As I understand, `quantitle` and `kmeans` are similar in the following aspects:
- They are both used to estimate/compute the bin edges.
- For accurate implementation, they both need to access/store all of the input data.

Now that sampling is used for `quantitle` to accelerate the process of computing bins, with a compromise on precision. Then it seems unclear why do we do different stuff to `kmeans`. 

If we do not support sampling for `kmeans`, it may only work on small datasets. Also, users can also set `numSamples` as `Integer.MAX_VALUE` to do `kmeans` with no sampling.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
zhipeng93: It may no be that efficient to reuse the logic in `KMeans` class:
- We do kmeans for each input column here, i.e., the input feature is simply a double, not a vector.
- Reusing `Kmeans` class means we introduce extra flink operators, and make the job more complex. Here a toy code for one-dimension kmeans is enough.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerModel.java
zhipeng93: Thanks for pointing this out. I have replaced `{}` with `\`\`` following [1]

[1] https://github.com/apache/spark/blob/e17d8ecabcad6e84428752b977120ff355a4007a/mllib/src/main/scala/org/apac

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
yunfengzhou-hub: Got it. I agree with it that we may try to support this parameter in future if we need it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
yunfengzhou-hub: Got it. I agree with it that we may try to support this parameter in the future if we need it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
yunfengzhou-hub: Got it. I agree with it that we may try to support this parameter in the future if we need it.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
yunfengzhou-hub: Sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
yunfengzhou-hub, line 60: nit: Shall we add a check that `subSamples >= 2` or `subSamples >= numBins`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+                    double width = 1.0 * features.length / numBins;
+                    for (int clusterId = 0; clusterId < numBins; clusterId++) {
+                        kMeansCentroids[clusterId] = features[(int) (clusterId * width)];
+                    }
+
+                    // Default values for KMeans following sklearn:
+                    // https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.KBinsDiscretizer.html#sklearn-preprocessing-kbinsdiscretizer.
```
yunfengzhou-hub, line 289: Do you think it would be better to remove comments like this? This information is useful for reviewers, that the implementation is following an existing practice. But I wonder if this information will also be necessary for future readers, or if this information will be outdated when sklearn got updated.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/KBinsDiscretizerTest.java
```diff
@@ -0,0 +1,296 @@
...
+        env = StreamExecutionEnvironment.getExecutionEnvironment(config);
+        env.setParallelism(4);
+        env.enableCheckpointing(100);
+        env.setRestartStrategy(RestartStrategies.noRestart());
+        tEnv = StreamTableEnvironment.create(env);
+        trainTable = tEnv.fromDataStream(env.fromCollection(TRAIN_INPUT)).as("input");
+        testTable = tEnv.fromDataStream(env.fromCollection(TEST_INPUT)).as("input");
+    }
+
+    @SuppressWarnings("all")
```
yunfengzhou-hub, line 138: nit: it might be better to replace this with the detailed warnings you want to suppress.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/KBinsDiscretizerTest.java
```diff
@@ -0,0 +1,296 @@
...
+    private void verifyPredictionResult(
+            List<Row> expectedOutput, Table output, String predictionCol) throws Exception {
+        List<Row> collectedResult =
+                IteratorUtils.toList(
+                        tEnv.toDataStream(output.select($(predictionCol))).executeAndCollect());
+        compareResultCollections(
+                expectedOutput,
+                collectedResult,
+                (o1, o2) -> {
```
yunfengzhou-hub, line 147: There is a method called `TestUtils::compare`, which might be useful here. And #142 has been generalizing the method from `DenseVector` to `Vector`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/KBinsDiscretizerTest.java
```diff
@@ -0,0 +1,296 @@
...
+
+    @Test
+    public void testFitOnEmptyData() {
+        Table emptyTable =
+                tEnv.fromDataStream(env.fromCollection(TRAIN_INPUT).filter(x -> x.getArity() == 0))
+                        .as("input");
+        KBinsDiscretizerModel model = new KBinsDiscretizer().fit(emptyTable);
+        Table modelDataTable = model.getModelData()[0];
+        try {
+            IteratorUtils.toList(
```
yunfengzhou-hub, line 288: nit: `modelDataTable.execute().collect().next();` might be simpler here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,340 @@
...
+                                    "reduceInEachPartition",
+                                    inputData.getType(),
+                                    new MinMaxReduceFunctionOperator())
+                            .transform(
+                                    "reduceInFinalPartition",
+                                    inputData.getType(),
+                                    new MinMaxReduceFunctionOperator())
+                            .setParallelism(1);
+        } else {
+            preprocessedData = DataStreamUtils.sample(inputData, getSubSamples(), 2022L);
```
yunfengzhou-hub, line 113: Would it be better to change the random seed from `2022L` to `getClass().getName().hashCode()`? If one day we would support the random seed parameter, this would be its default value.

yunfengzhou-hub (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/KBinsDiscretizerTest.java
zhipeng93: Thanks for pointing this out. I have updated it with `@SuppressWarnings("unchecked, ConstantConditions")`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
zhipeng93: I do not have a strong preference here. I have removed these comments since you pointed this out.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizerParams.java
zhipeng93: I agree that `subSamples < 2` is not a valid use case and I have udpated the code.

If `subSamples < numBins`, it should still work in `quantile` and `kmeans` case.

zhipeng93:
@yunfengzhou-hub Thanks for the review. I have addressed your comments in the PR. 
@lindong28 Do you want to take a look?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,341 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An Estimator which implements discretization (also known as quantization or binning), which
```
lindong28, line 52: nits: maybe rephrase this sentence like this:

```
 * An Estimator which implements discretization (also known as quantization or binning) to
 * transform continuous features into discrete ones. The output values are in [0, numBins).
```

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,341 @@
...
+            } else {
+                double width = 1.0 * features.length / numBins;
+                double[] tempBinEdges = new double[numBins + 1];
+
+                for (int binEdgeId = 0; binEdgeId < numBins; binEdgeId++) {
+                    tempBinEdges[binEdgeId] = features[(int) (binEdgeId * width)];
+                }
+                tempBinEdges[numBins] = features[numData - 1];
+
+                // Removes bins who are empty, i.e., the left edge equals to the right edge.
```
lindong28, line 229: nits: `who` -> `that`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -0,0 +1,341 @@
...
+            Arrays.sort(features);
+
+            if (features[0] == features[numData - 1]) {
+                LOG.warn("Feature " + columnId + " is constant and the output will all be zero.");
+                binEdges[columnId] = new double[] {Double.MIN_VALUE, Double.MAX_VALUE};
+            } else {
+                // Checks whether there are more than {numBins} distinct feature values in each
+                // column.
+                // If the number of distinct values is less than {numBins + 1}, then we do not need
+                // to conduct KMeans. Instead, we switch to use {@link
```
lindong28, line 264: nits: `use` -> `using`

lindong28 (commented):
Thanks @zhipeng93 for the PR and thanks @yunfengzhou-hub for the review!

The PR looks pretty good. Left just minor comments. @zhipeng93 Please feel free to merge the PR.

zhipeng93:
Thanks for the review @lindong28. Merging now.

PR #141, opened 2022-08-09
[FLINK-28805] Add Transformer for HashingTF

zhipeng93:
- Add Transformer for HashingTF[1] in Flink ML. 

## Brief change log
- Added Transformer for HashingTF.
- Added java test/example  for Transformer for HashingTF.
- Added python source/test/example for HashingTF.
- The public interface is consistent with HashingTF in Spark ML [1].

[1] https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.ml.feature.HashingTF.html

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
@@ -1,0 +1,198 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.hashingtf;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.ml.linalg.typeinfo.SparseVectorTypeInfo;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.catalog.ResolvedSchema;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.shaded.guava30.com.google.common.hash.Hashing.murmur3_32;
+
+/**
+ * A Transformer that maps a sequence of terms(strings, numbers, booleans) to a sparse vector with a
+ * specified dimension using the hashing trick.
+ *
+ * <p>If multiple features are projected into the same column, the output values are accumulated by
+ * default. Users could also enforce all non-zero output values as 1 by setting {@link
+ * HashingTFParams#BINARY} as true.
+ *
+ * <p>For the hashing trick, see https://en.wikipedia.org/wiki/Feature_hashing for details.
+ */
+public class HashingTF implements Transformer<HashingTF>, HashingTFParams<HashingTF> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    private static final org.apache.flink.shaded.guava30.com.google.common.hash.HashFunction
+            HASH_FUNC = murmur3_32(0);
+
+    public HashingTF() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+
+        ResolvedSchema tableSchema = inputs[0].getResolvedSchema();
+
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(tableSchema);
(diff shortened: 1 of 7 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
```diff
@@ -0,0 +1,193 @@
...
+        public Row map(Row row) throws Exception {
+            Object inputObj = row.getField(inputCol);
+
+            Iterable<Object> inputList;
+            if (inputObj.getClass().isArray()) {
+                inputList = Arrays.asList((Object[]) inputObj);
+            } else if (inputObj instanceof Iterable) {
+                inputList = (Iterable<Object>) inputObj;
+            } else {
+                throw new IllegalArgumentException(
```
yunfengzhou-hub, line 124: It might be better to do this check in `transform()` method and throw exceptions as early as during compilation.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
```diff
@@ -0,0 +1,193 @@
...
+ * default. Users could also enforce all non-zero output values as 1 by setting {@link
+ * HashingTFParams#BINARY} as true.
+ *
+ * <p>For the hashing trick, see https://en.wikipedia.org/wiki/Feature_hashing for details.
+ */
+public class HashingTF implements Transformer<HashingTF>, HashingTFParams<HashingTF> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    private static final org.apache.flink.shaded.guava30.com.google.common.hash.HashFunction
+            HASH_FUNC = murmur3_32(0);
```
yunfengzhou-hub, line 61: Spark seems to use 42 as the hash seed[1]. Do you know why Spark chooses this value?

If this value in Spark was just a personal choice, I agree that we can use 0 for the seed in our implementation.

[1] https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/mllib/feature/HashingTF.scala#L143

flink-ml-python/pyflink/ml/lib/feature/tests/test_hashingtf.py
```diff
@@ -0,0 +1,115 @@
...
+            self.assertEqual(expected_output[i], predicted_result[i])
+
+    def test_transform(self):
+        hashing_tf = HashingTF()
+
+        # Tests non-binary.
+        output = hashing_tf.transform(self.input_data_table)[0]
+        self.verify_output_result(output, self.expected_output)
+
+        # Tests non-binary.
```
yunfengzhou-hub, line 104: nit: this should be binary.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
```diff
@@ -0,0 +1,193 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.shaded.guava30.com.google.common.hash.Hashing.murmur3_32;
+
+/**
+ * A Transformer that maps a sequence of terms(strings, numbers, booleans) to a sparse vector with a
```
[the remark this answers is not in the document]
zhipeng93: Good catch. The output is indeed sparse vector. I have updated the output type information as SparseVectorTypeInfo.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
zhipeng93: I think it is just a personal choice.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
zhipeng93: Unfortunately we cannot do this check during compilation. If the input is a list, the typeinformation turns to be `generciType(ArrayList)`, which is hard to throw exceptions during compilation process.

zhipeng93:
@yunfengzhou-hub Thanks for the review. The PR is updated.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
```diff
@@ -0,0 +1,195 @@
...
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+
+        ResolvedSchema tableSchema = inputs[0].getResolvedSchema();
+
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(tableSchema);
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(), SparseVectorTypeInfo.INSTANCE),
```
yunfengzhou-hub, line 79: Do you think we may support generating dense vectors from input data for performance improvement in the future? If so, I think `VectorTypeInfo` would be a better choice than `SparseVectorTypeInfo` here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
```diff
@@ -0,0 +1,195 @@
...
+                if (map.containsKey(index)) {
+                    if (!binary) {
+                        map.put(index, map.get(index) + 1);
+                    }
+                } else {
+                    map.put(index, 1);
+                }
+            }
+
+            // Converts to map to a sparse vector.
```
yunfengzhou-hub, line 144: nit: "from map to a sparse vector".

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
```diff
@@ -0,0 +1,195 @@
...
+
+            Iterable<Object> inputList;
+            if (inputObj.getClass().isArray()) {
+                inputList = Arrays.asList((Object[]) inputObj);
+            } else if (inputObj instanceof Iterable) {
+                inputList = (Iterable<Object>) inputObj;
+            } else {
+                throw new IllegalArgumentException(
+                        "The input format is not supported. "
+                                + "Supported options are Array and Iterable.");
```
yunfengzhou-hub, line 128: nit: It might be better to provide the column name and the column type that has violated the algorithm's requirement in this error message.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
zhipeng93: No, the output should always be a sparse vector.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
zhipeng93: emmm, what's wrong with the current comment?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/hashingtf/HashingTF.java
zhipeng93: Thanks for the comment. The coment is updated as:

`"Input format " + inputObj.getClass().getCanonicalName() + " is not supported for input column " + inputCol + ". Supported options are Array and Iterable.");`

yunfengzhou-hub (commented):
Thanks for the update. LGTM.

lindong28:
Thanks for the PR! And thanks @yunfengzhou-hub for the review. LGTM.

PR #142, opened 2022-08-12
[FLINK-28943] Add Transformer and Estimator for MaxAbsScaler

weibozhao:
Add Transformer and Estimator of MaxAbsScaler in FlinkML. 
## Brief change log
* Add MaxAbsScaler (Java & Python)
* Add MaxAbsScalerModel (Java & Python)
* Add unit test for MaxAbsScaler and MaxAbsScalerModel (Java & Python)
* Add example code for MaxAbsScaler and MaxAbsScalerModel (Java&Python)
Does this pull request introduce a new feature? (no)

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
@@ -1,0 +1,180 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.maxabsscaler;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
+import org.apache.flink.iteration.operator.OperatorStateUtils;
+import org.apache.flink.ml.api.Estimator;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.typeinfo.DenseVectorTypeInfo;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.runtime.state.StateSnapshotContext;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
+import org.apache.flink.streaming.api.operators.BoundedOneInput;
+import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An Estimator which implements the MaxAbsScaler algorithm. This algorithm rescales feature values
+ * to the range [-1, 1] by dividing through the largest maximum absolute value in each feature. It
+ * does not shift/center the data and thus does not destroy any sparsity.
+ */
+public class MaxAbsScaler
+        implements Estimator<MaxAbsScaler, MaxAbsScalerModel>, MaxAbsScalerParams<MaxAbsScaler> {
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public MaxAbsScaler() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    @Override
+    public MaxAbsScalerModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        final String inputCol = getInputCol();
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+
+        DataStream<Vector> inputData =
+                tEnv.toDataStream(inputs[0])
+                        .map(
+                                (MapFunction<Row, Vector>)
+                                        value -> ((Vector) value.getField(inputCol)));
+
+        DataStream<Vector> maxAbsValues =
+                inputData
+                        .transform(
+                                "reduceInEachPartition",
+                                inputData.getType(),
+                                new MaxAbsReduceFunctionOperator())
+                        .transform(
+                                "reduceInFinalPartition",
+                                inputData.getType(),
+                                new MaxAbsReduceFunctionOperator())
+                        .setParallelism(1);
+
+        DataStream<MaxAbsScalerModelData> modelData =
+                maxAbsValues.map(
+                        (MapFunction<Vector, MaxAbsScalerModelData>)
+                                vector -> new MaxAbsScalerModelData((DenseVector) vector));
+
+        MaxAbsScalerModel model =
+                new MaxAbsScalerModel().setModelData(tEnv.fromDataStream(modelData));
+        ReadWriteUtils.updateExistingParams(model, getParamMap());
+        return model;
+    }
+
+    /**
+     * A stream operator to compute the maximum absolute values in each partition of the input
+     * bounded data stream.
+     */
+    private static class MaxAbsReduceFunctionOperator extends AbstractStreamOperator<Vector>
+            implements OneInputStreamOperator<Vector, Vector>, BoundedOneInput {
+        private ListState<DenseVector> maxAbsState;
+        private DenseVector maxAbsVector;
+
+        @Override
+        public void endInput() {
+            if (maxAbsVector != null) {
+                output.collect(new StreamRecord<>(maxAbsVector));
+            }
+        }
+
+        @Override
+        public void processElement(StreamRecord<Vector> streamRecord) {
+            Vector currentValue = streamRecord.getValue();
+            if (currentValue == null) {
+                throw new RuntimeException("Input column data cannot be null.");
+            }
+
+            maxAbsVector =
+                    maxAbsVector == null ? new DenseVector(currentValue.size()) : maxAbsVector;
+            Preconditions.checkArgument(
+                    currentValue.size() == maxAbsVector.size(),
+                    "The training data should all have same dimensions.");
+
+            if (currentValue instanceof DenseVector) {
+                double[] values = ((DenseVector) currentValue).values;
+                for (int i = 0; i < currentValue.size(); ++i) {
+                    maxAbsVector.values[i] = Math.max(maxAbsVector.values[i], Math.abs(values[i]));
+                }
+            } else if (currentValue instanceof SparseVector) {
+                int[] indices = ((SparseVector) currentValue).indices;
+                double[] values = ((SparseVector) currentValue).values;
+
+                for (int i = 0; i < indices.length; ++i) {
+                    maxAbsVector.values[indices[i]] =
+                            Math.max(maxAbsVector.values[indices[i]], Math.abs(values[i]));
+                }
+            }
+        }
+
+        @Override
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            maxAbsState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxAbsState", DenseVectorTypeInfo.INSTANCE));
+
+            OperatorStateUtils.getUniqueElement(maxAbsState, "maxAbsState")
+                    .ifPresent(x -> maxAbsVector = x);
+        }
+
+        @Override
+        public void snapshotState(StateSnapshotContext context) throws Exception {
+            super.snapshotState(context);
+            maxAbsState.clear();
+            if (maxAbsVector != null) {
+                maxAbsState.add(maxAbsVector);
+            }
+        }
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+
+    public static MaxAbsScaler load(StreamTableEnvironment tEnv, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
@@ -16,16 +16,17 @@
  * limitations under the License.
  */
 
-package org.apache.flink.ml.feature.minmaxscaler;
+package org.apache.flink.ml.feature.maxabsscaler;
 
 import org.apache.flink.api.common.functions.RichMapFunction;
-import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
 import org.apache.flink.ml.api.Model;
 import org.apache.flink.ml.common.broadcast.BroadcastUtils;
 import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.BLAS;
 import org.apache.flink.ml.linalg.DenseVector;
 import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.typeinfo.VectorTypeInfo;
 import org.apache.flink.ml.param.Param;
 import org.apache.flink.ml.util.ParamUtils;
 import org.apache.flink.ml.util.ReadWriteUtils;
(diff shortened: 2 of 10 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+                inputData
+                        .transform(
+                                "reduceInEachPartition",
+                                inputData.getType(),
+                                new MaxAbsReduceFunctionOperator())
+                        .transform(
+                                "reduceInFinalPartition",
+                                inputData.getType(),
+                                new MaxAbsReduceFunctionOperator())
+                        .setParallelism(1);
```
yunfengzhou-hub, line 87: Would it be better to reuse `DataStreamUtils.reduce` for this part?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+        DataStream<MaxAbsScalerModelData> modelData =
+                DataStreamUtils.mapPartition(
```
yunfengzhou-hub, line 89: Given that the upstream operator would generate only one stream record, it seems unnecessary to use `DataStreamUtils.mapPartition` here. Maybe a simple `map` function is enough.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * A stream operator to compute the max values in each partition of the input bounded data
+     * stream.
+     */
+    private static class MaxAbsReduceFunctionOperator extends AbstractStreamOperator<Vector>
+            implements OneInputStreamOperator<Vector, Vector>, BoundedOneInput {
+        private ListState<DenseVector> maxState;
+        private DenseVector maxVector;
```
yunfengzhou-hub, line 113: If all input records are sparse vectors, this practice might waste much memory space. Is there any better way to improve this implementation?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+            maxState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxState", TypeInformation.of(DenseVector.class)));
+            OperatorStateUtils.getUniqueElement(maxState, "maxState").ifPresent(x -> maxVector = x);
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
```
yunfengzhou-hub, line 178: Some SuppressWarnings like this are redundant. Could you please check and remove them?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,162 @@
...
+            this.inputCol = inputCol;
+        }
+
+        @Override
+        public Row map(Row row) {
+            if (scaleVector == null) {
+                MaxAbsScalerModelData maxAbsScalerModelData =
+                        (MaxAbsScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
+                scaleVector = maxAbsScalerModelData.maxVector;
```
yunfengzhou-hub, line 146: Could you please add document, support and tests for cases in which `maxVector` contains 0 values, which means `1.0 / scaleVector.values[i]` is invalid?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,162 @@
...
+                for (int i = 0; i < scaleVector.size(); ++i) {
+                    if (scaleVector.values[i] != 0) {
+                        scaleVector.values[i] = 1.0 / Math.abs(scaleVector.values[i]);
```
yunfengzhou-hub, line 149: Spark's implementation is similar to `1.0 / scaleVector.values[i]`, instead of `1.0 / Math.abs(scaleVector.values[i])`. Would it be better to follow Spark's practice?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.apache.flink.test.util.TestBaseUtils.compareResultCollections;
+import static org.junit.Assert.assertEquals;
+
+/** Tests {@link MaxAbsScaler} and {@link MaxAbsScalerModel}. */
+public class MaxAbsScalerTest {
```
yunfengzhou-hub, line 55: Could you please add a test to verify the case when all values of input vectors are negative values? There might be a bug here?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
```diff
@@ -0,0 +1,271 @@
...
+        assertEquals(
+                Collections.singletonList("maxVector"),
+                modelData.getResolvedSchema().getColumnNames());
+        DataStream<Row> output = tEnv.toDataStream(modelData);
+        List<Row> modelRows = IteratorUtils.toList(output.executeAndCollect());
+        assertEquals(new DenseVector(new double[] {200.0, 400.0}), modelRows.get(0).getField(0));
+    }
+
+    @Test
+    public void testSetModelData() throws Exception {
```
yunfengzhou-hub, line 262: Could you please add a test case that sets a model data that contains zero and negative values, and make sure the resulting behavior is the same as that in Spark?

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
weibozhao: If maxVector contains zero values, that means all values in these features are zero. I have added a ut for this case.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
weibozhao: I have fixed this bug.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
weibozhao: I think `DataStreamUtils.reduce` is not suit here. When the input table only has one vector and this vector has some negative values, the output data will not be correct.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
weibozhao: I think this case will not happen.  If all input records are sparse vectors, the maxAbsVector need not be a sparse vector. If a lot of values in maxAbsVector are zeros, which means the input data has a lot of invalid features, we need to check the input data.

yunfengzhou-hub: Got it. I'm OK with the current implementation.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,162 @@
...
+
+        @Override
+        public Row map(Row row) {
+            if (scaleVector == null) {
+                MaxAbsScalerModelData maxAbsScalerModelData =
+                        (MaxAbsScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastKey).get(0);
+                scaleVector = maxAbsScalerModelData.maxVector;
+                for (int i = 0; i < scaleVector.size(); ++i) {
+                    if (scaleVector.values[i] != 0) {
```
yunfengzhou-hub, line 148: Spark would set the scale to 1 if a value in maxVector is 0. Would it be better to follow the existing practice in Spark?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,162 @@
...
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<Row> data = tEnv.toDataStream(inputs[0]);
+        DataStream<MaxAbsScalerModelData> maxAbsScalerModel =
+                MaxAbsScalerModelData.getModelDataStream(modelDataTable);
+        final String broadcastModelKey = "broadcastModelKey";
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(
+                                inputTypeInfo.getFieldTypes(), TypeInformation.of(Vector.class)),
```
yunfengzhou-hub, line 84: nit: VectorTypeInfo.INSTANCE

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,162 @@
...
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A Model which do a MaxAbsScaler operation using the model data computed by {@link MaxAbsScaler}.
```
yunfengzhou-hub, line 48: nit: "does". Same for `MinMaxScaler` and their python documents.

Or we can refer to the JavaDoc of `StandardScalerModel` and keep all the three scaler algorithms with the same JavaDoc template.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                } else if (currentValue instanceof SparseVector) {
+                    int[] indices = ((SparseVector) currentValue).indices;
+                    double[] values = ((SparseVector) currentValue).values;
+                    for (int i = 0; i < indices.length; ++i) {
+                        maxAbsVector.values[indices[i]] =
+                                Math.max(maxAbsVector.values[indices[i]], Math.abs(values[i]));
+                    }
+                } else {
+                    throw new RuntimeException(
```
yunfengzhou-hub, line 152: nit: Let's throw a more detailed exception here, like `IllegalArgumentException`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        @Override
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            maxAbsState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxState", TypeInformation.of(DenseVector.class)));
```
yunfengzhou-hub, line 165: nit: `DenseVectorTypeInfo`.

yunfengzhou-hub (commented):
Thanks for the update. Left two small comments. The rest of this PR LGTM.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+        public void processElement(StreamRecord<Vector> streamRecord) {
+            Vector currentValue = streamRecord.getValue();
+            if (currentValue == null) {
+                throw new RuntimeException("Input column data cannot be null.");
+            }
+
+            if (maxAbsVector == null) {
+                int vecSize = currentValue.size();
+                maxAbsVector = new DenseVector(vecSize);
+                if (currentValue instanceof DenseVector) {
```
zhipeng93, line 125: How about we simplify the code as:
```
maxAbsVector = maxAbsVector == null? new DenseVector(currentValue.size): maxAbsVector
```
and remove the code from Line#121 to Line#139?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+                } else if (currentValue instanceof SparseVector) {
+                    int[] indices = ((SparseVector) currentValue).indices;
+                    double[] values = ((SparseVector) currentValue).values;
+
+                    for (int i = 0; i < indices.length; ++i) {
+                        maxAbsVector.values[indices[i]] =
+                                Math.max(maxAbsVector.values[indices[i]], Math.abs(values[i]));
+                    }
+                } else {
+                    throw new IllegalArgumentException(
```
zhipeng93, line 159: nit: This line could be removed because the Vector could only be SparseVector or DenseVector for now.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+        @Override
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            maxAbsState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxState", DenseVectorTypeInfo.INSTANCE));
+
+            OperatorStateUtils.getUniqueElement(maxAbsState, "maxState")
```
zhipeng93, line 174: nit: maxAbsState

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        @Override
+        public void initializeState(StateInitializationContext context) throws Exception {
+            super.initializeState(context);
+            maxAbsState =
+                    context.getOperatorStateStore()
+                            .getListState(
+                                    new ListStateDescriptor<>(
+                                            "maxState", DenseVectorTypeInfo.INSTANCE));
```
zhipeng93, line 172: nit: maxAbsState

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,169 @@
...
+            throws IOException {
+        MaxAbsScalerModel model = ReadWriteUtils.loadStageParam(path);
+
+        Table modelDataTable =
+                ReadWriteUtils.loadModelData(
+                        tEnv, path, new MaxAbsScalerModelData.ModelDataDecoder());
+        return model.setModelData(modelDataTable);
+    }
+
+    /** This operator loads model data and predicts result. */
```
zhipeng93, line 131: nit: This operator --> This function.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScalerModel.java
```diff
@@ -0,0 +1,169 @@
...
+                for (int i = 0; i < scaleVector.size(); ++i) {
+                    if (scaleVector.values[i] != 0) {
+                        scaleVector.values[i] = 1.0 / scaleVector.values[i];
+                    } else {
+                        scaleVector.values[i] = 1.0;
+                    }
+                }
+            }
+
+            if (row.getFieldAs(inputCol) == null) {
```
zhipeng93, line 159: Shall we remove Line#159 to Line#161 and assume that the `inputCol` field of the row is never null? We need to make sure that the behavior of the existing implementations are consistent.

Please refer to MinMaxScalerModel, StringIndexerModel and other existing algorithms.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
```diff
@@ -0,0 +1,286 @@
...
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) output).getTableEnvironment();
+
+        DataStream<Vector> stream =
+                tEnv.toDataStream(output)
+                        .map((MapFunction<Row, Vector>) row -> row.getFieldAs(outputCol));
+
+        List<Vector> result = IteratorUtils.toList(stream.executeAndCollect());
+        if (isSparse) {
+            compareResultCollections(
```
zhipeng93, line 144: Shall we make expectedData an argument of `verifyPredictionResult` function?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
```diff
@@ -0,0 +1,286 @@
...
+                                Row.of(Vectors.dense(200, 400))));
+
+        Table trainDataWithInvalidData =
+                tEnv.fromDataStream(env.fromCollection(trainData)).as("input");
+
+        try {
+            MaxAbsScaler maxAbsScaler = new MaxAbsScaler();
+            MaxAbsScalerModel maxAbsScalerModel = maxAbsScaler.fit(trainDataWithInvalidData);
+            IteratorUtils.toList(
+                    tEnv.toDataStream(maxAbsScalerModel.getModelData()[0]).executeAndCollect());
```
zhipeng93, line 200: Shall we add `fail()` next line? For now, the test passes even we do not get into the exception.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
```diff
@@ -0,0 +1,286 @@
...
+        } catch (Exception e) {
+            assertEquals(
+                    "Input column data cannot be null.",
+                    ExceptionUtils.getRootCause(e).getMessage());
+        }
+    }
+
+    @Test
+    public void testTransformDataWithNullValue() throws Exception {
```
zhipeng93, line 209: Probably we should not allow transforming invalid data.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/maxabsscaler/MaxAbsScaler.java
```diff
@@ -0,0 +1,201 @@
...
+
+                    for (int i = 0; i < indices.length; ++i) {
+                        maxAbsVector.values[indices[i]] =
+                                Math.max(maxAbsVector.values[indices[i]], Math.abs(values[i]));
+                    }
+                }
+            } else {
+                Preconditions.checkArgument(
+                        currentValue.size() == maxAbsVector.size(),
+                        "CurrentValue should has same size with maxVector.");
```
zhipeng93, line 142: nit: How about update the error message as: `The training data should all have same dimensions`.

zhipeng93 (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/MaxAbsScalerTest.java
```diff
@@ -0,0 +1,251 @@
...
+                                Row.of((Object) null),
+                                Row.of(Vectors.dense(6.1, 8.1)),
+                                Row.of(Vectors.dense(200, 400))));
+
+        Table trainDataWithInvalidData =
+                tEnv.fromDataStream(env.fromCollection(trainData)).as("input");
+
+        try {
+            MaxAbsScaler maxAbsScaler = new MaxAbsScaler();
+            maxAbsScaler.fit(trainDataWithInvalidData);
```
zhipeng93, line 193: Could you insert a `fail();` after this line to verify that the code never goes there?

zhipeng93 (commented):
Thanks for the PR. LGTM overall except one minor comment.

zhipeng93:
LGTM.

PR #143, opened 2022-08-16
[FLINK-27535] Optimize the unit test execution time

yunfengzhou-hub:

This PR reduces the total execution time of Flink ML CIs so as to improve the efficiency of Flink ML development.

This PR mainly focuses on optimizing the execution time of Java tests. Python tests might be further optimized in separate PRs.

Compared with the current CI on the master branch, which took about 15 minutes[1], the CI of this PR would complete the same Java tests in about three minutes[2].

[1] https://github.com/apache/flink-ml/actions/runs/2872725840
[2] https://github.com/apache/flink-ml/actions/runs/2879967153

## Brief change log
- Make sure all flink-ml-lib tests create a static mini-cluster for all test cases in its class, reducing the overhead to start and shut down clusters
- Disable unaligned checkpoints in test cases
- Split maven tests into multiple CI jobs to improve concurrency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -553,7 +553,6 @@ under the License.
           <trimStackTrace>false</trimStackTrace>
           <systemPropertyVariables>
             <forkNumber>0${surefire.forkNumber}</forkNumber>
-            <checkpointing.randomization>true</checkpointing.randomization>
             <project.basedir>${project.basedir}</project.basedir>
             <!--suppress MavenModelInspection -->
             <test.randomization.seed>${test.randomization.seed}</test.randomization.seed>
--- a/.github/workflows/java8-build.yml
+++ b/.github/workflows/java8-build.yml
@@ -3,9 +3,11 @@ name: Java 8 Build
 on: [push, pull_request]
 
 jobs:
-  build:
+  java-tests:
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        stage: ['core', 'lib', 'tests', 'misc']
     steps:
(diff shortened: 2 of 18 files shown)


pom.xml
```diff
@@ -548,12 +548,13 @@ under the License.
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.22.1</version>
         <configuration>
+          <parallel>classes</parallel>
+          <useUnlimitedThreads>true</useUnlimitedThreads>
```
lindong28, line 552: If the number of classes is much more than the physical cores, would it cause performance regression?

pom.xml
```diff
@@ -548,12 +548,13 @@ under the License.
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.22.1</version>
         <configuration>
+          <parallel>classes</parallel>
```
lindong28, line 551: Do you know why apache/flink does not use parallel=classes? It might be safer to understand Flink's approach to make sure it is safe to use parallelize tests across classes before making this change.

lindong28 (commented):

pom.xml
yunfengzhou-hub: I have not found discussions about using this parameter in Flink, and I agree with it that running tests in parallel might bring risks. I'll remove this parameter.

lindong28:
Thanks for the update. Can you provide the statistics of the CI execution time before and after this PR? And can you update the PR description as appropriate?

PR #145, opened 2022-08-22
[FLINK-29044] Add Transformer for DCT

yunfengzhou-hub:

Add Transformer for DCT in Flink ML.

## Brief change log
- Added Transformer for DCT.
- Added java test/example for DCT.
- Added python source/test/example for DCT.
- The public interface is consistent with DCT in Spark ML.

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/dct/DCT.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/dct/DCT.java
@@ -1,0 +1,139 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.dct;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.TableUtils;
+import org.apache.flink.ml.linalg.DenseVector;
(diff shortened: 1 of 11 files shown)


lindong28:
Thanks for the PR. LGTM.

@zhipeng93 Do you want to take another look?

zhipeng93:
> Thanks for the PR. LGTM.
> 
> @zhipeng93 Do you want to take another look?

Thanks for the review.  I will do a pass today.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/dct/DCT.java
```diff
@@ -0,0 +1,138 @@
...
+    public Table[] transform(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+
+        RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
+        RowTypeInfo outputTypeInfo =
+                new RowTypeInfo(
+                        ArrayUtils.addAll(inputTypeInfo.getFieldTypes(), VectorTypeInfo.INSTANCE),
```
zhipeng93, line 72: It seems that the output is always a dense vector. How about make it `DenseVectorTypeInfo.INSTANCE`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/dct/DCT.java
```diff
@@ -0,0 +1,138 @@
...
+            this.isInverse = isInverse;
+            this.dctTransformer = null;
+            this.previousVectorSize = -1;
+        }
+
+        @Override
+        public Row map(Row row) throws Exception {
+            Vector vector = row.getFieldAs(inputCol);
+
+            if (dctTransformer == null || previousVectorSize != vector.size()) {
```
zhipeng93, line 105: nit: `dctTransformer == null` seems unnecessary.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/dct/DCTParams.java
```diff
@@ -0,0 +1,46 @@
...
+                    "inverse",
+                    "Whether to perform the inverse DCT (true) or forward DCT (false).",
+                    false);
+
+    default boolean getInverse() {
+        return get(INVERSE);
+    }
+
+    default T setInverse(boolean value) {
+        set(INVERSE, value);
```
zhipeng93, line 43: nit: could be `return set(INVERSE, value);`

zhipeng93 (commented):
Thanks for the PR. LGTM overall. Left some minor comments below.

BTW, do you think it is a good idea to introduce `JLargeArrays` here? It needs to allocate off-heap memory when the array size is bigger than INT.MAX, which is not friendly to Flink memory management.

In this PR I am ok since the size is always smaller than INT.MAX.

zhipeng93:
LGTM.

PR #144, opened 2022-08-19
[FLINK-29043] Reduce HeadOperator overhead

yunfengzhou-hub:
## What is the purpose of the change
Improve the performance of Flink ML iteration mechanism by reducing the overhead on HeadOperator

The performance improvement brought by this PR on the execution time of unit tests of algorithms with iterations are as follows.

|                        | Before this PR[1] | After this PR [2] |
| ---------------------- | ----------------- | ----------------- |
| LinearSVCTest          | 34.597 s          | 15.623 s          |
| LogisticRegressionTest | 48.325 s          | 24.649 s          |
| KMeansTest             | 32.816 s          | 29.967 s          |
| LinearRegressionTest   | 49.92 s           | 33.991 s          |

[1] https://github.com/apache/flink-ml/runs/7892649470?check_suite_focus=true
[2] https://github.com/apache/flink-ml/runs/7916115762?check_suite_focus=true

## Brief change log
- Add `MailboxExecutor` subclass, `MailboxExecutorWithYieldTimeout`, that supports `yield()` method that would not block infinitely
- Pass `MailboxExecutorFactory` to `HeadOperator()` to make sure it, or the `MailboxExecutorWithYieldTimeout` instance, has exclusive access to the `MailboxExecutor` instance.
- replace the `tryYield` and `sleep` in `HeadOperator.endInput()` with the `yield()` with timeout.

## Does this pull request potentially affect one of the following parts:
- Dependencies (does it add or upgrade a dependency): (no)
- The public API, i.e., is any changed class annotated with @public(Evolving): (no)
- Does this pull request introduce a new feature? (no)
- If yes, how is the feature documented? (N/A)

PR #147, opened 2022-08-23
[FLINK-29043] Reduce overhead for non-feedback HeadOperator mails

yunfengzhou-hub:

Improve the performance of Flink ML iteration mechanism by reducing the overhead on HeadOperator.

While #144 has reduced the overhead brought by `HeadOperator.endInput()`, it is discovered that the implementation has only reduced the overhead of `Mail`s related to feedback edge. The other mails, like operator events, still need to wait until timeout before they are able to be processed, which is equivalent to `Thread.sleep(timeout)`. Thus we need to improve the implementation in #144 to reduce the overhead for all mails the `HeadOperator` would receive.

The performance improvement brought by this PR on the unit tests of algorithms with iterations is as follows.

|                        | Before this PR [1] | After this PR [2] |
| ---------------------- | ------------------ | ----------------- |
| LinearSVCTest          | 21.195 s           | 5.111 s           |
| LogisticRegressionTest | 25.164 s           | 5.852 s           |
| KMeansTest             | 31.86 s            | 12.23 s           |
| LinearRegressionTest   | 32.526 s           | 7.736 s           |

[1] https://github.com/apache/flink-ml/runs/7966776640?check_suite_focus=true
[2] https://github.com/apache/flink-ml/runs/7968880728?check_suite_focus=true

## Brief change log
- Replace the lock and condition used in `HeadOperator.MailboxExecutorWithYieldTimeout` with a timer to schedule the timeout event.

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
@@ -76,22 +76,22 @@ import org.apache.flink.streaming.api.operators.Output;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.tasks.ProcessingTimeService;
 import org.apache.flink.streaming.runtime.tasks.StreamTask;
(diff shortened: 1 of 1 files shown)


lindong28:
Thanks for the PR. LGTM.

PR #148, opened 2022-08-26
[FLINK-28906] Add AlgoOperator for AgglomerativeClustering

zhipeng93:
- Add AlgoOperator for AgglomerativeClustering[1] in Flink ML. 

## Brief change log
- Added AlgoOperator for AgglomerativeClustering.
- Added java/python test/example  for Transformer and Estimator for KBinsDiscretizer.
- Comparing with Sklearn [1], we made the following changes:
     - The distance between two data points in sklearn could be `precomputed`, but we do not support this option in this PR
     - sklearn uses `memory` to cache the output of the computation of the tree`, but we do not support this in this PR.
     - sklearn uses `connectivity` to define for each sample the neighboring samples, but we do not support this in this PR.

[1] https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html


zhipeng93:
@lindong28 @yunfengzhou-hub Can you help to review this PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
```diff
@@ -0,0 +1,415 @@
...
+        private final String featuresCol;
+        private final String linkage;
+        private final DistanceMeasure distanceMeasure;
+        private final Integer k;
+        private final Double distanceThreshold;
+        private final boolean computeFullTree;
+        private final OutputTag<Tuple4<Integer, Integer, Double, Integer>> mergeInfoOutputTag;
+
+        /** Cache for the input data. */
+        private List<Row> inputList;
```
yunfengzhou-hub, line 162: It seems that this variable can be converted to a local variable.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
```diff
@@ -0,0 +1,415 @@
...
+
+                // Updates cluster Ids for each data point if clustering is still running.
+                if (clusteringRunning) {
+                    int mergedClusterId = mergedCluster.clusterId;
+                    for (int dataPointId : mergedCluster.dataPointIds) {
+                        clusterIds[dataPointId] = mergedClusterId;
+                    }
+                }
+
+                clusteringRunning =
```
yunfengzhou-hub, line 311: How about changing the order of the logics here?
```java
                clusteringRunning =
                        (k != null && activeClusters.size() > k)
                                || (distanceThreshold != null && distanceThreshold > minDistance);

                if (clusteringRunning) {
                    int mergedClusterId = mergedCluster.clusterId;
                    for (int dataPointId : mergedCluster.dataPointIds) {
                        clusterIds[dataPointId] = mergedClusterId;
                    }
                }
```

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
```diff
@@ -0,0 +1,415 @@
...
+
+        Table outputTable = tEnv.fromDataStream(output);
+
+        DataStream<Tuple4<Integer, Integer, Double, Integer>> mergeInfo =
+                output.getSideOutput(mergeInfoOutputTag);
+        Table mergeInfoTable =
+                tEnv.fromDataStream(mergeInfo)
+                        .as("clusterId1", "clusterId2", "distance", "sizeOfMergedCluster");
+
+        return new Table[] {outputTable, mergeInfoTable};
```
yunfengzhou-hub, line 133: Let's describe in JavaDoc the column names of the `mergeInfoTable` and their meanings.

Besides, do you think it would be better to add some method like `Table getMergeInfo()`, instead of making the merge info table the second output table? I'm not sure the current API corresponds to the syntax of Flink ML AlgoOperators. The current API also makes it difficult to connect `AgglomerativeClustering` stage with other stages in a `Pipeline`.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.flink.ml.param.StringParam;
+
+/**
+ * Params of {@link AgglomerativeClustering}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface AgglomerativeClusteringParams<T>
+        extends HasDistanceMeasure<T>, HasFeaturesCol<T>, HasPredictionCol<T> {
+    Param<Integer> K = new IntParam("k", "The max number of clusters to create.", 2);
```
yunfengzhou-hub, line 38: `k` has a clear meaning in Kmeans and Knn, but I'm not sure users of Agglomerative Clustering algorithm. How about changing this parameter to `numClusters` or `maxClusters`?

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
```diff
@@ -0,0 +1,107 @@
...
+                    "Criterion for computing distance between two clusters.",
+                    LINKAGE_WARD,
+                    ParamValidators.inArray(
+                            LINKAGE_WARD, LINKAGE_COMPLETE, LINKAGE_AVERAGE, LINKAGE_SINGLE));
+
+    Param<Boolean> COMPUTE_FULL_TREE =
+            new BooleanParam(
+                    "computeFullTree",
+                    "Whether computes the full tree after convergence.",
+                    false,
```
yunfengzhou-hub, line 73: The default behavior here is different from that in sklearn. Do you think we should also support the `auto` option here?

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
```diff
@@ -0,0 +1,415 @@
...
+                    }
+                }
+
+                clusteringRunning =
+                        (k != null && activeClusters.size() > k)
+                                || (distanceThreshold != null && distanceThreshold > minDistance);
+            }
+        }
+
+        private double computeDistanceBetweenClusters(Cluster cluster1, Cluster cluster2) {
```
yunfengzhou-hub, line 317: How about adding a cache here for the distance between any two data points? We can avoid re-computing the distance if there is a cache hit.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        /** A cluster with cluster Id specified and data points that belong to this cluster. */
+        private static class Cluster {
+            private final int clusterId;
+            private final List<Integer> dataPointIds;
+
+            @SafeVarargs
+            public Cluster(
+                    int clusterId, List<Integer> dataPointIds, List<Integer>... otherDataPointIds) {
```
yunfengzhou-hub, line 406: It might be better to write two constructors, `Cluster(int, List<Integer>)` and `Cluster(int, List<Integer>, List<Integer>)`.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
zhipeng93: No, we need to merge the cluster in this round first.

Say when entering the loop, the number of active clusters is `K+1`. If we change the order as above, the number of output clusters will be `k+1`.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
zhipeng93: The memory overhead for caching the distance between two data points is O(n^2), where n is the number of data points.

Suppose we store the distance in a primitive double array, then the max number of data points we can deal with is Math.sqrt(Int.MAX), which is less than 100000. 

A possible optimization would be cache the distance between each two clusters at the later iterations, i.e., when the number of active clusters is smaller. But we can do this in a later PR.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
zhipeng93: `k` here is exactly the same with that in Kmeans. I didn't extract it as shared param because `k` is used in many different cases.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
zhipeng93: Thanks for the comment. I have added the following description in Java doc.
```
The output contains two tables. The first one assigns one cluster Id for each data point.
The second one contains the information of merging two clusters at each step. The data format
of the merging information is (clusterId1, clusterId2, distance, sizeOfMergedCluster).
```

I am not sure about adding a `Table getMergeInfo()` here, since it is somehow adding a public API. 

For the semantic of  `AgglomerativeClustering`, I made it an AlgoOperator because (1) it is not a Estimator since it has not model data and cannot be used to predict some new data. (2) It is also not a Transformer. 

It is clear that not all AlgoOperator/Estimator could be used in Pipeline. For those multi-input/output stages, we have `Graph`.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/AgglomerativeClusteringTest.java
```diff
@@ -0,0 +1,312 @@
...
+        Table[] outputs = agglomerativeClustering.transform(inputDataTable);
+        verifyClusteringResult(EUCLIDEAN_WARD_K_AS_TWO_RESULT, outputs[0]);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void verifyMergeInfo(List<Row> expected, Table mergeInfoTable) throws Exception {
+        List<Row> mergeInfo =
+                IteratorUtils.toList(tEnv.toDataStream(mergeInfoTable).executeAndCollect());
+        assertEquals(expected.size(), mergeInfo.size());
+        mergeInfo.sort(Comparator.comparingInt(o -> o.getFieldAs(0)));
```
yunfengzhou-hub, line 292: `mergeInfo` should be order-sensitive. I think we should verify that the order of the output merge information is also correct.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/AgglomerativeClusteringTest.java
```diff
@@ -0,0 +1,312 @@
...
+    @Before
+    public void before() {
+        Configuration config = new Configuration();
+        config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
+        env = StreamExecutionEnvironment.getExecutionEnvironment(config);
+        env.setParallelism(3);
+        env.enableCheckpointing(100);
+        env.setRestartStrategy(RestartStrategies.noRestart());
+        tEnv = StreamTableEnvironment.create(env);
+        inputDataTable = tEnv.fromDataStream(env.fromCollection(INPUT_DATA)).as("features");
```
yunfengzhou-hub, line 147: If the input data comes from an operator whose parallelism is 3, how can we make sure that they will arrive at the `LocalAgglomerativeClusteringOperator`, whose parallelism is 1, in the same order and be assigned the same cluster ids across difference runs? If each record might be assigned a different id in different runs, why would the merge info remains the same and can be verified with an expected value?

yunfengzhou-hub (commented):
Thanks for updating the PR. Left some comments as below.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
zhipeng93: With some offline discussion, we aggree that not supporting `auto` is ok for users.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
yunfengzhou-hub: Thanks for the explanation. I agree with it that the current API is better.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/AgglomerativeClusteringTest.java
zhipeng93: Good catch. I have changed the unit test to verify the order of the merge information.

flink-ml-lib/src/test/java/org/apache/flink/ml/clustering/AgglomerativeClusteringTest.java
zhipeng93: Good catch. I made a mistake --- and the mergeInfo should be different in different runs and only the distances should be the same accross different runs. 

I have added a `map` after `env.fromCollection()` to force it as a parallel stream and also re-wrote the corresponding unit test.

yunfengzhou-hub (commented):
Thanks for the update. LGTM.

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
```diff
@@ -0,0 +1,108 @@
...
+
+    String LINKAGE_WARD = "ward";
+    String LINKAGE_COMPLETE = "complete";
+    String LINKAGE_SINGLE = "single";
+    String LINKAGE_AVERAGE = "average";
+    /**
+     * Supported options to compute the distance between two clusters.
+     *
+     * <ul>
+     *   <li>ward: difference between the sum of the variance of the two clusters and the merged
```
lindong28, line 55: It seems inconsistent to have `merged one` for this option but not the options described below.
 
How about this:

`ward: the sum of variance of the two clusters`

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
```diff
@@ -0,0 +1,108 @@
...
+     *       one.
+     *   <li>complete: the maximum distance between all observations of the two clusters.
+     *   <li>single: the minimum distance between all observations of the two cluster.
+     *   <li>average: the average of the distance of all observations of the two cluster.
```
lindong28, line 59: Would the following be more consistent with the options described above:

```
the average distance between all observations of the two clusters.
```

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
```diff
@@ -0,0 +1,108 @@
...
+    String LINKAGE_SINGLE = "single";
+    String LINKAGE_AVERAGE = "average";
+    /**
+     * Supported options to compute the distance between two clusters.
+     *
+     * <ul>
+     *   <li>ward: difference between the sum of the variance of the two clusters and the merged
+     *       one.
+     *   <li>complete: the maximum distance between all observations of the two clusters.
+     *   <li>single: the minimum distance between all observations of the two cluster.
```
lindong28, line 58: cluster -> clusters

flink-ml-python/pyflink/ml/lib/clustering/agglomerativeclustering.py
```diff
@@ -0,0 +1,135 @@
...
+        return self.get_linkage()
+
+    @property
+    def compute_full_tree(self):
+        return self.get_compute_full_tree()
+
+
+class AgglomerativeClustering(JavaClusteringAlgoOperator, _AgglomerativeClusteringParams):
+    """
+    An AlgoOperator that performs a hierarchical clustering using a bottom up approach. Each
```
lindong28, line 113: bottom up -> bottom-up

flink-ml-python/pyflink/ml/lib/clustering/agglomerativeclustering.py
```diff
@@ -0,0 +1,135 @@
...
+    observation starts in its own cluster and the clusters are merged together one by one.
+    Users can choose different strategies to merge two clusters by setting
+    {@link AgglomerativeClusteringParams#LINKAGE} and different distance measure by setting
```
lindong28, line 116: distance measure -> distance measures

flink-ml-python/pyflink/ml/lib/clustering/agglomerativeclustering.py
```diff
@@ -0,0 +1,135 @@
...
+    {@link AgglomerativeClusteringParams#DISTANCE_MEASURE}.
+
+    <p>The output contains two tables. The first one assigns one cluster Id for each data point.
+    The second one contains the information of merging two clusters at each step. The data format
+    of the merging information is (clusterId1, clusterId2, distance, sizeOfMergedCluster).
```
lindong28, line 121: Maybe provide path to the scripts that visualize the 2nd output.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClustering.java
```diff
@@ -0,0 +1,421 @@
...
+                    BLAS.axpy(1.0, mean2, meanMerged);
+                    BLAS.scal(1.0 / size1, mean1);
+                    BLAS.scal(1.0 / size2, mean2);
+                    BLAS.scal(1.0 / (size1 + size2), meanMerged);
+                    double essInc =
+                            size1 * BLAS.dot(mean1, mean1)
+                                    + size2 * BLAS.dot(mean2, mean2)
+                                    - (size1 + size2) * BLAS.dot(meanMerged, meanMerged);
+
+                    distance = Math.sqrt(2 * essInc);
```
lindong28, line 394: Since the calculated distance is used to find the pair of clusters that minimize this distance, would it be more efficient to just return `essInc` as the distance?

Suppose `distance_1 = Math.sqrt(2 * essInc_1)` and `distance_2 = Math.sqrt(2 * essInc_2)`, it is guaranteed that `distance_1 < distance_2` if and only if `essInc_1 < essInc_2`.

zhipeng93: The distance will be returned as merge info, which is consistent with sklearn's implementation.

flink-ml-python/pyflink/ml/lib/clustering/agglomerativeclustering.py
zhipeng93: Thanks for the comment. An example script to visualize the merge info is added [1]. It is also mentioned in the user doc [2].


[1] flink-ml-dist/src/main/flink-ml-bin/bin/agglomerativeclustering-visualize.py
[2] docs/content/docs/operators/clustering/agglomerativeclustering.md

flink-ml-lib/src/main/java/org/apache/flink/ml/clustering/agglomerativeclustering/AgglomerativeClusteringParams.java
```diff
@@ -0,0 +1,108 @@
...
+                    "Threshold to decide whether two clusters should be merged.",
+                    null);
+
+    String LINKAGE_WARD = "ward";
+    String LINKAGE_COMPLETE = "complete";
+    String LINKAGE_SINGLE = "single";
+    String LINKAGE_AVERAGE = "average";
+    /**
+     * Supported options to compute the distance between two clusters. The algorithm will merge the
+     * pairs of cluster that minimize this criterion
```
lindong28, line 53: nits: Add `.` at the end?

lindong28 (commented):
Thanks

PR #149, opened 2022-08-29
[FLINK-29115] Improve Python API installation process and quickstart document

yunfengzhou-hub:

This PR improves the installation process of Flink ML's Python SDK, so that Python users could set up Flink ML dependency with a one-line `pip install` command, instead of spending time compiling both Java and Python SDK before they can use Flink ML. This PR also modifies the python example codes and documents to reflect the latest setup process and to provide a quickstart example for python users.

## Brief change log
- Add Flink ML jars to the Python `sdist` package
  The setup code is referenced from the initial version Flink's `setup.py`[1], when PyFlink has not been divided into `apache-flink` and `apache-flink-libraries`.
- Remove the process to load jars in Python tests.
- Remove the instruction to set up jars in Python examples.
- Add documents about compiling, installing, and running a quickstart example of Flink ML's Python SDK.
  The document is referenced from Flink's document to compile or install PyFlink[2][3].

[1] https://github.com/apache/flink/blob/f059b1981b8d2a62acfd2fac3456b8ca3a5bd18f/flink-python/setup.py
[2] https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/python/installation/
[3] https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/flinkdev/building/

## Verifying changes in this PR

Reviewers of this PR can view the newly introduced documents by starting a Hugo server in the local environment with the following command:
```shell
cd docs; hugo -b "" serve
```

The command above would start the Flink ML's document website at localhost:1313.

The changes in the Python installation process can be verified by following the Python compilation and quick-start documents. Simply speaking, they can be verified with the following commands:
```shell
mvn clean install -DskipTests
python -m pip install -r flink-ml-python/dev/dev-requirements.txt
cd flink-ml-python; python setup.py sdist; cd ..;
python -m pip install flink-ml-python/dist/*.tar.gz
python -m pyflink.examples.ml.clustering.kmeans_example
```

- If yes, how is the feature documented? (website document)

### DIFF
--- a/.github/workflows/python-checks.yml
+++ b/.github/workflows/python-checks.yml
@@ -1,3 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 name: Python Checks
 
 on:
--- a/.github/workflows/java8-build.yml
+++ b/.github/workflows/java8-build.yml
@@ -1,3 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 name: Java 8 Build
 
 on: [push, pull_request]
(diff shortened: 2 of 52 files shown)


.github/workflows/python-checks.yml
```diff
@@ -32,6 +32,12 @@ jobs:
       - name: Install dependencies
         working-directory: flink-ml-python/dev
         run: python -m pip install -r dev-requirements.txt
+      - name: Build Python package
+        working-directory: flink-ml-python
+        run: python setup.py sdist
+      - name: Install Python package
+        working-directory: flink-ml-python
+        run: python -m pip install dist/*.tar.gz
```
HuangXingBo, line 40: I think we can create a dedicated workflow action to build python package and this can be used in the release.

flink-ml-python/pyflink/ml/tests/test_utils.py
```diff
@@ -43,30 +40,3 @@ def setUp(self):
...
-                self.env.add_classpaths("file://{0}/{1}".format(FLINK_ML_LIB_PATH, file))
-
-        # load flink-ml-lib/flink-ml-lib-*-tests.jar
-        FLINK_ML_LIB_SOURCE_PATH = os.path.abspath(os.path.join(
-            this_directory, "../../../../flink-ml-lib"))
-
-        ml_test_jar = glob.glob(os.path.join(
-            FLINK_ML_LIB_SOURCE_PATH, "target", "flink-ml-lib-*-tests.jar"))[0]
-
-        self.env.add_classpaths("file://{0}".format(ml_test_jar))
```
HuangXingBo, line 72: Why remove this logic? We can't let developers need to reinstall every time when they modify the code to make the changed code take effect

.github/workflows/python-checks.yml
HuangXingBo: I found we forgot to add license header to `java8-build.yml` and `python-checks.yml`

yunfengzhou-hub: I have added the license header to Github Action config files.

About creating a workflow to build the python package, I think there is no need to do so now, given that the workload to build the package on a local machine is acceptable so far, and the process is not complicated.

yunfengzhou-hub: I have added the license header to Github Action config files.

About creating a workflow to build the python package, I think there is no need to do so now, given that the workload to build the package on a local machine is acceptable so far, and the process is not complicated.

HuangXingBo (approved):

PR #154, opened 2022-09-05
[FLINK-29193] Add Transformer for Normalizer

weibozhao:
Add Transformer for Normalizer in Flink ML.

## Brief change log
Added java/python source/test/example for Transformer for Normalizer in Flink ML.
The public interface is consistent with Normalizer in Spark ML.
Dependencies (does it add or upgrade a dependency): (yes)

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
@@ -129,9 +129,41 @@ public class BLAS {
         return JAVA_BLAS.dnrm2(x.values.length, x.values, 1);
     }
 
+    /** Calculates the p-norm of the vector x. */
+    public static double norm(Vector x, double p) {
+        Preconditions.checkArgument(p >= 1.0, "p value must >= 1.0, but the current p is : " + p);
+        double norm = 0.0;
+        double[] data =
+                (x instanceof DenseVector) ? ((DenseVector) x).values : ((SparseVector) x).values;
+
+        if (p == 1.0) {
+            for (double datum : data) {
+                norm += Math.abs(datum);
+            }
+        } else if (p == 2.0) {
+            norm = norm2(x);
+        } else if (p == Double.POSITIVE_INFINITY) {
+            for (double datum : data) {
+                norm = Math.max(Math.abs(datum), norm);
+            }
+        } else {
+            for (double datum : data) {
+                norm += Math.pow(Math.abs(datum), p);
+            }
+            norm = Math.pow(norm, 1.0 / p);
+        }
+
+        return norm;
+    }
+
     /** x = x * a . */
-    public static void scal(double a, DenseVector x) {
-        JAVA_BLAS.dscal(x.size(), a, x.values, 1);
+    public static void scal(double a, Vector x) {
+        if (x instanceof DenseVector) {
+            JAVA_BLAS.dscal(x.size(), a, ((DenseVector) x).values, 1);
+        } else {
+            double[] values = ((SparseVector) x).values;
+            JAVA_BLAS.dscal(values.length, a, values, 1);
+        }
     }
 
     /**
--- a/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/NormalizerExample.java
+++ b/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/NormalizerExample.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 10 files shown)


flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -129,6 +129,33 @@ private static double norm2(SparseVector x) {
         return JAVA_BLAS.dnrm2(x.values.length, x.values, 1);
     }
 
+    /** Calculates the p-norm of the vector x. */
+    public static double norm(Vector x, double p) {
+        Preconditions.checkArgument(p >= 1.0, "p value must >= 1.0.");
```
zhipeng93, line 134: How about we update the error message to also print the value of `p`?

e.g., "To compute p-norm of a vector, p should be greater than or equal to 1.0. However the specified p is " + p.

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -129,6 +129,33 @@ private static double norm2(SparseVector x) {
...
+        double[] data =
+                (x instanceof DenseVector) ? ((DenseVector) x).values : ((SparseVector) x).values;
+
+        if (p == 1.0) {
+            for (double datum : data) {
+                norm += Math.abs(datum);
+            }
+        } else if (p == 2.0) {
+            norm = norm2(x);
+        } else if (Double.isInfinite(p)) {
```
zhipeng93, line 145: It should be `p == Double.POSITIVE_INFINITY`

flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/NormalizerExample.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.flink.ml.linalg.Vector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+
+/** Simple program that creates a Normalizer instance and uses it for feature engineering. */
```
zhipeng93, line 31: It should be `an` Normalizer. 

Same for all other cases in this PR.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/normalizer/Normalizer.java
```diff
@@ -0,0 +1,115 @@
...
+            this.inputCol = inputCol;
+        }
+
+        @Override
+        public Row map(Row row) throws Exception {
+            Vector inputVec = row.getFieldAs(inputCol);
+            Vector outputVec = inputVec.clone();
+            double norm = BLAS.norm(inputVec, p);
+
+            double[] data =
```
zhipeng93, line 104: This logic could be simplified with `BLAS.scal`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/normalizer/Normalizer.java
```diff
@@ -0,0 +1,115 @@
...
+    public static Normalizer load(StreamTableEnvironment env, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    /** Normalize function which normalizes a vector to have unit norm. */
```
zhipeng93, line 88: How about updating the class name to `NormalizationFunction` and update the comment as:

"Normalization function that transforms a vector to have unit norm."

flink-ml-core/src/main/java/org/apache/flink/ml/linalg/BLAS.java
```diff
@@ -129,6 +129,33 @@ private static double norm2(SparseVector x) {
         return JAVA_BLAS.dnrm2(x.values.length, x.values, 1);
     }
 
+    /** Calculates the p-norm of the vector x. */
+    public static double norm(Vector x, double p) {
```
zhipeng93, line 133: Could you please also add a unit test for function `norm`?

zhipeng93 (commented):
Thanks for the PR. LGTM overall. Left some minor comments.

flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/NormalizerExample.java
weibozhao: It should be a Normalizer According to Grammer prompt.

zhipeng93:
Thanks for the updat. LGTM.

PR #156, opened 2022-09-16
[FLINK-29323] Add inputSizes parameter for VectorAssembler

weibozhao:
Add inputSizes parameter for VectorAssembler

## Brief change log
Add inputSizes parameter for VectorAssembler.

The public API, i.e., is any changed class annotated with @public(Evolving): (yes)

### DIFF
--- a/docs/content/docs/operators/feature/vectorassembler.md
+++ b/docs/content/docs/operators/feature/vectorassembler.md
@@ -26,10 +26,20 @@ under the License.
 -->
 
 ## Vector Assembler
-
-Vector Assembler combines a given list of input columns into a vector column.
-Types of input columns must be either vector or numerical value.
-
+A Transformer which combines a given list of input columns into a vector column. Input columns
+would be numerical or vectors whose sizes are specified by the {@link #INPUT_SIZES} parameter.
+Invalid input data with null values or values with wrong sizes would be dealt with according to
+the strategy specified by the {@link HasHandleInvalid} parameter as follows:
+<ul>
+   <li>keep: If the input column data is null, a vector would be created with the specified size
+       and NaN values. The vector would be used in the assembling process to represent the input
+       column data. If the input column data is a vector, the data would be used in the assembling
+       process even if it has a wrong size.
+   <li>skip: If the input column data is null or a vector with wrong size, the input row would be
+       filtered out and not be sent to downstream operators.
+   <li>error: If the input column data is null or a vector with wrong size, an exception would be
+       thrown.
+</ul>
 ### Input Columns
 
 | Param name | Type          | Default | Description                     |
@@ -44,11 +54,12 @@ Types of input columns must be either vector or numerical value.
 
 ### Parameters
 
-| Key           | Default    | Type     | Required | Description                                                                    |
-|---------------|------------|----------|----------|--------------------------------------------------------------------------------|
-| inputCols     | `null`     | String[] | yes      | Input column names.                                                            |
-| outputCol     | `"output"` | String   | no       | Output column name.                                                            |
-| handleInvalid | `"error"`  | String   | no       | Strategy to handle invalid entries. Supported values: 'error', 'skip', 'keep'. |
+| Key             | Default    | Type      | Required | Description                                                                    |
+|-----------------|------------|-----------|----------|--------------------------------------------------------------------------------|
+| inputCols       | `null`     | String[]  | yes      | Input column names.                                                            |
+| outputCol       | `"output"` | String    | no       | Output column name.                                                            |
+| inputSizes      | `null`     | Integer[] | yes      | Sizes of the input elements to be assembled.                                   |
+| handleInvalid   | `"error"`  | String    | no       | Strategy to handle invalid entries. Supported values: 'error', 'skip', 'keep'. |
 
 ### Examples
 
@@ -95,7 +106,8 @@ public class VectorAssemblerExample {
         VectorAssembler vectorAssembler =
                 new VectorAssembler()
                         .setInputCols("vec", "num", "sparseVec")
-                        .setOutputCol("assembledVec");
+                        .setOutputCol("assembledVec")
+                        .setInputSizes(2, 1, 5);
 
         // Uses the VectorAssembler object for feature transformations.
         Table outputTable = vectorAssembler.transform(inputTable)[0];
@@ -159,9 +171,10 @@ input_data_table = t_env.from_data_stream(
 vector_assembler = VectorAssembler() \
     .set_input_cols('vec', 'num', 'sparse_vec') \
     .set_output_col('assembled_vec') \
+    .set_input_sizes(2, 1, 5) \
     .set_handle_invalid('keep')
 
-# use the vector assembler model for feature engineering
+# use the vector assembler for feature engineering
 output = vector_assembler.transform(input_data_table)[0]
 
 # extract and display the results
--- a/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/VectorAssemblerExample.java
+++ b/flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/VectorAssemblerExample.java
@@ -56,7 +56,8 @@ public class VectorAssemblerExample {
         VectorAssembler vectorAssembler =
                 new VectorAssembler()
                         .setInputCols("vec", "num", "sparseVec")
-                        .setOutputCol("assembledVec");
+                        .setOutputCol("assembledVec")
+                        .setInputSizes(2, 1, 5);
 
         // Uses the VectorAssembler object for feature transformations.
         Table outputTable = vectorAssembler.transform(inputTable)[0];
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
@@ -19,6 +19,7 @@
 package org.apache.flink.ml.feature.vectorassembler;
 
 import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
 import org.apache.flink.ml.api.Transformer;
 import org.apache.flink.ml.common.datastream.TableUtils;
@@ -46,11 +47,21 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * A Transformer which combines a given list of input columns into a vector column. Types of input
- * columns must be either vector or numerical value.
+ * A Transformer which combines a given list of input columns into a vector column. Input columns
+ * would be numerical or vectors whose sizes are specified by the {@link #INPUT_SIZES} parameter.
+ * Invalid input data with null values or values with wrong sizes would be dealt with according to
+ * the strategy specified by the {@link HasHandleInvalid} parameter as follows:
  *
- * <p>The `keep` option of {@link HasHandleInvalid} means that we output bad rows with output column
- * set to null.
+ * <ul>
+ *   <li>keep: If the input column data is null, a vector would be created with the specified size
+ *       and NaN values. The vector would be used in the assembling process to represent the input
+ *       column data. If the input column data is a vector, the data would be used in the assembling
+ *       process even if it has a wrong size.
+ *   <li>skip: If the input column data is null or a vector with wrong size, the input row would be
+ *       filtered out and not be sent to downstream operators.
+ *   <li>error: If the input column data is null or a vector with wrong size, an exception would be
+ *       thrown.
+ * </ul>
  */
 public class VectorAssembler
         implements Transformer<VectorAssembler>, VectorAssemblerParams<VectorAssembler> {
@@ -64,6 +75,7 @@ public class VectorAssembler
     @Override
     public Table[] transform(Table... inputs) {
         Preconditions.checkArgument(inputs.length == 1);
+        Preconditions.checkArgument(getInputSizes().length == getInputCols().length);
         StreamTableEnvironment tEnv =
                 (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
         RowTypeInfo inputTypeInfo = TableUtils.getRowTypeInfo(inputs[0].getResolvedSchema());
@@ -74,64 +86,107 @@ public class VectorAssembler
         DataStream<Row> output =
                 tEnv.toDataStream(inputs[0])
                         .flatMap(
-                                new AssemblerFunc(getInputCols(), getHandleInvalid()),
+                                new AssemblerFunction(
+                                        getInputCols(), getHandleInvalid(), getInputSizes()),
                                 outputTypeInfo);
         Table outputTable = tEnv.fromDataStream(output);
         return new Table[] {outputTable};
     }
 
-    private static class AssemblerFunc implements FlatMapFunction<Row, Row> {
+    private static class AssemblerFunction implements FlatMapFunction<Row, Row> {
         private final String[] inputCols;
         private final String handleInvalid;
+        private final Integer[] inputSizes;
+        private final boolean keepInvalid;
 
-        public AssemblerFunc(String[] inputCols, String handleInvalid) {
+        public AssemblerFunction(String[] inputCols, String handleInvalid, Integer[] inputSizes) {
             this.inputCols = inputCols;
             this.handleInvalid = handleInvalid;
+            this.inputSizes = inputSizes;
+            keepInvalid = handleInvalid.equals(HasHandleInvalid.KEEP_INVALID);
         }
 
         @Override
         public void flatMap(Row value, Collector<Row> out) {
-            int nnz = 0;
-            int vectorSize = 0;
             try {
-                for (String inputCol : inputCols) {
-                    Object object = value.getField(inputCol);
-                    Preconditions.checkNotNull(object, "Input column value should not be null.");
-                    if (object instanceof Number) {
-                        nnz += 1;
-                        vectorSize += 1;
-                    } else if (object instanceof SparseVector) {
-                        nnz += ((SparseVector) object).indices.length;
-                        vectorSize += ((SparseVector) object).size();
-                    } else if (object instanceof DenseVector) {
-                        nnz += ((DenseVector) object).size();
-                        vectorSize += ((DenseVector) object).size();
-                    } else {
-                        throw new IllegalArgumentException(
-                                "Input type has not been supported yet.");
-                    }
-                }
+                Tuple2<Integer, Integer> vectorSizeAndNnz = computeVectorSizeAndNnz(value);
+                int vectorSize = vectorSizeAndNnz.f0;
+                int nnz = vectorSizeAndNnz.f1;
+                Vector assembledVec =
+                        nnz * RATIO > vectorSize
+                                ? assembleDense(inputCols, value, vectorSize)
+                                : assembleSparse(inputCols, value, vectorSize, nnz);
+                out.collect(Row.join(value, Row.of(assembledVec)));
             } catch (Exception e) {
-                switch (handleInvalid) {
-                    case ERROR_INVALID:
-                        throw e;
-                    case SKIP_INVALID:
-                        return;
-                    case KEEP_INVALID:
-                        out.collect(Row.join(value, Row.of((Object) null)));
-                        return;
-                    default:
-                        throw new UnsupportedOperationException(
-                                "Unsupported " + HANDLE_INVALID + " type: " + handleInvalid);
+                if (handleInvalid.equals(ERROR_INVALID)) {
+                    throw new RuntimeException("Vector assembler failed with exception : " + e);
                 }
             }
+        }
 
-            boolean toDense = nnz * RATIO > vectorSize;
-            Vector assembledVec =
-                    toDense
-                            ? assembleDense(inputCols, value, vectorSize)
-                            : assembleSparse(inputCols, value, vectorSize, nnz);
-            out.collect(Row.join(value, Row.of(assembledVec)));
+        private Tuple2<Integer, Integer> computeVectorSizeAndNnz(Row value) {
+            int vectorSize = 0;
+            int nnz = 0;
+            for (int i = 0; i < inputCols.length; ++i) {
+                Object object = value.getField(inputCols[i]);
+                if (object != null) {
+                    if (object instanceof Number) {
+                        checkSize(inputSizes[i], 1);
+                        if (Double.isNaN(((Number) object).doubleValue()) && !keepInvalid) {
+                            throw new RuntimeException(
+                                    "Encountered NaN while assembling a row with handleInvalid = 'error'. Consider "
+                                            + "removing NaNs from dataset or using handleInvalid = 'keep' or 'skip'.");
+                        }
+                        vectorSize += 1;
+                        nnz += 1;
+                    } else if (object instanceof SparseVector) {
+                        int localSize = ((SparseVector) object).size();
+                        checkSize(inputSizes[i], localSize);
+                        nnz += ((SparseVector) object).indices.length;
+                        vectorSize += localSize;
+                    } else if (object instanceof DenseVector) {
+                        int localSize = ((DenseVector) object).size();
+                        checkSize(inputSizes[i], localSize);
+                        vectorSize += localSize;
+                        nnz += ((DenseVector) object).size();
+                    } else {
+                        throw new IllegalArgumentException(
+                                String.format(
+                                        "Input type %s has not been supported yet. Only Vector and Number types are supported.",
+                                        object.getClass()));
+                    }
+                } else {
+                    vectorSize += inputSizes[i];
+                    nnz += inputSizes[i];
+                    if (keepInvalid) {
+                        if (inputSizes[i] > 1) {
+                            DenseVector tmpVec = new DenseVector(inputSizes[i]);
+                            for (int j = 0; j < inputSizes[i]; ++j) {
+                                tmpVec.values[j] = Double.NaN;
+                            }
+                            value.setField(inputCols[i], tmpVec);
+                        } else {
+                            value.setField(inputCols[i], Double.NaN);
+                        }
+                    } else {
+                        throw new RuntimeException(
+                                "Input column value is null. Please check the input data or using handleInvalid = 'keep'.");
+                    }
+                }
+            }
+            return Tuple2.of(vectorSize, nnz);
+        }
+
+        private void checkSize(int expectedSize, int currentSize) {
+            if (keepInvalid) {
+                return;
+            }
+            if (currentSize != expectedSize) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Input vector/number size does not meet with expected. Expected size: %d, actual size: %s.",
+                                expectedSize, currentSize));
+            }
         }
     }
 
@@ -167,8 +222,7 @@ public class VectorAssembler
 
             } else {
                 DenseVector denseVector = (DenseVector) object;
-                System.arraycopy(
-                        denseVector.values, 0, values, currentOffset, denseVector.values.length);
+                System.arraycopy(denseVector.values, 0, values, currentOffset, denseVector.size());
 
                 currentOffset += denseVector.size();
             }
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
@@ -21,6 +21,9 @@ package org.apache.flink.ml.feature.vectorassembler;
 import org.apache.flink.ml.common.param.HasHandleInvalid;
 import org.apache.flink.ml.common.param.HasInputCols;
 import org.apache.flink.ml.common.param.HasOutputCol;
+import org.apache.flink.ml.param.IntArrayParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidator;
 
 /**
  * Params of {@link VectorAssembler}.
@@ -28,4 +31,34 @@ import org.apache.flink.ml.common.param.HasOutputCol;
  * @param <T> The class type of this instance.
  */
 public interface VectorAssemblerParams<T>
-        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {}
+        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {
+    Param<Integer[]> INPUT_SIZES =
+            new IntArrayParam(
+                    "inputSizes",
+                    "Sizes of the input elements to be assembled.",
+                    null,
+                    sizesValidator());
+
+    default Integer[] getInputSizes() {
+        return get(INPUT_SIZES);
+    }
+
+    default T setInputSizes(Integer... value) {
+        return set(INPUT_SIZES, value);
+    }
+
+    // Checks the inputSizes parameter.
+    static ParamValidator<Integer[]> sizesValidator() {
+        return inputSizes -> {
+            if (inputSizes == null) {
+                return false;
+            }
+            for (Integer size : inputSizes) {
+                if (size <= 0) {
+                    return false;
+                }
+            }
+            return inputSizes.length != 0;
+        };
+    }
+}
--- a/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
+++ b/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
@@ -45,13 +45,14 @@ import java.util.List;
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
 
 /** Tests VectorAssembler. */
 public class VectorAssemblerTest extends AbstractTestBase {
 
     private StreamTableEnvironment tEnv;
     private Table inputDataTable;
+    private Table inputNullDataTable;
+    private Table inputNanDataTable;
 
     private static final List<Row> INPUT_DATA =
             Arrays.asList(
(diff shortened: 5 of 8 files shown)


yunfengzhou-hub:
Hi @weibozhao , could you please refine the description and JavaDocs of this PR, illustrating why do we need to add the `size` parameter to VectorAssembler? What additional functionalities has this parameter support, and how is this parameter compatible with the original public API?

flink-ml-examples/src/main/java/org/apache/flink/ml/examples/feature/VectorAssemblerExample.java
```diff
@@ -56,7 +56,8 @@ public static void main(String[] args) {
         VectorAssembler vectorAssembler =
                 new VectorAssembler()
                         .setInputCols("vec", "num", "sparseVec")
-                        .setOutputCol("assembledVec");
+                        .setOutputCol("assembledVec")
+                        .setSizes(2, 1, 5);
```
yunfengzhou-hub, line 60: Let's update VectorAssembler's markdown document accordingly, including its parameter list, examples and possibly descriptions.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
...
 
 /**
  * Params of {@link VectorAssembler}.
  *
  * @param <T> The class type of this instance.
  */
 public interface VectorAssemblerParams<T>
-        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {}
+        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {
+    Param<Integer[]> SIZES =
```
yunfengzhou-hub, line 37: In referent to the `splitsArray` parameter in Bucketizer and `VectorSizeHint` in Spark VectorAssembler, do you think it would be better to rename this parameter to `vectorSizeArray` or `elementSizeArray`?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,38 +74,65 @@ public Table[] transform(Table... inputs) {
...
             int nnz = 0;
             int vectorSize = 0;
             try {
-                for (String inputCol : inputCols) {
+                for (int i = 0; i < inputCols.length; ++i) {
+                    String inputCol = inputCols[i];
                     Object object = value.getField(inputCol);
                     Preconditions.checkNotNull(object, "Input column value should not be null.");
                     if (object instanceof Number) {
+                        Preconditions.checkArgument(
```
yunfengzhou-hub, line 105: There might be performance issues if we perform these checks for each record. Can we try to avoid this? For example, can the assembling process fail if the input data size does match with expected, so that no check is needed?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
 import org.apache.flink.ml.common.param.HasHandleInvalid;
 import org.apache.flink.ml.common.param.HasInputCols;
 import org.apache.flink.ml.common.param.HasOutputCol;
+import org.apache.flink.ml.param.IntArrayParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+
+import java.util.Arrays;
 
 /**
```
yunfengzhou-hub, line 30: Would it be better to add some document describing how users need to tackle the sizes parameter? For example, Spark has some descriptions of the function of VectorSizeHint in the JavaDoc of `VectorAssembler.handleInvalid`. Maybe we can add similar descriptions to the JavaDoc of `VectorAssembler`.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
weibozhao: Splits is an array, splitsArray is a `double[][]`.  VectorSizeHint has no parameter like splitsArray.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: I don't know which record has the error size, then I must check the sizes for every record. 
When the sizes are error, the code also can run OK, that's why we need to check every record.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,38 +76,65 @@ public Table[] transform(Table... inputs) {
...
         public void flatMap(Row value, Collector<Row> out) {
             int nnz = 0;
             int vectorSize = 0;
             try {
-                for (String inputCol : inputCols) {
+                for (int i = 0; i < inputCols.length; ++i) {
+                    String inputCol = inputCols[i];
                     Object object = value.getField(inputCol);
                     Preconditions.checkNotNull(object, "Input column value should not be null.");
                     if (object instanceof Number) {
```
yunfengzhou-hub, line 106: During offline discussion, I remembered that the purpose of adding the parameter `size` is to output a vector of size `sum(size)` when some input values might be null. If this is the case, this check and corresponding behavior in this algorithm needs be changed.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,38 +76,65 @@ public Table[] transform(Table... inputs) {
...
+                        Preconditions.checkArgument(
+                                sizeArray[i] == 1,
+                                "Inconsistent vector size, setSize is "
+                                        + sizeArray[i]
```
yunfengzhou-hub, line 110: String concatenation could cost more cpu time than assembling vectors, and we should try to avoid this unless necessary. For example, the following code might work much better.
```java
if (sizeArray[i] != 1) {
    throw new IllegalArgumentException("..." + "...");
}
```

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,38 +76,65 @@ public Table[] transform(Table... inputs) {
...
-                        vectorSize += ((SparseVector) object).size();
+                        vectorSize += sizeArray[i];
                     } else if (object instanceof DenseVector) {
+                        int localSize = ((DenseVector) object).size();
+                        Preconditions.checkArgument(
+                                sizeArray[i] == localSize,
+                                "Inconsistent vector size, setSize is "
+                                        + sizeArray[i]
+                                        + ", but current vector size is "
+                                        + localSize
```
yunfengzhou-hub, line 134: The three `checkArgument`s are almost the same and we may merge them into one common function.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -47,7 +47,9 @@
 
 /**
  * A Transformer which combines a given list of input columns into a vector column. Types of input
- * columns must be either vector or numerical value.
+ * columns must be either vector or numerical types. The elements assembled in the same column must
+ * have the same size. If the size of the element is not equal to sizes[columnIdx], it will throw an
+ * IllegalArgumentException.
```
yunfengzhou-hub, line 52: Do we need to throw exception when input element is null?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
...
+        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {
+    Param<Integer[]> SIZES =
+            new IntArrayParam(
+                    "sizeArray",
+                    "Sizes of the assembling elements.",
+                    null,
+                    ParamValidators.notNull());
+
+    default int[] getSizes() {
+        return Arrays.stream(get(SIZES)).mapToInt(Integer::intValue).toArray();
```
yunfengzhou-hub, line 45: nit: `ArrayUtils.toPrimitive`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
...
+    }
+
+    default T setSizes(Integer... value) {
```
yunfengzhou-hub, line 48: nit: `T setSizes(int... value)`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
yunfengzhou-hub, line 37: In Spark's equivalent algorithm, `getSize` has a clear context that it is about `VectorSizeHint`, or the size of the vector. But in Flink ML's VectorAssembler, there might be ambiguity for lacking in this context. Thus I would still prefer to add some delimiters to this parameter, like `setInputSizes`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
...
  * Params of {@link VectorAssembler}.
  *
  * @param <T> The class type of this instance.
  */
 public interface VectorAssemblerParams<T>
-        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {}
+        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {
+    Param<Integer[]> SIZES =
+            new IntArrayParam(
+                    "sizeArray",
```
yunfengzhou-hub, line 39: Let's keep this name consistent with the name of the parameter variable.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
...
+                    "Sizes of the assembling elements.",
```
yunfengzhou-hub, line 40: nit: "Sizes of the input elements to be assembled."

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,38 +76,65 @@ public Table[] transform(Table... inputs) {
...
+            this.sizeArray = sizeArray;
         }
 
         @Override
         public void flatMap(Row value, Collector<Row> out) {
             int nnz = 0;
             int vectorSize = 0;
             try {
-                for (String inputCol : inputCols) {
+                for (int i = 0; i < inputCols.length; ++i) {
```
yunfengzhou-hub, line 102: If `sizeArray.length > inputCols.length`, the code seems to still work, but an exception is expected.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,38 +76,65 @@ public Table[] transform(Table... inputs) {
...
     }
 
-    private static class AssemblerFunc implements FlatMapFunction<Row, Row> {
+    private static class AssemblerFunction implements FlatMapFunction<Row, Row> {
         private final String[] inputCols;
         private final String handleInvalid;
+        private final int[] sizeArray;
 
-        public AssemblerFunc(String[] inputCols, String handleInvalid) {
+        public AssemblerFunction(String[] inputCols, String handleInvalid, int[] sizeArray) {
```
yunfengzhou-hub, line 91: Let's either use `sizeArray` or `sizes` across all references to this parameter.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
weibozhao: this `nit` is not suited.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: If the job is a streaming job, I think passing this record is more proper.

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -27,8 +27,10 @@ under the License.
```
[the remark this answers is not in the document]
weibozhao: Discuss with some people, we need not add this parameter to these two algorithms.

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -44,11 +46,12 @@ Types of input columns must be either vector or numerical value.
...
-| inputCols     | `null`     | String[] | yes      | Input column names.                                                            |
-| outputCol     | `"output"` | String   | no       | Output column name.                                                            |
-| handleInvalid | `"error"`  | String   | no       | Strategy to handle invalid entries. Supported values: 'error', 'skip', 'keep'. |
+| Key             | Default    | Type      | Required | Description                                                                    |
+|-----------------|------------|-----------|----------|--------------------------------------------------------------------------------|
+| inputCols       | `null`     | String[]  | yes      | Input column names.                                                            |
+| outputCol       | `"output"` | String    | no       | Output column name.                                                            |
+| inputSizes      | `null`     | Integer[] | yes      | Sizes of the assembling elements.                                              |
```
yunfengzhou-hub, line 53: The type of this parameter should be `int[]` instead of `Integer[]`.

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -156,12 +160,13 @@ input_data_table = t_env.from_data_stream(
             [DenseVectorTypeInfo(), Types.DOUBLE(), SparseVectorTypeInfo()])))
 
 # create a vector assembler object and initialize its parameters
-vector_assembler = VectorAssembler() \
-    .set_input_cols('vec', 'num', 'sparse_vec') \
-    .set_output_col('assembled_vec') \
+vector_assembler = VectorAssembler()
+    .set_input_cols('vec', 'num', 'sparse_vec')
+    .set_output_col('assembled_vec')
+    .set_input_sizes(2, 1, 5)
```
yunfengzhou-hub, line 166: The example code in the document is not the same as that in `flink-ml-python/pyflink/examples` folder. For example, this line should be `.set_input_sizes(2, 1, 5) \` instead of `.set_input_sizes(2, 1, 5)`.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasVectorSize.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.IntParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared vector size. */
+public interface HasVectorSize<T> extends WithParams<T> {
```
yunfengzhou-hub, line 27: nit: This parameter is not used and can be deleted.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,31 @@
...
+                    "inputSizes",
+                    "Sizes of the input elements to be assembled.",
+                    null,
+                    ParamValidators.notNull());
+
+    default int[] getInputSizes() {
+        return ArrayUtils.toPrimitive(get(INPUT_SIZES));
+    }
+
+    default T setInputSizes(Integer... value) {
```
yunfengzhou-hub, line 48: nit: `default T setInputSizes(int... value)`

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
yunfengzhou-hub, line 52: It is determined by `HasHandleInvalid` whether to throw exceptions. Let's improve the JavaDoc to describe this behavior.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -133,6 +145,17 @@ public void flatMap(Row value, Collector<Row> out) {
                             : assembleSparse(inputCols, value, vectorSize, nnz);
             out.collect(Row.join(value, Row.of(assembledVec)));
         }
+
+        private void sizeHint(int expectedSize, int currentSize) {
```
yunfengzhou-hub, line 149: How about renaming this method to `checkVectorAndNumberSize()`? The current name is inherited from Spark's `VectorSizeHint` operator, so users may find this name hard to understand if they have not known Spark before.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -133,6 +145,17 @@ public void flatMap(Row value, Collector<Row> out) {
...
+            if (currentSize != expectedSize) {
+                throw new IllegalArgumentException(
+                        "Inconsistent vector size, setSize is "
```
yunfengzhou-hub, line 152: How about the following error message?
```java
throw new IllegalArgumentException(
        String.format(
                "Input vector/number size does not meet with expected. Expected size: %d, actual size: %s",
                expectedSize, currentSize));
```

yunfengzhou-hub (commented):

docs/content/docs/operators/feature/vectorassembler.md
weibozhao: I think the integer is much better.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -47,10 +47,15 @@
 
 /**
  * A Transformer which combines a given list of input columns into a vector column. Types of input
- * columns must be either vector or numerical value.
+ * columns must be either vector or numerical types. The elements assembled in the same column must
+ * have the same size. If the element is null or has the wrong size, we will process this case with
```
yunfengzhou-hub, line 51: Let's avoid writing the JavaDoc using the first-person perspective. For example, instead of saying "we will process", let's say "the operator deals with null values or records with wrong sizes according to the strategy specified by the {@link HasHandleInvalid} parameter as follows".

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -47,10 +47,15 @@
...
+ * {@link HasHandleInvalid} parameter as follows:
  *
- * <p>The `keep` option of {@link HasHandleInvalid} means that we output bad rows with output column
- * set to null.
+ * <p>The `keep` option means that we do not check the vector size, and keep all rows.
```
yunfengzhou-hub, line 54: If the size does not match the expected, what would be the output value of the row?

flink-ml-python/pyflink/ml/lib/feature/vectorassembler.py
```diff
@@ -31,17 +33,38 @@ class _VectorAssemblerParams(
     Params for :class:`VectorAssembler`.
     """
 
+    INPUT_SIZES: Param[Tuple[int, ...]] = IntArrayParam(
+        "input_sizes",
+        "Sizes of the assembling elements.",
```
yunfengzhou-hub, line 38: Let's keep the description the same as that in Java.

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -44,11 +50,12 @@ Types of input columns must be either vector or numerical value.
...
-| inputCols     | `null`     | String[] | yes      | Input column names.                                                            |
-| outputCol     | `"output"` | String   | no       | Output column name.                                                            |
-| handleInvalid | `"error"`  | String   | no       | Strategy to handle invalid entries. Supported values: 'error', 'skip', 'keep'. |
+| Key             | Default    | Type      | Required | Description                                                                    |
+|-----------------|------------|-----------|----------|--------------------------------------------------------------------------------|
+| inputCols       | `null`     | String[]  | yes      | Input column names.                                                            |
+| outputCol       | `"output"` | String    | no       | Output column name.                                                            |
+| inputSizes      | `null`     | Integer[] | yes      | Sizes of the assembling elements.                                              |
```
yunfengzhou-hub, line 57: Let's keep the description the same as that in Java.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: The output value will have a different vector size with the normal ones.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -47,10 +47,15 @@
...
 /**
  * A Transformer which combines a given list of input columns into a vector column. Types of input
- * columns must be either vector or numerical value.
+ * columns must be either vector or numerical types. The elements assembled in the same column must
+ * have the same size. The operator deals with null values or records with wrong sizes according to
+ * the strategy specified by the {@link HasHandleInvalid} parameter as follows:
  *
- * <p>The `keep` option of {@link HasHandleInvalid} means that we output bad rows with output column
- * set to null.
+ * <p>The `keep` option means that we do the assembling action without checking the vector size.
```
yunfengzhou-hub, line 54: Let's also remove the "we"s used here, making sure that the first-person perspective is not used anywhere in the documents.

yunfengzhou-hub, line 54: Could you please add a detailed description of the behavior when "keep" option is set? As I can think of, any of the following might be a correct behavior.
- assembling `[1,2]`, `[3,4]` with sizes `2,3` may get `[1, 2, 3, 4]`
- assembling `[1,2]`, `[3,4]` with sizes `2,3` may get `[1, 2, 3, 4, 0]` (padding with zeros)
- assembling `[1,2]`, `[3,4,5]` with sizes `2,2` may get `[1, 2, 3, 4]` (trim to fit size)
- assembling `[1,2]`, `[3,4,5]` with sizes `2,2` may get `[1, 2, NaN, NaN]`
- assembling `[1,2]`, `null` with sizes `2,2` may get `[1, 2, NaN, NaN]`
- assembling `[1,2]`, `null` with sizes `2,2` may get `[1, 2, 0, 0]`

Let's check the option selected by Spark, implement that behavior, and clarify this behavior in JavaDoc.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,29 @@
...
  */
 public interface VectorAssemblerParams<T>
-        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {}
+        extends HasInputCols<T>, HasOutputCol<T>, HasHandleInvalid<T> {
+    Param<Integer[]> INPUT_SIZES =
+            new IntArrayParam(
+                    "inputSizes",
+                    "Sizes of the input elements to be assembled.",
+                    null,
+                    ParamValidators.notNull());
```
yunfengzhou-hub, line 40: In Spark, VectorAssembler can infer the vector sizes in some cases, which means VectorSizeHint is not a compulsory prerequisite. Let's also support those situations in Flink ML, and remove the `ParamValidators.notNull()`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -47,10 +47,15 @@
 
 /**
  * A Transformer which combines a given list of input columns into a vector column. Types of input
- * columns must be either vector or numerical value.
+ * columns must be either vector or numerical types. The elements assembled in the same column must
+ * have the same size. The operator deals with null values or records with wrong sizes according to
```
yunfengzhou-hub, line 51: The statement that "the elements must have the same size" seems not accurate, as elements can be null or vectors of different sizes when handleInvalid is set to keep.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
weibozhao: assembling [1,2], [3,4,5] with sizes 2,2 may get [1, 2, 3, 4] (trim to fit size)
assembling [1,2], [3,4,5] with sizes 2,2 may get [1, 2, NaN, NaN]
assembling [1,2], null with sizes 2,2 may get [1, 2, NaN, NaN]
assembling [1,2], null with sizes 2,2 may get [1, 2, 0, 0]

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -45,13 +45,16 @@
```
yunfengzhou-hub, line 45: Let's further add test cases to verify different handleInvalid strategies when the input vectors are valid (no null or NaN values), but the size of the vectors does not match with that set in `inputSizes`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -47,10 +47,16 @@
...
+ * follows:
  *
- * <p>The `keep` option of {@link HasHandleInvalid} means that we output bad rows with output column
- * set to null.
+ * <p>The `keep` option means that if the input column data is NaN, then it keeps this value and if
+ * data is null vector, then uses a NaN vector to replace it.
+ *
+ * <p>The `skip` option means that it filters out rows with invalid elements.
+ *
+ * <p>The `error` option means that it throws an error exception when meeting some invalid data.
```
yunfengzhou-hub, line 59: The following tags could better format the JavaDoc.
```html
<ul>
    <li> AAA
    <li> BBB
    <li> CCC
</ul>
```

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,44 @@
...
+
+    default T setInputSizes(Integer... value) {
+        return set(INPUT_SIZES, value);
+    }
+
+    // Checks the inputSizes parameter.
+    static ParamValidator<Integer[]> sizesValidator() {
+        return inputSizes -> {
+            if (inputSizes == null) {
+                return false;
```
yunfengzhou-hub, line 54: In Spark, VectorAssembler can infer the vector sizes in some cases, which means VectorSizeHint is not a compulsory prerequisite. Let's also support those situations in Flink ML.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,64 +81,96 @@ public Table[] transform(Table... inputs) {
...
                         nnz += ((SparseVector) object).indices.length;
-                        vectorSize += ((SparseVector) object).size();
                     } else if (object instanceof DenseVector) {
+                        int localSize = ((DenseVector) object).size();
+                        checkVectorAndNumberSize(inputSizes[i], localSize);
                         nnz += ((DenseVector) object).size();
-                        vectorSize += ((DenseVector) object).size();
                     } else {
                         throw new IllegalArgumentException(
                                 "Input type has not been supported yet.");
```
yunfengzhou-hub, line 149: Let's add descriptions about what is the type of input value that has not been supported yet.

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -27,8 +27,13 @@ under the License.
 
 ## Vector Assembler
 
-Vector Assembler combines a given list of input columns into a vector column.
-Types of input columns must be either vector or numerical value.
+A Transformer which combines a given list of input columns into a vector column. Types of input
+columns must be either vector or numerical types. If the element is null or has the wrong size, 
+we will process this case with {@link HasHandleInvalid} parameter as follows:
+
+ * The `keep` option means that if the input column data is NaN, then it keeps this value and if data is null vector, then uses a NaN vector to replace it.
```
yunfengzhou-hub, line 34: Let's wrap the lines so that each line has no more than 80 characters.

yunfengzhou-hub (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
weibozhao: I have add this case. If input vectors are valid (no null or NaN values), but the size of the vectors does not match with that set in inputSizes, the algorithm will throw Exception all the time. Because this err maybe data error or parameter error, I don't know which one is.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
weibozhao: We have discuss this case already. If record-1 and record-2 has different sizes, then we can't decide which one is used. For spark stores the sizes in meta info, but flink-ml has no meta info.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
weibozhao: I have refine code as spark's functions.

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -156,12 +168,13 @@ input_data_table = t_env.from_data_stream(
             [DenseVectorTypeInfo(), Types.DOUBLE(), SparseVectorTypeInfo()])))
 
 # create a vector assembler object and initialize its parameters
-vector_assembler = VectorAssembler() \
-    .set_input_cols('vec', 'num', 'sparse_vec') \
-    .set_output_col('assembled_vec') \
+vector_assembler = VectorAssembler()\
+    .set_input_cols('vec', 'num', 'sparse_vec')\
+    .set_output_col('assembled_vec')\
+    .set_input_sizes(2, 1, 5)\
```
yunfengzhou-hub, line 174: It might be better to change `)\` to `) \`, keeping the same as python example file.

flink-ml-python/pyflink/ml/lib/feature/vectorassembler.py
```diff
@@ -27,21 +29,61 @@ class _VectorAssemblerParams(
...
+    """
+    Checks the inputSizes parameter.
+    """
+    def sizes_validator(self) -> ParamValidator[Tuple[int]]:
+        class SizesValidator(ParamValidator[Tuple[int]]):
+            def validate(self, indices: Tuple[int]) -> bool:
+                if indices is None:
+                    return False
+                for val in indices:
+                    if val < 0:
```
yunfengzhou-hub, line 42: Can the values be 0?

flink-ml-python/pyflink/ml/lib/feature/vectorassembler.py
```diff
@@ -27,21 +29,61 @@ class _VectorAssemblerParams(
...
+
     """
     Params for :class:`VectorAssembler`.
     """
 
+    INPUT_SIZES: Param[Tuple[int, ...]] = IntArrayParam(
+        "input_sizes",
+        "Sizes of the input elements to be assembled.",
+        None,
+        sizes_validator(None))
```
yunfengzhou-hub, line 55: Would it be enough to use `SizesValidator()` directly, instead of adding a `sizes_validator` method?

flink-ml-python/pyflink/ml/lib/feature/vectorassembler.py
```diff
@@ -27,21 +29,61 @@ class _VectorAssemblerParams(
...
     """
-    A Transformer which combines a given list of input columns into a vector column. Types of
-    input columns must be either vector or numerical value.
-
-    The `keep` option of :class:HasHandleInvalid means that we output bad rows with output column
-    set to null.
+     A Transformer which combines a given list of input columns into a vector column. Input columns
+     would be numerics or vectors whose size is specified by the {@link #INPUT_SIZES} parameter.
+     Invalid input data with null values or values with wrong sizes would be dealt with according to
+     the strategy specified by the {@link HasHandleInvalid} parameter as follows:
```
yunfengzhou-hub, line 76: `{@link ...}` is JavaDoc grammar and unsuitable for python documents. Let's change it to the way it has originally used, like `:class:HasHandleInvalid`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/VectorAssemblerTest.java
```diff
@@ -45,13 +45,15 @@
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
 
 /** Tests VectorAssembler. */
 public class VectorAssemblerTest extends AbstractTestBase {
 
     private StreamTableEnvironment tEnv;
+    private StreamExecutionEnvironment env;
```
yunfengzhou-hub, line 53: It is enough to keep this variable as local variable as it has been.

yunfengzhou-hub (commented):
Thanks for the update, left some comments as below.

@zhipeng93 Could you please take a look as this PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssemblerParams.java
```diff
@@ -21,11 +21,44 @@
...
+            }
+            for (Integer ele : inputSizes) {
+                if (ele < 0) {
```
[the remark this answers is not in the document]
weibozhao: If size of element equals to 0, return false as `return inputSizes.length != 0;`

docs/content/docs/operators/feature/vectorassembler.md
```diff
@@ -26,10 +26,20 @@ under the License.
 -->
 
 ## Vector Assembler
-
-Vector Assembler combines a given list of input columns into a vector column.
-Types of input columns must be either vector or numerical value.
-
+A Transformer which combines a given list of input columns into a vector column. Input columns
+would be numerics or vectors whose size is specified by the {@link #INPUT_SIZES} parameter.
```
zhipeng93, line 30: nit: `Input columns would be numerics` --> `Input Columns could be numerical or vectors whose sizes are specified by...`

Same for the java docs in python and java source code.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/vectorassembler/VectorAssembler.java
```diff
@@ -74,64 +85,109 @@ public Table[] transform(Table... inputs) {
...
-                    if (object instanceof Number) {
-                        nnz += 1;
-                        vectorSize += 1;
-                    } else if (object instanceof SparseVector) {
-                        nnz += ((SparseVector) object).indices.length;
-                        vectorSize += ((SparseVector) object).size();
-                    } else if (object instanceof DenseVector) {
-                        nnz += ((DenseVector) object).size();
-                        vectorSize += ((DenseVector) object).size();
+                for (int i = 0; i < inputCols.length; ++i) {
```
zhipeng93, line 113: This logic is a bit messy here. How about we split it into several functions? e.g.,
`assembleNullValues()` and `assembleNotNullValues()`?

zhipeng93 (commented):
Thanks for the update. I left two comments below.

PR #162, opened 2022-10-12
[FLINK-29593] Add QuantileSummary to help calculate approximate quantiles

jiangxin369:

Add QuantileSummary to help calculate approximate quantiles

## Brief change log

  - Add implementation of GK algorithm.
  - Add merge operation to combine results of distributed datasets.
  - Remove redundant Internal annotations from BroadcastUtils

## Documentation

  - Does this pull request introduce a new feature? (yes )

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
@@ -1,0 +1,414 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.common.util;
+
+import org.apache.flink.util.Preconditions;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Helper class to compute an approximate quantile summary. This implementation is based on the
+ * algorithm proposed in the paper: "Space-efficient Online Computation of Quantile Summaries" by
+ * Greenwald, Michael and Khanna, Sanjeev. (https://doi.org/10.1145/375663.375670)
+ */
+public class QuantileSummary implements Serializable {
+
+    /** The default size of head buffer. */
+    private static final int DEFAULT_HEAD_SIZE = 50000;
+
+    /** The default compression threshold. */
+    private static final int DEFAULT_COMPRESS_THRESHOLD = 10000;
+
+    /** The target relative error. */
+    private final double relativeError;
+
+    /**
+     * The compression threshold. After the internal buffer of statistics crosses this size, it
+     * attempts to compress the statistics together.
+     */
+    private final int compressThreshold;
+
+    /** The count of all the elements inserted to be calculated. */
+    private final long count;
+
+    /** A buffer of quantile statistics. */
+    private final List<StatsTuple> sampled;
+
+    /** A buffer of the latest samples seen so far. */
+    private final List<Double> headBuffer = new ArrayList<>();
+
+    /** Whether the quantile summary has been compressed. */
+    private boolean compressed;
+
+    /**
+     * QuantileSummary Constructor.
+     *
+     * @param relativeError The target relative error.
+     */
+    public QuantileSummary(double relativeError) {
+        this(relativeError, DEFAULT_COMPRESS_THRESHOLD);
+    }
+
+    /**
+     * QuantileSummary Constructor.
+     *
+     * @param relativeError The target relative error.
+     * @param compressThreshold the compression threshold. After the internal buffer of statistics
+     *     crosses this size, it attempts to compress the statistics together.
+     */
+    @SuppressWarnings("unchecked")
+    public QuantileSummary(double relativeError, int compressThreshold) {
+        this(relativeError, compressThreshold, Collections.EMPTY_LIST, 0, false);
+    }
+
+    /**
+     * QuantileSummary Constructor.
+     *
+     * @param relativeError The target relative error.
+     * @param compressThreshold the compression threshold.
+     * @param sampled A buffer of quantile statistics. See the G-K article for more details.
+     * @param count The count of all the elements inserted in the sampled buffer.
+     * @param compressed Whether the statistics have been compressed.
+     */
+    private QuantileSummary(
+            double relativeError,
+            int compressThreshold,
+            List<StatsTuple> sampled,
+            long count,
+            boolean compressed) {
+        Preconditions.checkArgument(
+                relativeError > 0 && relativeError < 1,
+                "An appropriate relative error must lay between 0 and 1.");
+        Preconditions.checkArgument(
+                compressThreshold > 0, "An compress threshold must greater than 0.");
+        this.relativeError = relativeError;
+        this.compressThreshold = compressThreshold;
+        this.sampled = sampled;
+        this.count = count;
+        this.compressed = compressed;
+    }
+
+    /**
+     * Insert a new observation into the summary.
+     *
+     * @param item The new observation to insert into the summary.
+     * @return A summary with the given observation inserted into the summary.
+     */
+    public QuantileSummary insert(double item) {
+        headBuffer.add(item);
+        compressed = false;
+        if (headBuffer.size() >= DEFAULT_HEAD_SIZE) {
+            QuantileSummary result = insertHeadBuffer();
+            if (result.sampled.size() >= compressThreshold) {
+                return result.compress();
+            } else {
+                return result;
+            }
+        } else {
+            return this;
+        }
+    }
+
+    /**
+     * Returns a new summary that compresses the summary statistics and the head buffer.
+     *
+     * <p>This implements the COMPRESS function of the GK algorithm.
+     *
+     * @return The compressed summary.
+     */
+    public QuantileSummary compress() {
+        if (compressed) {
+            return this;
+        }
+        QuantileSummary inserted = insertHeadBuffer();
+        Preconditions.checkState(inserted.headBuffer.isEmpty());
+        Preconditions.checkState(inserted.count == count + headBuffer.size());
+
+        List<StatsTuple> compressed =
+                compressInternal(inserted.sampled, 2 * relativeError * inserted.count);
+        return new QuantileSummary(
+                relativeError, compressThreshold, compressed, inserted.count, true);
+    }
+
+    /**
+     * Merges two summaries together.
+     *
+     * @param other The summary to be merged.
+     * @return The merged summary.
+     */
+    public QuantileSummary merge(QuantileSummary other) {
+        Preconditions.checkState(
+                headBuffer.isEmpty(), "Current buffer needs to be compressed before merge.");
+        Preconditions.checkState(
+                other.headBuffer.isEmpty(), "Other buffer needs to be compressed before merge.");
+
+        if (other.count == 0) {
+            return shallowCopy();
+        } else if (count == 0) {
+            return other.shallowCopy();
+        } else {
+            List<StatsTuple> mergedSampled = new ArrayList<>();
+            double mergedRelativeError = Math.max(relativeError, other.relativeError);
+            long mergedCount = count + other.count;
+            long additionalSelfDelta =
+                    Double.valueOf(Math.floor(2 * other.relativeError * other.count)).longValue();
+            long additionalOtherDelta =
+                    Double.valueOf(Math.floor(2 * relativeError * count)).longValue();
+
+            int selfIdx = 0;
+            int otherIdx = 0;
+            while (selfIdx < sampled.size() && otherIdx < other.sampled.size()) {
+                StatsTuple selfSample = sampled.get(selfIdx);
+                StatsTuple otherSample = other.sampled.get(otherIdx);
+                StatsTuple nextSample;
+                long additionalDelta = 0;
+                if (selfSample.value < otherSample.value) {
+                    nextSample = selfSample;
+                    if (otherIdx > 0) {
+                        additionalDelta = additionalSelfDelta;
+                    }
+                    selfIdx++;
+                } else {
+                    nextSample = otherSample;
+                    if (selfIdx > 0) {
+                        additionalDelta = additionalOtherDelta;
+                    }
+                    otherIdx++;
+                }
+                nextSample = nextSample.shallowCopy();
+                nextSample.delta = nextSample.delta + additionalDelta;
+                mergedSampled.add(nextSample);
+            }
+            IntStream.range(selfIdx, sampled.size())
+                    .forEach(i -> mergedSampled.add(sampled.get(i)));
+            IntStream.range(otherIdx, other.sampled.size())
+                    .forEach(i -> mergedSampled.add(other.sampled.get(i)));
+
+            List<StatsTuple> comp =
+                    compressInternal(mergedSampled, 2 * mergedRelativeError * mergedCount);
+            return new QuantileSummary(
+                    mergedRelativeError, compressThreshold, comp, mergedCount, true);
+        }
+    }
+
+    /**
+     * Runs a query for a given percentile. The query can only be run on a compressed summary, you
+     * need to call compress() before using it.
+     *
+     * @param percentile The target percentile.
+     * @return The corresponding approximate quantile.
+     */
+    public double query(double percentile) {
+        return query(new double[] {percentile})[0];
+    }
+
+    /**
+     * Runs a query for a given sequence of percentiles. The query can only be run on a compressed
+     * summary, you need to call compress() before using it.
+     *
+     * @param percentiles A list of the target percentiles.
+     * @return A list of the corresponding approximate quantiles, in the same order as the input.
+     */
+    public double[] query(double[] percentiles) {
+        Arrays.stream(percentiles)
+                .forEach(
+                        x ->
+                                Preconditions.checkState(
+                                        x >= 0 && x <= 1.0,
+                                        "percentile should be in the range [0.0, 1.0]."));
+        Preconditions.checkState(
+                headBuffer.isEmpty(),
+                "Cannot operate on an uncompressed summary, call compress() first.");
+        Preconditions.checkState(
+                sampled != null && !sampled.isEmpty(),
+                "Cannot query percentiles without any records inserted.");
+        double targetError = Long.MIN_VALUE;
+        for (StatsTuple tuple : sampled) {
+            targetError = Math.max(targetError, (tuple.delta + tuple.g));
+        }
+        targetError = targetError / 2;
+        Map<Double, Integer> zipWithIndex = new HashMap<>(percentiles.length);
+        IntStream.range(0, percentiles.length).forEach(i -> zipWithIndex.put(percentiles[i], i));
+
+        int index = 0;
+        long minRank = sampled.get(0).g;
+        double[] sorted = Arrays.stream(percentiles).sorted().toArray();
+        double[] result = new double[percentiles.length];
+
+        for (double item : sorted) {
+            int percentileIndex = zipWithIndex.get(item);
+            if (item <= relativeError) {
+                result[percentileIndex] = sampled.get(0).value;
+            } else if (item >= 1 - relativeError) {
+                result[percentileIndex] = sampled.get(sampled.size() - 1).value;
+            } else {
+                QueryResult queryResult =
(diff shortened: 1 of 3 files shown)


flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+ * algorithm proposed in the paper: "Space-efficient Online Computation of Quantile Summaries" by
+ * Greenwald, Michael and Khanna, Sanjeev. (https://doi.org/10.1145/375663.375670)
+ */
+public class QuantileSummary {
+
+    private static final int BUFFER_SIZE = 10000;
+    private static final int COMPRESS_THRESHOLD = 10000;
+
+    private double relativeError;
+    private long count;
```
zhipeng93, line 47: Could you add explanations for the class variable here for better readability?

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+import static org.apache.flink.util.Preconditions.checkState;
+
+/**
+ * Helper class to compute an approximate quantile summary. This implementation is based on the
+ * algorithm proposed in the paper: "Space-efficient Online Computation of Quantile Summaries" by
+ * Greenwald, Michael and Khanna, Sanjeev. (https://doi.org/10.1145/375663.375670)
+ */
+public class QuantileSummary {
+
+    private static final int BUFFER_SIZE = 10000;
```
zhipeng93, line 43: How is the default value for `BUFFER_SIZE` chosen? It seems that the corresponding value in Spark is 5000. [1]

By the way, `defaultHeadSize` seems a better name for this.

[1] https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/QuantileSummaries.scala#L338

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    private static final int COMPRESS_THRESHOLD = 10000;
```
zhipeng93, line 44: Should we also let `COMPRESS_THRESHOLD` a non-final value and can be set by users?
Since it will also be used by LSH. [1]

[1] https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/feature/LSH.scala#L143

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    private long count;
+    private List<StatsTuple> sampled;
+    private List<Double> headBuffer = new ArrayList<>(BUFFER_SIZE);
+
+    /**
+     * * QuantileSummaries Constructor.
+     *
+     * @param relativeError The target relative error.
+     */
+    public QuantileSummary(double relativeError) {
```
zhipeng93, line 56: As from LSH and RobustScaler, the constructor needs to take both `relativeError` and `compressThreshold` as input parameters.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    private static final int BUFFER_SIZE = 10000;
+    private static final int COMPRESS_THRESHOLD = 10000;
+
+    private double relativeError;
+    private long count;
+    private List<StatsTuple> sampled;
+    private List<Double> headBuffer = new ArrayList<>(BUFFER_SIZE);
+
+    /**
+     * * QuantileSummaries Constructor.
```
zhipeng93, line 52: nit: The comment is in-consistent with the class name.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /**
+     * QuantileSummaries Constructor.
+     *
+     * @param relativeError The target relative error.
+     * @param sampled A buffer of quantile statistics. See the G-K article for more details.
+     * @param count The count of all the elements inserted in the sampled buffer.
+     */
+    public QuantileSummary(double relativeError, List<StatsTuple> sampled, long count) {
```
zhipeng93, line 67: It seems that this method not necessary for LSH or RobustScaler. Can we make it private for now?

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /**
+     * * Merges two compressed summaries together.
+     *
+     * @param other The summary to be merged.
+     * @return The merged summary.
+     */
+    public QuantileSummary merge(QuantileSummary other) {
+        checkState(headBuffer.isEmpty(), "Current buffer needs to be compressed before merge");
```
zhipeng93, line 118: nit: The usage of `checkState` and `Preconditions.checkState` seems inconsistent in this PR.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    /**
+     * Runs a query for a given percentile. The query can only be run on a compressed summary, you
+     * need to call compress() before using it.
+     *
+     * @param percentile The target percentile.
+     * @return The corresponding approximate quantile.
+     */
+    public double query(double percentile) {
+        double[] percentiles = {percentile};
+        return query(percentiles)[0];
```
zhipeng93, line 178: If the `sample` is null, it may throw NPE here.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+import java.util.stream.IntStream;
+
+import static org.apache.flink.util.Preconditions.checkState;
+
+/**
+ * Helper class to compute an approximate quantile summary. This implementation is based on the
+ * algorithm proposed in the paper: "Space-efficient Online Computation of Quantile Summaries" by
+ * Greenwald, Michael and Khanna, Sanjeev. (https://doi.org/10.1145/375663.375670)
+ */
+public class QuantileSummary {
```
zhipeng93, line 41: Should we mark it as Internal or do we expect end users to use this class?

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * This implements the COMPRESS function of the GK algorithm.
+     *
+     * @return A new summary that compresses the summary statistics and the head buffer.
+     */
+    public QuantileSummary compress() {
+        QuantileSummary inserted = insertHeadBuffer();
+        Preconditions.checkState(inserted.headBuffer.isEmpty());
+        Preconditions.checkState(inserted.count == count + headBuffer.size());
```
zhipeng93, line 104: Is `Preconditions.checkState(inserted.count == count)` enough since headBuffer is empty?

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,359 @@
...
+    public void setRelativeError(double relativeError) {
+        this.relativeError = relativeError;
+    }
+
+    private QuantileSummary shallowCopy() {
+        return new QuantileSummary(relativeError, sampled, count);
+    }
+
+    /** Wrapper class to hold all information returned after querying. */
+    private class QueryResult {
```
zhipeng93, line 320: Could this be a static inner class? Same for `StatsTuple`.

zhipeng93 (commented):
Thanks for the PR! @jiangxin369  Left some comments below.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
jiangxin369: I will add another public constructor with both `relativeError` and `compressThreshold` but also keep the construct with only `relativeError` meanwhile, because in most cases users do not need to worry about the `compressThreshold`, so do LSH and RobustScaler.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
jiangxin369: Sure

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
jiangxin369: Maybe it should be marked as internal because the class is only expected to be used by algorithms, however, I didn't get what is the principle of marking a class as internal, some classes under package `utils` or `common` are not marked either.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
jiangxin369: It is `inserted.headBuffer` which is empty, not `headBuffer`, so I think the check statement is right.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
jiangxin369: I agree that `defaultHeadSize` is better. As for the default value, I think `10000` is enough instead of `50000` of Spark, because we don't want to allocate so much memory for each QuantileSummary.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
zhipeng93: Yeah, some of them seems not marked appropriately. Technically some of them should be marked as `Internal`. 

Can you mark this class as `Internal` in this PR? We may do others later.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
zhipeng93: Hmm.. Is there any technical reasons for choosing `10000` instead of `50000`? If we want to avoid OOM, then why don't you choose `1000`?

flink-ml-core/src/test/java/org/apache/flink/ml/util/QuantileSummaryTest.java
```diff
@@ -0,0 +1,135 @@
...
+                                        + Arrays.stream(data).filter(x -> x < approx).count())
+                                / 2.0);
+        double lower = Math.floor((percentile - summary.getRelativeError()) * data.length);
+        double upper = Math.ceil((percentile + summary.getRelativeError()) * data.length);
+        String errMessage =
+                String.format(
+                        "Rank not in [%s, %s], percentile: %s, approx returned: %s",
+                        lower, upper, percentile, approx);
+        assertTrue(errMessage, rank >= lower);
+        assertTrue(errMessage, rank <= upper);
```
yunfengzhou-hub, line 82: Would it be better to use the following method? `assertEquals(String errMessage, double expected, double actual, double epsilon)`

yunfengzhou-hub (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,391 @@
...
+        Preconditions.checkState(inserted.headBuffer.isEmpty());
+        Preconditions.checkState(inserted.count == count + headBuffer.size());
+
+        List<StatsTuple> compressed =
+                compressInternal(inserted.sampled, 2 * relativeError * inserted.count);
+        return new QuantileSummary(relativeError, compressThreshold, compressed, inserted.count);
+    }
+
+    /**
+     * Merges two compressed summaries together.
```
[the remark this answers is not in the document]
jiangxin369: Lets just remove the `compressed`, `Merges two summaries together` is enough.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,391 @@
...
+    /**
+     * Runs a query for a given percentile. The query can only be run on a compressed summary, you
+     * need to call compress() before using it.
+     *
+     * @param percentile The target percentile.
+     * @return The corresponding approximate quantile.
+     */
+    public double query(double percentile) {
+        double[] result = query(new double[] {percentile});
+        return result == null ? null : result[0];
```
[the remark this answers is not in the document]
jiangxin369: I will remove the `return null` from `double[] query(double[] percentiles)` method, so no NPE would produced.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
```diff
@@ -0,0 +1,391 @@
...
+        }
+    }
+
+    /**
+     * Wrapper class to hold all statistics from the Greenwald-Khanna paper. It contains the
+     * following information:
+     *
+     * <ul>
+     *   <li>value: the sampled value.
+     *   <li>g: the difference between the least rank of this element and the rank of the preceding
```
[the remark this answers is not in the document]
jiangxin369: I prefer to keep it there because it shows that these items are all semantics in the GK paper.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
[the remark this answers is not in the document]
jiangxin369: I'll add a `isEmpty` method.

flink-ml-core/src/main/java/org/apache/flink/ml/util/QuantileSummary.java
jiangxin369: Of cause `50000` is also acceptable, in that way each QuantileSummary needs `50000 * 8B + 10000 * 24B = 640k` memory, a TM with 2GB mem can process 3000 features, if exceeded, the parallelism needs to be expanded.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -0,0 +1,400 @@
...
+    private final List<StatsTuple> sampled;
+
+    /** The default size of head buffer. */
+    private static final int DEFAULT_HEAD_SIZE = 50000;
+
+    /** The default compression threshold. */
+    private static final int DEFAULT_COMPRESS_THRESHOLD = 10000;
+
+    /** A buffer of the latest samples seen so far. */
+    private List<Double> headBuffer = new ArrayList<>(DEFAULT_HEAD_SIZE);
```
zhipeng93, line 65: nit: Do you think making it a `double[]` could be more memory efficient? I am also fine the keep the current implementation.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -0,0 +1,400 @@
...
+        Preconditions.checkArgument(
+                compressThreshold > 0, "An compress threshold must greater than 0.");
+        this.relativeError = relativeError;
+        this.compressThreshold = compressThreshold;
+        this.sampled = sampled;
+        this.count = count;
+    }
+
+    /**
+     * Insert a new observation to the summary.
```
zhipeng93, line 109: nit: `Inserts a new observation into the summary`

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -0,0 +1,400 @@
...
+                index = queryResult.index;
+                minRank = queryResult.minRankAtIndex;
+                result[percentileIndex] = queryResult.percentile;
+            }
+        }
+        return result;
+    }
+
+    /**
+     * Check whether the QuantileSummary has inserted rows. Running query on an empty
```
zhipeng93, line 267: nit: `Checks`

zhipeng93 (commented):
Thanks for the PR. LGTM overall. I left some minor comments.

Do you want to take another look? @yunfengzhou-hub @lindong28

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -0,0 +1,400 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Helper class to compute an approximate quantile summary. This implementation is based on the
+ * algorithm proposed in the paper: "Space-efficient Online Computation of Quantile Summaries" by
+ * Greenwald, Michael and Khanna, Sanjeev. (https://doi.org/10.1145/375663.375670)
+ */
+@Internal
```
lindong28, line 40: According to the Java doc of `@Internal`, it is an "Annotation to mark methods within stable, public APIs as an internal developer API".

Thus it seems unnecessary to add this annotation.

@zhipeng93 Could you confirm that we should remove this annotation?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -0,0 +1,400 @@
...
+    private final int compressThreshold;
+
+    /** The count of all the elements inserted to be calculated. */
+    private final long count;
+
+    /** A buffer of quantile statistics. */
+    private final List<StatsTuple> sampled;
+
+    /** The default size of head buffer. */
+    private static final int DEFAULT_HEAD_SIZE = 50000;
```
lindong28, line 59: `private static final` variables are typically put above `private final` variables.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -0,0 +1,400 @@
...
+        return result;
+    }
+
+    /**
+     * Check whether the QuantileSummary has inserted rows. Running query on an empty
+     * QuantileSummary would cause {@link java.lang.IllegalStateException}.
+     *
+     * @return True if the QuantileSummary is empty, otherwise false.
+     */
+    public boolean isEmpty() {
```
lindong28, line 272: Should it be `private`?

lindong28 (commented):
Thanks for the update. Looks good overall. Left some minor comments.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
jiangxin369: It's a public API. Running query on an empty QuantileSummary is illegal and exceptions would be thrown. So we'd better provide a public API to check if the QuantileSummary is empty.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
zhipeng93: It is a developper API and is expected to be used in different algorithms. 

It is some how similar as we marked `withBroadcast` as `Internal`. [1]

[1] https://github.com/apache/flink-ml/commit/2edf453251c7ed9d8d1778c9c0115194e3d13ca7

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
jiangxin369: Indeed `double[]` is more memory efficient, but after repeated thinking I think leave it as an empty ArrayList is better, it is more friendly to small dataset, which is the same to Spark.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
lindong28: @jiangxin369 Since this is public API, should we add a test so that every public API's behavior can be covered by unit tests?

jiangxin369: Sure

jiangxin369:
Thanks for the review. Seems that all comments are resolved, could you help to merge this PR? @zhipeng93

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
lindong28: Is this comment addressed? QuantileSummaryTest does not seem to have any test that calls `IsEmpty()`, right?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
lindong28: @jiangxin369 If you need to update this PR, it might be simpler to also update `withBroadcast()` to follow the best practice of using `@Internal`.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
jiangxin369: Thanks for reminding me, I lost it.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
jiangxin369: Updated.

zhipeng93:
Thanks for the update. Merging.

PR #167, opened 2022-10-28
[FLINK-29786] VarianceThresholdSelector should implement HasInputCol

jiangxin369:

Replace `FeaturesCol` with `InputCol` in VarianceThresholdSelector.

## Brief change log

  - Add Javadoc for `HasFeaturesCol` to distinguish `HasInputCol`
  - Replace `FeaturesCol` with `InputCol` in VarianceThresholdSelector.

## Documentation

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasFeaturesCol.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasFeaturesCol.java
@@ -18,12 +18,18 @@
 
 package org.apache.flink.ml.common.param;
 
+import org.apache.flink.ml.api.Stage;
 import org.apache.flink.ml.param.Param;
 import org.apache.flink.ml.param.ParamValidators;
 import org.apache.flink.ml.param.StringParam;
 import org.apache.flink.ml.param.WithParams;
 
-/** Interface for the shared featuresCol param. */
+/**
+ * Interface for the shared featuresCol param.
+ *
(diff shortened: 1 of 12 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasFeaturesCol.java
```diff
@@ -24,6 +24,14 @@
 import org.apache.flink.ml.param.WithParams;
 
 /** Interface for the shared featuresCol param. */
+
+/**
+ * Interface for the shared featuresCol param.
```
lindong28, line 29: Should we remove the redundant description above?

And it might be simpler and softer to use the following doc:

```
/**
 * Interface for the shared featuresCol param.
 *
 * {@link HasFeaturesCol} is typically used for {@link Stage}s that implement {@link HasLabelCol}. It is preferred
 * to use {@link HasInputCol} for other cases.
 */
```

Please also update the corresponding python doc if we make the change here.

lindong28 (commented):
Thanks for the PR. LGTM overall. Left just one minor comment.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasFeaturesCol.java
jiangxin369: Thanks for the review, I updated the docs of Java and Python according to your expression.

lindong28:
Thanks for the update. LGTM.

PR #172, opened 2022-11-08
[FLINK-29592] Add Estimator and Transformer for RobustScaler

jiangxin369:

Add Estimator and Transformer for RobustScaler.

## Brief change log

  - Adds Transformer and Estimator implementation of RobustScaler in Java and Python
  - Adds examples and documentation of RobustScaler

## Documentation

### DIFF
--- a/docs/content/docs/operators/feature/robustscaler.md
+++ b/docs/content/docs/operators/feature/robustscaler.md
@@ -1,0 +1,211 @@
+---
+title: "Robust Scaler"
+weight: 1
+type: docs
+aliases:
+- /operators/feature/robustscaler.html
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Robust Scaler
+
+Robust Scaler is an algorithm that scales features using statistics that are
+robust to outliers.
+
+This Scaler removes the median and scales the data according to the quantile
+range (defaults to IQR: Interquartile Range). The IQR is the range between 
+the 1st quartile (25th quantile) and the 3rd quartile (75th quantile) but can
+be configured.
+
+Centering and scaling happen independently on each feature by computing the 
+relevant statistics on the samples in the training set. Median and quantile 
+range are then stored to be used on later data using the transform method.
+
+Standardization of a dataset is a common requirement for many machine learning
+estimators. Typically this is done by removing the mean and scaling to unit 
+variance. However, outliers can often influence the sample mean / variance 
+in a negative way. In such cases, the median and the interquartile range 
+often give better results.
+
+Note that NaN values are ignored in the computation of medians and ranges.
+
+### Input Columns
+
+| Param name | Type   | Default   | Description            |
+|:-----------|:-------|:----------|:-----------------------|
+| inputCol   | Vector | `"input"` | Features to be scaled. |
+
+### Output Columns
+
+| Param name | Type   | Default    | Description      |
+|:-----------|:-------|:-----------|:-----------------|
+| outputCol  | Vector | `"output"` | Scaled features. |
+
+### Parameters
+
+Below are the parameters required by `RobustScalerModel`.
+
+| Key           | Default    | Type        | Required | Description                                                           |
+|---------------|------------|-------------|----------|-----------------------------------------------------------------------|
+| inputCol      | `"input"`  | String      | no       | Input column name.                                                    |
+| outputCol     | `"output"` | String      | no       | Output column name.                                                   |
+| withCentering | `false`    | Boolean     | no       | Whether to center the data with median before scaling.                |
+| withScaling   | `true`     | Boolean     | no       | Whether to scale the data to quantile range.                          |
+
+`RobustScaler` needs parameters above and also below.
+
+| Key           | Default      | Type        | Required | Description                                                           |
+|---------------|--------------|-------------|----------|-----------------------------------------------------------------------|
+| lower         | `0.25`       | Double      | no       | Lower quantile to calculate quantile range.                           |
+| upper         | `0.75`       | Double      | no       | Upper quantile to calculate quantile range.                           |
+| relativeError | `0.001`      | Double      | no       | The relative target precision for the approximate quantile algorithm. |
+
+### Examples
+
+{{< tabs examples >}}
+
+{{< tab "Java">}}
+
+```java
+import org.apache.flink.ml.feature.robustscaler.RobustScaler;
+import org.apache.flink.ml.feature.robustscaler.RobustScalerModel;
+import org.apache.flink.ml.linalg.DenseVector;
+import org.apache.flink.ml.linalg.Vectors;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+
+/** Simple program that trains a {@link RobustScaler} model and uses it for feature selection. */
+public class RobustScalerExample {
+    public static void main(String[] args) {
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        StreamTableEnvironment tEnv = StreamTableEnvironment.create(env);
+
+        // Generates input training and prediction data.
+        DataStream<Row> trainStream =
+                env.fromElements(
+                        Row.of(1, Vectors.dense(0.0, 0.0)),
+                        Row.of(2, Vectors.dense(1.0, -1.0)),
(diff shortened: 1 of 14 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScaler.java
```diff
@@ -0,0 +1,183 @@
...
+ * the 3rd quartile (75th quantile) but can be configured.
+ *
+ * <p>Centering and scaling happen independently on each feature by computing the relevant
+ * statistics on the samples in the training set. Median and quantile range are then stored to be
+ * used on later data using the transform method.
+ *
+ * <p>Standardization of a dataset is a common requirement for many machine learning estimators.
+ * Typically this is done by removing the mean and scaling to unit variance. However, outliers can
+ * often influence the sample mean / variance in a negative way. In such cases, the median and the
+ * interquartile range often give better results.
```
yunfengzhou-hub, line 58: I guess "range" should be "ranges". Let's also check if there are other grammar errors.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScaler.java
```diff
@@ -0,0 +1,183 @@
...
+                        inputData,
+                        new QuantileAggregator(getRelativeError(), getLower(), getUpper()));
+        RobustScalerModel model =
+                new RobustScalerModel().setModelData(tEnv.fromDataStream(modelData));
+        ReadWriteUtils.updateExistingParams(model, getParamMap());
+        return model;
+    }
+
+    /**
+     * A stream operator to compute the medians and quantile ranges from feature column of the input
```
yunfengzhou-hub, line 90: It might be better to change "stream operator" to "function" or "aggregate function".

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScalerModel.java
```diff
@@ -0,0 +1,179 @@
...
+        public Row map(Row row) throws Exception {
+            if (medians == null) {
+                RobustScalerModelData modelData =
+                        (RobustScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastModelKey).get(0);
+                medians = modelData.medians;
+                scales =
+                        new DenseVector(
+                                Arrays.stream(modelData.ranges.values)
+                                        .map(range -> range == 0 ? 1 : 1 / range)
```
yunfengzhou-hub, line 125: Could you please explain why this PR chooses to make scale  = 1 when range == 0? Spark seems to make scale = 0 in this case.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/RobustScalerTest.java
```diff
@@ -0,0 +1,295 @@
...
+        assertFalse(robustScaler.getWithCentering());
+        assertTrue(robustScaler.getWithScaling());
+
+        robustScaler
+                .setInputCol("test_input")
+                .setOutputCol("test_output")
+                .setLower(0.1)
+                .setUpper(0.9)
+                .setRelativeError(0.01)
+                .setWithCentering(false)
```
yunfengzhou-hub, line 135: It might be better to set it to a non-default value.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScaler.java
jiangxin369: Sorry, I don't think it's a grammar error, what's the reason?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScalerModel.java
jiangxin369: IMO, if set scale = 1, the model would do nothing to the original data which is chosen by sklearn, but if set scale = 0, all records would be replaced by 0 which makes this feature useless.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScaler.java
yunfengzhou-hub: Sorry that I mistook the meaning as "the median range and the interquartile range". I agree that there is no grammar error now.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/RobustScalerTest.java
```diff
@@ -0,0 +1,318 @@
...
+                Arrays.asList("medians", "ranges"), modelData.getResolvedSchema().getColumnNames());
+        DataStream<Row> output = tEnv.toDataStream(modelData);
+        List<Row> modelRows = IteratorUtils.toList(output.executeAndCollect());
+        DenseVector medians = (DenseVector) modelRows.get(0).getField(0);
+        DenseVector ranges = (DenseVector) modelRows.get(0).getField(1);
+
+        double[] expectedMedians = {4.0, -4.0};
+        double[] expectedRanges = {4.0, 4.0};
+        for (int i = 0; i < expectedMedians.length; i++) {
+            assertEquals(expectedMedians[i], medians.get(i), EPS);
```
yunfengzhou-hub, line 303: `assertArrayEquals` might be better.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScalerModel.java
```diff
@@ -0,0 +1,179 @@
...
+                        (RobustScalerModelData)
+                                getRuntimeContext().getBroadcastVariable(broadcastModelKey).get(0);
+                medians = modelData.medians;
+                scales =
+                        new DenseVector(
+                                Arrays.stream(modelData.ranges.values)
+                                        .map(range -> range == 0 ? 0 : 1 / range)
+                                        .toArray());
+            }
+            DenseVector outputVec = ((Vector) row.getField(inputCol)).clone().toDense();
```
yunfengzhou-hub, line 128: `toDense()` might be redundant, as it is supposed that `BLAS.axpy` and `BLAS.hDot` can handle sparse vectors properly.

flink-ml-python/pyflink/ml/lib/feature/tests/test_robustscaler.py
```diff
@@ -0,0 +1,135 @@
...
+        self.assertEqual(0.01, robust_scaler.relative_error)
+        self.assertTrue(robust_scaler.with_centering)
+        self.assertFalse(robust_scaler.with_scaling)
+
+    def test_output_schema(self):
+        robust_scaler = RobustScaler()
+        model = robust_scaler.fit(self.train_table)
+        output = model.transform(self.predict_table.alias('test_input'))[0]
+        self.assertEqual(
+            ['test_input', 'output'],
```
yunfengzhou-hub, line 98: Let's test non-default output column names as well, and use `set_input_col` and `set_output_col` to specify the non-default values.

flink-ml-python/pyflink/examples/ml/feature/robustscaler_example.py
```diff
@@ -0,0 +1,74 @@
...
+    .set_with_scaling(True)\
+    .set_with_centering(True)
+
+# Trains the RobustScaler Model.
+model = robust_scaler.fit(train_data)
+
+# Uses the RobustScaler Model for predictions.
+output = model.transform(train_data)[0]
+
+# Extracts and displays the results.
```
yunfengzhou-hub, line 68: Comments in python examples uses the original form of the verbs and do not use uppercase for the first letter.

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScalerModel.java
jiangxin369: `BLAS.axpy` requires DenseVector.

docs/content/docs/operators/feature/robustscaler.md
```diff
@@ -0,0 +1,208 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## Robust Scaler
+
+An Estimator which scales features using statistics that are robust to outliers.
```
yunfengzhou-hub, line 30: The description here should stand for the algorithm as a whole, including both `RobustScaler` and `RobustScalerModel`. Thus it might be improper to say "Estimator" here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScalerModel.java
```diff
@@ -0,0 +1,179 @@
...
+            Preconditions.checkState(
+                    medians.size() == outputVec.size(),
+                    "Number of features must be %s but got %s.",
+                    medians.size(),
+                    outputVec.size());
+
+            if (withCentering) {
+                BLAS.axpy(-1, medians, outputVec);
```
yunfengzhou-hub, line 136: Would it be better if we add the method `void axpy(double a, Vector x, Vector y)` and do `Vector`-to-`DenseVector` conversions in this method if it is the best practice?

yunfengzhou-hub (commented):
Thanks for the update. Left two comments as below. The rest LGTM.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/robustscaler/RobustScalerModel.java
yunfengzhou-hub: According to our offline discussion, I think that we should avoid allocating/deallocating memories in BLAS operations. Thus the current implementation is acceptable to me.

lindong28:
Thanks @jiangxin369 for the PR! Thanks @yunfengzhou-hub for the review. LGTM.

PR #174, opened 2022-11-14
[FLINK-29604] Add Estimator and Transformer for CountVectorizer

jiangxin369:

Add Estimator and Transformer for CountVectorizer.

## Brief change log

- Adds Transformer and Estimator implementation of CountVectorizer in Java and Python.
- Adds examples and documentation of CountVectorizer.

## Documentation

### DIFF
--- a/docs/content/docs/operators/feature/countvectorizer.md
+++ b/docs/content/docs/operators/feature/countvectorizer.md
@@ -1,0 +1,182 @@
+---
+title: "Count Vectorizer"
+weight: 1
+type: docs
+aliases:
+- /operators/feature/countvectorizer.html
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions dand limitations
+under the License.
+-->
+
+## Count Vectorizer
+
+CountVectorizer is an algorithm that converts a collection of text
+documents to vectors of token counts. When an a-priori dictionary is not 
+available, CountVectorizer can be used as an estimator to extract the 
+vocabulary, and generates a CountVectorizerModel. The model produces sparse
+representations for the documents over the vocabulary, which can then be 
+passed to other algorithms like LDA.
+
+### Input Columns
+
+| Param name | Type     | Default   | Description         |
+|:-----------|:---------|:----------|:--------------------|
+| inputCol   | String[] | `"input"` | Input string array. |
+
+### Output Columns
+
+| Param name | Type         | Default    | Description             |
+|:-----------|:-------------|:-----------|:------------------------|
+| outputCol  | SparseVector | `"output"` | Vector of token counts. |
+
+### Parameters
+
+Below are the parameters required by `CountVectorizerModel`.
+
+| Key        | Default    | Type    | Required | Description                                                                                                                                                                                                                                                                                                                                     |
+|------------|------------|---------|----------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| inputCol   | `"input"`  | String  | no       | Input column name.                                                                                                                                                                                                                                                                                                                              |
+| outputCol  | `"output"` | String  | no       | Output column name.                                                                                                                                                                                                                                                                                                                             |
+| minTF      | `1.0`      | Double  | no       | Filter to ignore rare words in a document. For each document, terms with frequency/count less than the given threshold are ignored. If this is an integer >= 1, then this specifies a count (of times the term must appear in the document); if this is a double in [0,1), then this specifies a fraction (out of the document's token count).  |
+| binary     | `false`    | Boolean | no       | Binary toggle to control the output vector values. If True, all nonzero counts (after minTF filter applied) are set to 1.0.                                                                                                                                                                                                                     |
+
+`CountVectorizer` needs parameters above and also below.
+
+| Key            | Default    | Type     | Required | Description                                                                                                                                                                                                                                                                                                                                                                                  |
+|:---------------|:-----------|:---------|:---------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| vocabularySize | `2^18`     | Integer  | no       | Max size of the vocabulary. CountVectorizer will build a vocabulary that only considers the top vocabulary size terms ordered by term frequency across the corpus.                                                                                                                                                                                                                           |
+| minDF          | `1.0`      | Double   | no       | Specifies the minimum number of different documents a term must appear in to be included in the vocabulary. If this is an integer >= 1, this specifies the number of documents the term must appear in; if this is a double in [0,1), then this specifies the fraction of documents.                                                                                                         |
+| maxDF          | `2^63 - 1` | Double   | no       | Specifies the maximum number of different documents a term could appear in to be included in the vocabulary. A term that appears more than the threshold will be ignored. If this is an integer >= 1, this specifies the maximum number of documents the term could appear in; if this is a double in [0,1), then this specifies the maximum fraction of documents the term could appear in. |
+
+### Examples
+
+{{< tabs examples >}}
+
+{{< tab "Java">}}
+
+```java
+import org.apache.flink.ml.feature.countvectorizer.CountVectorizer;
+import org.apache.flink.ml.feature.countvectorizer.CountVectorizerModel;
+import org.apache.flink.ml.linalg.SparseVector;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CloseableIterator;
+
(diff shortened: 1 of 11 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizer.java
```diff
@@ -0,0 +1,219 @@
...
+        public CountVectorizerModelData getResult(
+                Tuple2<Long, Map<String, Tuple2<Long, Long>>> vocabAccumulator) {
+            Preconditions.checkState(vocabAccumulator.f0 > 0, "The training set is empty.");
+
+            boolean filteringRequired =
+                    !MIN_DF.defaultValue.equals(minDF) || !MAX_DF.defaultValue.equals(maxDF);
+            if (filteringRequired) {
+                long rowNum = vocabAccumulator.f0;
+                double actualMinDF = minDF >= 1.0 ? minDF : minDF * rowNum;
+                double actualMaxDF = maxDF >= 1.0 ? maxDF : maxDF * rowNum;
```
yunfengzhou-hub, line 152: It might be better to declare these two variables as integers, as they represent the number of documents.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/CountVectorizerTest.java
```diff
@@ -0,0 +1,391 @@
...
+        assertEquals("test_output", countVectorizer.getOutputCol());
+        assertEquals(0.9, countVectorizer.getMaxDF(), EPS);
+        assertEquals(0.1, countVectorizer.getMinDF(), EPS);
+        assertEquals(10, countVectorizer.getMinTF(), EPS);
+        assertEquals(1000, countVectorizer.getVocabularySize());
+        assertTrue(countVectorizer.getBinary());
+    }
+
+    @Test
+    public void testInvalidParam() {
```
yunfengzhou-hub, line 153: This test case only tests `minDF` and `maxDF`, not generally all parameters. We can modify the name of this test case to reflect this, or add tests for other possibly invalid parameters.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/CountVectorizerTest.java
```diff
@@ -0,0 +1,391 @@
...
+            assertEquals(errMessage, ExceptionUtils.getRootCause(e).getMessage());
+        }
+    }
+
+    @Test
+    public void testOutputSchema() {
+        CountVectorizer countVectorizer = new CountVectorizer();
+        CountVectorizerModel model = countVectorizer.fit(trainDataTable);
+        Table output = model.transform(trainDataTable)[0];
+        assertEquals(Arrays.asList("input", "output"), output.getResolvedSchema().getColumnNames());
```
yunfengzhou-hub, line 199: Let's make the column names different from default values.

docs/content/docs/operators/feature/countvectorizer.md
```diff
@@ -0,0 +1,182 @@
...
+| minDF          | `1.0`      | Double   | no       | Specifies the minimum number of different documents a term must appear in to be included in the vocabulary. If this is an integer >= 1, this specifies the number of documents the term must appear in; if this is a double in [0,1), then this specifies the fraction of documents.                                                                                                         |
+| maxDF          | `2^65 - 1` | Double   | no       | Specifies the maximum number of different documents a term could appear in to be included in the vocabulary. A term that appears more than the threshold will be ignored. If this is an integer >= 1, this specifies the maximum number of documents the term could appear in; if this is a double in [0,1), then this specifies the maximum fraction of documents the term could appear in. |
```
yunfengzhou-hub, line 64: 2^63 - 1.

docs/content/docs/operators/feature/countvectorizer.md
```diff
@@ -0,0 +1,182 @@
...
+|:---------------|:-----------|:---------|:---------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| vocabularySize | `262144`   | Integer  | no       | Max size of the vocabulary. CountVectorizer will build a vocabulary that only considers the top vocabulary size terms ordered by term frequency across the corpus.                                                                                                                                                                                                                           |
```
yunfengzhou-hub, line 62: Would it be better to keep consistent with `maxDF`, and write `2^18` here?

yunfengzhou-hub (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizer.java
jiangxin369: The benefit of using a double is that we don't need to care about whether to use `floor` or `ceil`, `>` or `>=` which makes the judgment more complex.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizerModel.java
```diff
@@ -0,0 +1,185 @@
...
+
+            String[] document = (String[]) row.getField(inputCol);
+            double[] termCounts = new double[vocabulary.size()];
+            for (String word : document) {
+                if (vocabulary.containsKey(word)) {
+                    termCounts[vocabulary.get(word)] += 1;
+                }
+            }
+
+            double actualMinTF = minTf >= 1.0 ? minTf : document.length * minTf;
```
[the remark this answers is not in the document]
jiangxin369: Same to the below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizerParams.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Params of {@link CountVectorizer}.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface CountVectorizerParams<T> extends CountVectorizerModelParams<T> {
+    Param<Integer> VOCABULARY_SIZE =
+            new IntParam(
+                    "vocabularySize",
+                    "Max size of the vocabulary. CountVectorizer will build a vocabulary"
```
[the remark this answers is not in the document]
jiangxin369: IMO, these explanations are all important to users and need to be shown.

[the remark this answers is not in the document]
yunfengzhou-hub: I agree with it that the descriptions here are different from that of the binary parameter. We can keep this description.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizer.java
yunfengzhou-hub: Got it. I think the current implementation is OK then.

docs/content/docs/operators/feature/countvectorizer.md
```diff
@@ -0,0 +1,182 @@
...
+
+        // Generates input training and prediction data.
+        DataStream<Row> trainStream =
+                env.fromElements(
+                        Row.of((Object) new String[] {"a", "c", "b", "c"}),
+                        Row.of((Object) new String[] {"c", "d", "e"}),
+                        Row.of((Object) new String[] {"a", "b", "c"}),
+                        Row.of((Object) new String[] {"e", "f"}),
+                        Row.of((Object) new String[] {"a", "c", "a"}));
+        Table trainTable = tEnv.fromDataStream(trainStream).as("input");
```
yunfengzhou-hub, line 102: This variable contains both training and prediction data. It might be better to rename this variable as `inputTable`. Same for that in the python example.

docs/content/docs/operators/feature/countvectorizer.md
```diff
@@ -0,0 +1,182 @@
...
+| inputCol   | `"input"`  | String  | no       | Input column name.                                                                                                                                                                                                                                                                                                                              |
+| outputCol  | `"output"` | String  | no       | Output column name.                                                                                                                                                                                                                                                                                                                             |
+| minTF      | `1.0`      | Double  | no       | Filter to ignore rare words in a document. For each document, terms with frequency/count less than the given threshold are ignored. If this is an integer >= 1, then this specifies a count (of times the term must appear in the document); if this is a double in [0,1), then this specifies a fraction (out of the document's token count).  |
```
yunfengzhou-hub, line 55: In fact, users can set the value of `minTF` to any double >= 1, in which case the specified count is `Math.floor(minTF)`. Do you think it would be better if we clarify that if `minTF >= 1`, it does not have to be an integer? Save for `minDF` and `maxDF`.

yunfengzhou-hub (commented):

docs/content/docs/operators/feature/countvectorizer.md
jiangxin369: With the current expression, users are recommended to set this param to an integer if specifies the count, which makes the meaning of this parameter clear.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizer.java
```diff
@@ -0,0 +1,218 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * {@link CountVectorizer} aims to help convert a collection of text documents to vectors of token
+ * counts. When an a-priori dictionary is not available, {@link CountVectorizer} can be used as an
+ * estimator to extract the vocabulary, and generates a {@link CountVectorizerModel}. The model
+ * produces sparse representations for the documents over the vocabulary, which can then be passed
+ * to other algorithms like LDA.
```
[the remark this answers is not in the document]
jiangxin369: * The JavaDoc often provides a description of the Estimator and starts with "An Estimator which", while the markdown starts with "CountVectorizer is an algorithm that ".
* I'll remove the `@link`.

docs/content/docs/operators/feature/countvectorizer.md
yunfengzhou-hub: According to offline discussion, we'll allow users to set the parameter to any double value larger than 1, so as to keep aligned with the function of the count vectorizer in Spark. However, it is still not recommended for users to that this parameter to such values.

lindong28:
Thanks for the PR! Thanks @yunfengzhou-hub for the review. LGTM.

PR #175, opened 2022-11-15
[FLINK-29602] Add Transformer for SQLTransformer

yunfengzhou-hub:

This PR adds the Transformer for the SQL transformer algorithm.

## Brief change log

  - Adds Transformer implementation of SQL transformer in Java and Python
  - Adds examples and documentation of SQL transformer

## Documentation

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/sqltransformer/SQLTransformer.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/sqltransformer/SQLTransformer.java
@@ -1,0 +1,193 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.sqltransformer;
+
+import org.apache.flink.api.common.functions.AggregateFunction;
+import org.apache.flink.api.common.functions.FlatMapFunction;
+import org.apache.flink.api.common.typeinfo.Types;
+import org.apache.flink.ml.api.Transformer;
+import org.apache.flink.ml.common.datastream.EndOfStreamWindows;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableException;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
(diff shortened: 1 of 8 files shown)


jiangxin369 (commented):

jiangxin369:
Thanks for the update. LGTM overall.

@lindong28  Could you please take a look at this PR?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/SQLTransformerTest.java
```diff
@@ -0,0 +1,241 @@
...
+                new SQLTransformer().setStatement("SELECT SUM(v1) AS v3 FROM __THIS__ GROUP BY v2");
+
+        Table outputTable = sqlTransformer.transform(inputTable)[0];
+
+        verifyOutputResult(outputTable, EXPECTED_BATCH_AGGREGATION_OUTPUT, false);
+    }
+
+    @Test
+    public void testStreamingAggregation() {
+        env.setRuntimeMode(RuntimeExecutionMode.STREAMING);
```
lindong28, line 165: Runtime mode is streaming by default. It seems better to remove this line for consistency with existing tests.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/sqltransformer/SQLTransformer.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * SQLTransformer implements the transformations that are defined by SQL statement.
+ *
+ * <p>Currently we only support SQL syntax like 'SELECT ... FROM __THIS__ ...' where "__THIS__"
```
lindong28, line 37: Can we also support `GROUP BY`? 

When the statement includes `group by`, this operator can emit results at the end of stream, similar to the behavior of many existing operators in Flink ML.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/sqltransformer/SQLTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+        @Override
+        public List<Row> createAccumulator() {
+            return new ArrayList<>();
+        }
+
+        @Override
+        public List<Row> add(Row value, List<Row> accumulator) {
+            switch (value.getKind()) {
+                case INSERT:
+                    accumulator.add(value);
```
lindong28, line 157: I am not sure if buffering the row will prevent this operator from being used with object-reuse enabled.

Could you update `SQLTransformerTest` to enable object re-use?

We also need to update tests of all existing Flink ML operator to enable object-reuse. Can you open a separate PR to do this?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/SQLTransformerTest.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.Comparator;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/** Tests {@link SQLTransformer}. */
+public class SQLTransformerTest extends AbstractTestBase {
+    private static final List<Row> INPUT_DATA =
+            Arrays.asList(Row.of(0, 1.0, 3.0), Row.of(2, 2.0, 3.0));
```
lindong28, line 53: Would you update this input data so that when it is used by the query `SELECT SUM(v1) AS v3 FROM __THIS__ GROUP BY v2`, we can have multiple distinct group-by keys in the output, and each key has multiple matched rows?

lindong28 (commented):
Thanks for the update. Left two minor comments below.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/sqltransformer/SQLTransformer.java
yunfengzhou-hub: I agree. I'll enable object re-use for `SQLTransformerTest` now and open a separate PR for all other algorithms.

lindong28:
Thanks for the update! LGTM.

PR #177, opened 2022-11-17
[FLINK-30014][hotfix] Ensure saving non-null value in aggregate util function state

yunfengzhou-hub:

This PR fixes the NPE that would possibly be thrown when using `DataStreamUtils.aggregate`.

## Brief change log

  - Save only non-null values in the aggregate utility function's state.
  - Clarify the behavior of the aggregate utility function in respect of input parallelism and add a test for the corner case.

## Documentation

### DIFF
--- a/flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
+++ b/flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
@@ -87,6 +87,27 @@ public class DataStreamUtilsTest {
     }
 
     @Test
+    public void testAggregateWithNonNeutralInitialAccumulator() throws Exception {
+        DataStream<Long> dataStream =
+                env.fromParallelCollection(new NumberSequenceIterator(0L, 19L), Types.LONG);
+        DataStream<String> result =
+                DataStreamUtils.aggregate(
+                        dataStream, new TestAggregateFuncWithNonNeutralInitialAccumulator());
+        List<String> stringSum = IteratorUtils.toList(result.executeAndCollect());
+        assertEquals(1, stringSum.size());
+        assertEquals(Integer.toString(190 + env.getParallelism()), stringSum.get(0));
+
+        env.setParallelism(env.getParallelism() + 1);
+        dataStream = env.fromParallelCollection(new NumberSequenceIterator(0L, 19L), Types.LONG);
+        result =
+                DataStreamUtils.aggregate(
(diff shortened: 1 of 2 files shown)


flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
```diff
@@ -86,6 +87,29 @@ public void testAggregate() throws Exception {
...
+        DataStream<Long> dataStream =
+                env.fromParallelCollection(new NumberSequenceIterator(0L, 19L), Types.LONG);
+        DataStream<String> result =
+                DataStreamUtils.aggregate(
+                        dataStream, new TestAggregateFuncWithNonNeutralInitialAccumulator());
+        List<String> stringSumList = IteratorUtils.toList(result.executeAndCollect());
+        assertEquals(1, stringSumList.size());
+        String stringSum1 = stringSumList.get(0);
+
+        env.setParallelism(env.getParallelism() + 1);
```
zhipeng93, line 101: Let's reset the parallelism as the default value at the end of this test.

flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
```diff
@@ -86,6 +87,29 @@ public void testAggregate() throws Exception {
...
+        dataStream = env.fromParallelCollection(new NumberSequenceIterator(0L, 19L), Types.LONG);
+        result =
+                DataStreamUtils.aggregate(
+                        dataStream, new TestAggregateFuncWithNonNeutralInitialAccumulator());
+        stringSumList = IteratorUtils.toList(result.executeAndCollect());
+        assertEquals(1, stringSumList.size());
+        String stringSum2 = stringSumList.get(0);
+
+        assertNotEquals(stringSum1, stringSum2);
```
zhipeng93, line 110: Let's directly check the value of the output.

zhipeng93 (commented):
Thanks for the fix. Left two minor comments below.

flink-ml-core/src/test/java/org/apache/flink/ml/common/datastream/DataStreamUtilsTest.java
yunfengzhou-hub: The parallelism of the environment would be reset in the `before()` method, which happens before each test case, so there is no need to reset the parallelism to the default value here.

zhipeng93:
Thanks for the update. LGTM.

PR #193, opened 2022-12-28
[FLINK-30515] Add benchmark for CountVectorizer, Imputer, RobustScaler, UnivariateFeatureSelector and VarianceThresholdSelector

jiangxin369:

Add benchmark for CountVectorizer, Imputer, RobustScaler, UnivariateFeatureSelector, and VarianceThresholdSelector.

## Brief change log

  - Add benchmark for CountVectorizer, Imputer, RobustScaler, UnivariateFeatureSelector, and VarianceThresholdSelector
  - Add `RandomStringArrayGenerator` to help generate a table of string arrays.
  - Make `HasNumDistinctValues` a public param.

## Documentation


jiangxin369:
@lindong28 Could you help review this PR?

flink-ml-benchmark/src/main/resources/stringindexer-benchmark.json
```diff
@@ -17,7 +17,7 @@
   "version": 1,
   "stringindexer100000000": {
     "inputData": {
-      "className": "org.apache.flink.ml.benchmark.datagenerator.common.RandStringGenerator",
+      "className": "org.apache.flink.ml.benchmark.datagenerator.common.RandomStringGenerator",
```
yunfengzhou-hub, line 20: Let's add a test case to verify that all existing benchmark json files are valid. This means the class names correspond to existing classes, and that the param maps are reasonable.

yunfengzhou-hub (commented):
Thanks for the PR! This PR generally LGTM. Left one small comment as below.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/param/HasNumDistinctValues.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.flink.ml.benchmark.datagenerator.param;
+
+import org.apache.flink.ml.param.IntParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the benchmark num distinct values param. */
+public interface HasNumDistinctValues<T> extends WithParams<T> {
```
yunfengzhou-hub, line 27: Would it be better if we merge this parameter with the `ARITY` parameter in `DoubleGenerator`? They provide similar functions.

yunfengzhou-hub, line 27: Would it be better if we merge this parameter with the `ARITY` parameter in DoubleGenerator? They provide similar functions.

jiangxin369: The `ARITY` param has different behavior when it is set to 0, so I think we don't have to merge them.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
```diff
@@ -54,6 +56,33 @@ public void testParseJsonFile() throws Exception {
...
+        File[] jsonFiles =
+                resourcesDir.listFiles(
+                        (dir, name) ->
+                                name.endsWith(".json") && !name.equals("benchmark-demo.json"));
+
+        for (File file : jsonFiles) {
+            Map<String, Map<String, Map<String, ?>>> benchmarks =
+                    BenchmarkUtils.parseJsonFile(file.getAbsolutePath());
+            for (Map<String, Map<String, ?>> params : benchmarks.values()) {
+                ReadWriteUtils.instantiateWithParams(params.get("stage"));
```
yunfengzhou-hub, line 77: nit: we can also add verifications to these aspects:

- `jsonFiles` is not empty.
- `params` contains no other keys except `stage`, `inputData` and `modelData`.
- The `stage` object is a subclass of `Stage`.
- The `inputData` or `modelData` object is a subclass of `DataGenerator`.

yunfengzhou-hub (commented):
Thanks for the update! Left one small comment as below.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
jiangxin369: For the first aspect,`BenchmarkUtils.parseJsonFile` would check if `jsonfile` is empty. Others are accepted.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
```diff
@@ -54,6 +60,44 @@ public void testParseJsonFile() throws Exception {
...
+        for (File file : jsonFiles) {
+            Map<String, Map<String, Map<String, ?>>> benchmarks =
+                    BenchmarkUtils.parseJsonFile(file.getAbsolutePath());
+            for (Map<String, Map<String, ?>> params : benchmarks.values()) {
+                assertTrue(
+                        Arrays.asList("stage", "inputData", "modelData")
+                                .containsAll(params.keySet()));
+
+                WithParams stage = ReadWriteUtils.instantiateWithParams(params.get("stage"));
+                assertTrue(stage instanceof Stage);
```
lindong28, line 86: It is nice to add this test 👍

Would it be simpler to just copy/paste the following lines from `BenchmarkUtils#runBenchmark` instead using `instanceof`? This approach also reduces the `Raw use of parameterized class ...` warnings. 

```
Stage stage = ReadWriteUtils.instantiateWithParams(params.get("stage"));
InputDataGenerator inputDataGenerator =
        ReadWriteUtils.instantiateWithParams(params.get("inputData"));
DataGenerator modelDataGenerator = null;
if (params.containsKey("modelData")) {
    modelDataGenerator = ReadWriteUtils.instantiateWithParams(params.get("modelData"));
}
```

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
```diff
@@ -54,6 +60,44 @@ public void testParseJsonFile() throws Exception {
         assertTrue(benchmarks.containsKey("KMeansModel-1"));
     }
 
+    @Test
+    public void testJsonFileLegality() throws IOException, ClassNotFoundException {
+        File resourcesDir =
```
lindong28, line 65: Can you help update `testParseJsonFile()` to also use this approach instead of having to copy the file to `test-conf.json`?

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
```diff
@@ -54,6 +60,44 @@ public void testParseJsonFile() throws Exception {
...
+                new File(
+                                this.getClass()
+                                        .getClassLoader()
+                                        .getResource("benchmark-demo.json")
+                                        .getPath())
+                        .getParentFile();
+        File[] jsonFiles =
+                resourcesDir.listFiles(
+                        (dir, name) ->
+                                name.endsWith(".json") && !name.equals("benchmark-demo.json"));
```
lindong28, line 75: It might be a bit simpler to remove `name.equals("benchmark-demo.json")` since the overhead of parsing this file is negligible compared to all the other work we need to do in the CI.

flink-ml-benchmark/src/main/resources/robustscaler-benchmark.json
```diff
@@ -0,0 +1,42 @@
...
+          ]
+        ],
+        "seed": 2,
+        "numValues": 10000000
+      }
+    },
+    "stage": {
+      "className": "org.apache.flink.ml.feature.robustscaler.RobustScaler",
+      "paramMap": {
+        "inputCol": "features",
```
lindong28, line 35: Would it be simpler to remove this line and use the default value? Same for `outputCol`.

Same for similar parameters used in other benchmark json files added in this PR.

lindong28 (commented):

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
jiangxin369: Sure.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
jiangxin369: Using `instanceOf` gives a clearer error message if someone writes an incorrect class in json file, instead of `ClassCastException`. The approach in `BenchmarkUtils#runBenchmark` has no warning because of the `@SuppressWarnings` annotation, I can add this annotation.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
jiangxin369: The `benchmark-demo.json` is not included because some cases in the file like `Unmatch-Input` has invalid paramMap.  It is by design which makes it cannot be parsed correctly.

lindong28: I see. Sounds good.

flink-ml-benchmark/src/main/resources/robustscaler-benchmark.json
jiangxin369: I prefer to keep the `inputCol` because it shows the relationship between `DataGenerator` and `Stage`. As for `outputCol` we can remove it.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
lindong28: Error message from unit test is supposed be read by developers and it is OK for ask developers to understand and debug ClassCastException in this case, right?

If we want error message to be more meaningful to users who run benchmarks, should we update `ReadWriteUtils#instantiateWithParams()` to catch `ClassCastException` and replace it properly?

Can you provide an example stack trace in the case described above so that we can check whether we need to improve it?

flink-ml-benchmark/src/main/resources/robustscaler-benchmark.json
lindong28: How about we change the column name from `features` to `inputCol` for simplicity and consistency with other tests?

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
jiangxin369: An example stack trace is like this:
```
java.lang.ClassCastException: org.apache.flink.ml.feature.imputer.ImputerTest cannot be cast to org.apache.flink.ml.api.Stage
	at org.apache.flink.ml.benchmark.BenchmarkTest.testJsonFileLegality(BenchmarkTest.java:137)
	at org.apache.flink.util.TestNameProvider$1.evaluate(TestNameProvider.java:45)
```

flink-ml-benchmark/src/main/resources/robustscaler-benchmark.json
jiangxin369: Actually, there's no consistency within existing tests. Currently, we have both `input` and `features` as the column name. To keep consistency, I'll change all of them to `input`, is that acceptable to you?

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
lindong28: Sounds good. Thanks for the example.

lindong28: Sounds good. Thanks for the example.

flink-ml-benchmark/src/test/java/org/apache/flink/ml/benchmark/BenchmarkTest.java
```diff
@@ -103,4 +166,32 @@ public void testRunBenchmark() throws Exception {
                 result.outputThroughput,
                 1e-5);
     }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private void dryRunBenchmark(Map<String, Map<String, ?>> params)
```
lindong28, line 171: It seems that this method duplicates a lot of existing logic implemented by `BenchmarkUtils#runBenchmark`. Would it be simpler to just add a `boolean isDryRun` parameter to that method?

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/clustering/KMeansModelDataGenerator.java
```diff
@@ -55,20 +61,36 @@ public Table[] getData(StreamTableEnvironment tEnv) {
         InputDataGenerator<?> vectorArrayGenerator = new DenseVectorArrayGenerator();
         ReadWriteUtils.updateExistingParams(vectorArrayGenerator, paramMap);
         vectorArrayGenerator.setNumValues(1);
+        vectorArrayGenerator.setColNames(new String[] {"centroids"});
```
lindong28, line 64: Why do we need to make this change? This is change necessary to add benchmarks for algorithms targeted by this PR?

flink-ml-benchmark/src/main/resources/minmaxscaler-benchmark.json
```diff
@@ -22,19 +22,15 @@
         "vectorDim": 100,
         "colNames": [
           [
-            "features"
+            "inputCol"
```
lindong28, line 25: Can you verify that this configuration still works after the change?

Same for other configuration files updated by this PR.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/clustering/KMeansModelDataGenerator.java
jiangxin369: Yes, it is necessary. Because there's a mistake in `KmeansModelDataGenerator` which would break the test for `benchmark-demo.json`.

flink-ml-benchmark/src/main/resources/minmaxscaler-benchmark.json
jiangxin369: Thanks for reminding, I verified all updated files and they work well.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/clustering/KMeansModelDataGenerator.java
jiangxin369: Adding the first line `vectorArrayGenerator.setColNames(new String[] {"centroids"});` is because `DenseVectorArrayGenerator` requires `columnNames` is set, otherwise NPE would be thrown. 
The other changes are to reduce the operation of `toDataStream` and `fromDataStream`, just do the conversion base on `Table`.

lindong28: Sounds good. Thanks for the explanation.

lindong28:
Thanks for the update. LGTM.

PR #196, opened 2023-01-02
[FLINK-30541] Add Transformer and Estimator for OnlineStandardScaler

zhipeng93:
- Add Transformer and Estimator for OnlineStandardScaler

## Brief change log
- Added Transformer and Estimator for OnlineStandardScaler (java/python source/test/example and docs).
- Added two metrics for exposing ML model attributes (i.e., `timestamp` and `version` of model data).

## Documentation

  - If yes, how is the feature documented? (docs)


flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerParams.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.feature.standardscaler;
+
+import org.apache.flink.ml.common.param.HasWindows;
+
+/** Params for {@link OnlineStandardScaler}. */
+public interface OnlineStandardScalerParams<T>
+        extends HasWindows<T>, OnlineStandardScalerModelParams<T> {}
```
jiangxin369, line 25: From my testing, not all `Windows` are supported, e.g., when using `ProcessingTimeTumblingWindows`, the `OnlineStandardScaler` cannot train any models. Is that expected behavior? If so, maybe we can add a validator and document the windows we supported.

jiangxin369 (commented):
@zhipeng93 Thanks for the PR, left some comments.

flink-ml-python/pyflink/ml/lib/param.py
```diff
@@ -581,3 +581,48 @@ def get_flatten(self) -> bool:
     @property
     def flatten(self):
         return self.get(self.FLATTEN)
+
+
+class HasModelVersionCol(WithParams, ABC):
```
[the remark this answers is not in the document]
zhipeng93: Good catch. I will add the test there.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScaler.java
```diff
@@ -0,0 +1,215 @@
...
+                            .orElse(0L);
+            long modelVersion =
+                    OperatorStateUtils.getUniqueElement(modelVersionState, "modelVersionState")
+                            .orElse(0L);
+
+            long numElementsBefore = numElements;
+            for (Row element : iterable) {
+                Vector inputVec =
+                        ((Vector) Objects.requireNonNull(element.getField(inputCol))).clone();
+                if (numElements == 0) {
```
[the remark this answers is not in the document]
zhipeng93: Puting it inside the loop is to get the size of the vector for initializing `sum` and `squareSum`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScaler.java
```diff
@@ -0,0 +1,215 @@
...
+                    OperatorStateUtils.getUniqueElement(numElementsState, "numElementsState")
+                            .orElse(0L);
+            long modelVersion =
+                    OperatorStateUtils.getUniqueElement(modelVersionState, "modelVersionState")
+                            .orElse(0L);
+
+            long numElementsBefore = numElements;
+            for (Row element : iterable) {
+                Vector inputVec =
+                        ((Vector) Objects.requireNonNull(element.getField(inputCol))).clone();
```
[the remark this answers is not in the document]
zhipeng93: Because we need to do some in-place operations in Line#154, i.e., `BLAS.hDot(inputVec, inputVec);`.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/OnlineStandardScalerTest.java
```diff
@@ -0,0 +1,396 @@
...
+        OnlineStandardScaler standardScaler = new OnlineStandardScaler();
+        Table output;
+        int windowSizeMs = 3000;
+
+        // Tests event time window with maxAllowedModelDelayMs as 0.
+        standardScaler
+                .setWindows(EventTimeTumblingWindows.of(Time.milliseconds(windowSizeMs)))
+                .setModelVersionCol("modelVersionCol");
+        output = standardScaler.fit(inputTableWithEventTime).transform(inputTableWithEventTime)[0];
+        verifyUsedModelVersion(
```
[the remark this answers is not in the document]
zhipeng93: The prediction result is undetermined because we may use new model to predict old data.
So we can only check the model version here.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerParams.java
zhipeng93: It supports `ProcessintTimeTumbingWindows` and I have added a corresponding unit test. Please take a look.

zhipeng93:
> @zhipeng93 Thanks for the PR, left some comments.

Thanks @jiangxin369 for the review. I have addressed your comments in the latest commit. Please take a look.

jiangxin369 (commented):
@zhipeng93 Thanks for the updates, left some minor comments.

docs/content/docs/operators/feature/onlinestandardscaler.md
```diff
@@ -0,0 +1,261 @@
...
+For each window, it computes the mean and standard deviation using the data seen
+so far (i.e., not only the data in the current window, but also the history data).
+The model data generated by OnlineStandardScaler is a model stream. 
+There is one model data for each window.
+
+During the inference phase (i.e., using OnlineStandardScalerModel for prediction),
+users could output the model version that is used for predicting each data point. 
+Moreover,
+- When the train data and test data both contain event time, users could 
+specify the maximum difference between timestamp of the input and model data,
```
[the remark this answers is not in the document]
zhipeng93: ah, thanks for pointing this out, it is fixed now.

jiangxin369:
@zhipeng93 Thanks for the updates, LGTM. @lindong28 Could you take a look at the PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
```diff
@@ -0,0 +1,298 @@
...
+            for (StreamRecord dataPoint : bufferedPointsState.get()) {
+                if (dataPoint.getTimestamp() <= modelTimeStamp + maxAllowedModelDelayMs) {
+                    doPrediction(dataPoint);
+                } else {
+                    unprocessedElements.add(dataPoint);
+                }
+            }
+            bufferedPointsState.clear();
+            if (unprocessedElements.size() > 0) {
+                bufferedPointsState.update(unprocessedElements);
```
lindong28, line 241: As a minor optimization, can we only clear/update the state if at least one element is predicted in the above for loop?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
```diff
@@ -0,0 +1,298 @@
...
+                            .getMetricGroup()
+                            .addGroup(MLMetrics.ML_GROUP)
+                            .addGroup(MLMetrics.ML_MODEL_GROUP);
+            mlModelMetricGroup.gauge(MLMetrics.TIMESTAMP, (Gauge<Long>) () -> modelTimeStamp);
+            mlModelMetricGroup.gauge(MLMetrics.VERSION, (Gauge<Long>) () -> modelVersion);
+        }
+
+        @Override
+        public void processElement1(StreamRecord<Row> dataPoint) throws Exception {
+            if (dataPoint.getTimestamp() <= modelTimeStamp + maxAllowedModelDelayMs
```
lindong28, line 204: According to the maxAllowedModelDelayMs doc, this parameter should only be used if the input contains event time. I suppose there should be logic that differentiates between event time and system time. Can you explain where is this logic?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        @Override
+        public void open() throws Exception {
+            super.open();
+            MetricGroup mlModelMetricGroup =
+                    getRuntimeContext()
+                            .getMetricGroup()
+                            .addGroup(MLMetrics.ML_GROUP)
+                            .addGroup(MLMetrics.ML_MODEL_GROUP);
```
lindong28, line 197: Since we might have multiple models running in the same Flink Job, it might be better to specify the model name to reduce chance of conflict.

How about using `addGroup(MLMetrics.ML_MODEL_GROUP, "OnlineStandardScalerModel")` here?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
```diff
@@ -0,0 +1,298 @@
...
+            mlModelMetricGroup.gauge(MLMetrics.VERSION, (Gauge<Long>) () -> modelVersion);
+        }
+
+        @Override
+        public void processElement1(StreamRecord<Row> dataPoint) throws Exception {
+            if (dataPoint.getTimestamp() <= modelTimeStamp + maxAllowedModelDelayMs
+                    && mean != null) {
+                doPrediction(dataPoint);
+            } else {
+                bufferedPointsState.add(dataPoint);
```
lindong28, line 208: Given that we only update `modelDataState` in `snapshotState()`, I suppose it is because updating `modelDataState()` every time we receive a model data is more expensive than updating the in-memory `modelData`.

Should we follow the same approach for `dataPoint` consistency and performance?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasModelVersionCol.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.flink.ml.param.StringParam;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared model version column param. */
+public interface HasModelVersionCol<T> extends WithParams<T> {
+    Param<String> MODEL_VERSION_COL =
+            new StringParam(
+                    "modelVersionCol",
+                    "The version of the model data that the input data is predicted with.",
+                    null);
```
lindong28, line 31: Would it be more usable to using "version" as the default value and always output this information? The reason is that I suppose adding an extra long column should not have an observable performance impact for nearline (not batch) prediction.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
zhipeng93: Good catch, it is indeed a bug. I have fixed the training logic as follows:
- When the window strategy is based on event-time, the timestamp of model data is the maximum timestamp of the window that generates this model.
- Otherwise, the timestamp of the model data is `LONG.MAX_VALUE`.

Moreover, to avoid the possible overflow, the `if condition` here is updated as `if (dataPoint.getTimestamp() - maxAllowedModelDelayMs <= modelTimeStamp`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
zhipeng93: There is a bit difference between `modelDataState` and `bufferedPointsState` --- There is one single element in model data in `modelDataState` and more than one elements in `bufferedPointsState`.

Appending elements into liststate is an asyncronous operation and thus can help to improve the efficiency when caching data points. On the other side, there is only one model data and we updating it may introduce unnessary state  operations (e.g., add it to state and remove it).

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/StandardScalerTest.java
```diff
@@ -206,7 +206,7 @@ public void testSaveLoadAndPredict() throws Exception {
         model = TestUtils.saveAndReload(tEnv, model, tempFolder.newFolder().getAbsolutePath());
 
         assertEquals(
-                Arrays.asList("mean", "std"),
+                Arrays.asList("mean", "std", "version", "timestamp"),
```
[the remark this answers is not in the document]
zhipeng93: Well, I think letting `StandardScalerModel` have model version and timestamp is reasonable, since offline learning could be viewed as a special case of online learning. What we need to do is to make sure that version and timestamp has a clear meaning. For example, the version of offline model is always zero and time stamp is `LONG.MAX_VALUE`.

What do you mean by verify `model data contains those columns`? Do you mean we only verify `model data` and leave `model version` and `timestamp` unchecked?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasModelVersionCol.java
zhipeng93: I agree that it should not raise an observable performance impact here. I set the default value as `null` since it is consistent with the exiting prediction case. Is there any other benefits (except debugging) to output the model version?

zhipeng93:
@lindong28 Thanks for the review. I have updated the PR according to your comments and closed some of the issues we both aggree on. Please  take a look.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/StandardScalerTest.java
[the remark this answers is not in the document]
lindong28: > What do you mean by verify model data contains those columns? Do you mean we only verify model data and leave model version and timestamp unchecked?

Yes that is what I means. IMO StandardScalerTest should verify the expected behavior of StandardScalerModel from the perspective of users who want to use StandardScalerModel. IMO users of StandardScalerModel should not need to know concepts such as model version or timestamp. Users should `OnlineStandardScaler` needs to know these information, and this behavior is covered by `OnlineStandardScalerTest`.

> I think letting StandardScalerModel have model version and timestamp is reasonable, since offline learning could be viewed as a special case of online learning

Yes we can tell users this. But for users of StandardScaler, there is no benefit to knowing this. So it seems simpler not to tell users of `StandardScalerModel` about version etc.?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
lindong28: Thanks for the explanation. Sounds good.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasModelVersionCol.java
lindong28: By `the exiting prediction case`, do you mean `HasPredictionCol`? Currently `HasPredictionCol` defaults to prediction. Can you clarify it here?

The benefit is to minimize the config that users need to tune for the common scenario targeted by algorithms that depends on `HasModelVersionCol`.

I suppose that for algorithms such as `OnlineStandardScaler`, their users will run the algorithm continuously and observe multiple models emitted by it. And it seems useful for users to know "which version of the model data that the input data is predicted with". Since there is observable benefit and negligible cost, it seems better to enable this column by default?

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/StandardScalerTest.java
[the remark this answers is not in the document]
zhipeng93: Thanks for the explanation. I have updated the PR as you suggested.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasModelVersionCol.java
zhipeng93: Thanks for the explaination. I have changed the default value of `HasModelVersionCol` as `version`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
```diff
@@ -0,0 +1,304 @@
...
+                            .addGroup(
+                                    MLMetrics.ML_MODEL_GROUP,
+                                    OnlineStandardScalerModel.class.getSimpleName());
+            mlModelMetricGroup.gauge(MLMetrics.TIMESTAMP, (Gauge<Long>) () -> modelTimeStamp);
+            mlModelMetricGroup.gauge(MLMetrics.VERSION, (Gauge<Long>) () -> modelVersion);
+        }
+
+        @Override
+        public void processElement1(StreamRecord<Row> dataPoint) throws Exception {
+            if (dataPoint.getTimestamp() - maxAllowedModelDelayMs <= modelTimeStamp
```
lindong28, line 206: What is the value of `modelTimeStamp` if `processElement2()` has not been invoked? Should we explicitly give it a default value?

In general it is a good practice to make sure all variables are initialized in the constructor (or `open()` in this case). Maybe do it for other variables as well.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
```diff
@@ -0,0 +1,304 @@
...
+                    && mean != null) {
```
lindong28, line 207: Is this line needed to make sure that we have received model data before processing data?

Do we still need it if set the default value of modelTimeStamp to `Long.MIN_VALUE`?

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
zhipeng93: Yes.

We probably still need the check since the timestamp of the data point could also be `LONG.MIN_VALUE`.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/standardscaler/OnlineStandardScalerModel.java
lindong28: Thanks for the fix. LGTM.

lindong28:
Thanks for the update! LGTM.

PR #197, opened 2023-01-05
    [FLINK-30566] Add benchmark for AgglomerativeClustering, HashingTF, IDF, KBinsDiscretizer, LinearRegression, LinearSVC, LogisticRegression, NGram, RegexTokenizer, Tokenizer and Vectorindexer

zhipeng93:
- Add benchmark configurations for some of the existing algorithms.

## Brief change log
- Add benchmark configuration files for AgglomerativeClustering, HashingTF, IDF, KBinsDiscretizer, LinearRegression, LinearSVC, LogisticRegression, NGram, RegexTokenizer, Tokenizer and Vectorindexer

## Documentation

### DIFF
--- a/flink-ml-benchmark/src/main/resources/linearregression-benchmark.json
+++ b/flink-ml-benchmark/src/main/resources/linearregression-benchmark.json
@@ -1,0 +1,41 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+{
+  "version": 1,
+  "linearregression": {
+    "stage": {
+      "className": "org.apache.flink.ml.regression.linearregression.LinearRegression",
+      "paramMap": {
+        "maxIter": 20,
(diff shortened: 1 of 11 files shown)


flink-ml-benchmark/src/main/resources/linearregression-benchmark.json
```diff
@@ -0,0 +1,45 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+{
+  "version": 1,
+  "linearregression": {
+    "stage": {
+      "className": "org.apache.flink.ml.regression.linearregression.LinearRegression",
+      "paramMap": {
+        "featuresCol": "features",
```
lindong28, line 22: Would it be simpler (and more consistent with existing benchmark configs) to skip specifying the default values for `featuresCol`?

Same for `labelCol`, `weightCol`, `inputCol` and `outputCol`.

lindong28 (commented):
Thanks for the PR. LGTM overall. Left one minor comment.

flink-ml-benchmark/src/main/resources/linearregression-benchmark.json
zhipeng93: I hold a bit different opinion here.

Given that we have to specify column names for input data `features, labels, weight`, adding this configurations in the algorithms (i.e., `linear regression` here) is supposed to make it easier to follow. What do you think?

lindong28: Strictly speaking, if a user expects to use the default value for a column parameter, he/she would not have to specify the column name, right?

Could you come up with a use-case for explicitly specifying the default column name for those column parameters? It is typically useful to specify values for parameters that users are likely to tune (e.g. input data count). IMO it is unlikely users would care the column names used in a benchmark. Thus it seems simpler to skip them. What do you think?

zhipeng93: Thanks for the detailed explanation. I agree that skipping the default column names are more friendly to users.

zhipeng93:
Thanks for the review @lindong28 . I have updated the PR in the latest PR. Please take a look.

lindong28:
Thanks for the update! LGTM.

PR #199, opened 2023-01-14
[demo-only] Refactor modules

lindong28:
This PR provides an example of refactoring flink-ml-core into two separate modules (i.e. flink-ml-core and flink-ml-servable-core), such that flink-ml-servable-core provides all the core functionalities needed to serve Flink ML model without depend on Flink runtime.


flink-ml-servable-core/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+    <groupId>org.apache.flink</groupId>
+    <artifactId>flink-ml-parent</artifactId>
+    <version>2.2-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>flink-ml-servable-core</artifactId>
+  <name>Flink ML : Servable : Core</name>
+
+  <dependencies>
+    <dependency>
```
Fanoid, line 35: It seems not to meet the requirements of excluding all Flink libraries here, since we still take flink-core and flink-shaded-jackson as dependencies.

lindong28: flinik-core is introduced here because we need to use the serializer/deserialize library in this module in the short term etc. for Flink ML 2.2 release. But since our public API does not depend on anything from flink-core or flink-shaded-jackson, we should be able to remove those dependencies without having any impact on users and thus achieve our goal in the long term.

Does this address your concern?

Fanoid: Hi, thanks for your elaboration about my concern.

IMHO, I think our public APIs are still depending on flink-core somehow:
The annotation `@PublicEvolving` is imported from flink-core, which I think is part of the public APIs.  The annotation `@TypeInfo` is also from flink-core. Without `@TypeInfo`, users have to manual register FlinkML types, like DenseVector, SparseVector, etc., and their `TypeInformation` s. (Maybe there are more such cases.)

So, maybe there is no need to make half-done changes public in a hurry, since they are likely to be modified significantly later (and affects users).  
Actually, what I assumed is to resolve above issues in one PR, otherwise another PR is still required.

lindong28: @Fanoid I think @PublicEvolving tags whether an API is public or not. We can change this to a class defined in flink-ml-servable without having any impact on user (i.e. users program can still compile with Flink ML API/library without issue).

The @TypeInfo is only used inside the flink-ml-servable implementation. In the future Flink should provide API for users to register custom serializer/deserializer for selected classes. Then we can use this API in flink-ml-core and flink-ml-servable-core won't need to depend on @TypeInfo.

Thus for the two cases described above, we are free to update flink-ml-servable in a backward compatible way such that flink-ml-servable-core can avoid depending on flink runtime. What do you think?

Fanoid: Hi, @lindong28 , I understand your point. For the issues mentioned above, I agree that we could figure out solutions sooner or later.  My concern is the compatibility issue that will be introduced in the future when resolving above issues.

Assuming current PR is merged (after reviewed),  people may start using `flink-ml-servable-core`. End-users are free to use `TypeInformation`s like `DenseVectorTypeInfo`, since they are public and necessary if they use `DenseVector` in Flink jobs. However, to achieve our goal, we have to eliminate `flink-core` from `flink-ml-servable-core` dependencies in future changes. The elimination is likely to lead to moving `DenseVectorTypeInfo` out from `flink-ml-servable-core`. As a consequence, this could breaks end-users' usages described above during this time.

Hope this can clarify my concern 😊.

lindong28: Thanks for the detailed explanation @Fanoid :) 

In the case you mentioned, user's code will be using a classA from a transitive Maven dependency without explicitly specifying this dependencies in pom.xml. This is a classic problem that has been discussed in various threads like this [1]. This issue is due to misbehavior from user side. The general solution as mentioned in [1] is that, "If your code depends on a module, your code should depends only directly on classes of this module, not a transitive-dependency of this module".

Note that it is not considered a public API change if we remove an internal dependency from a Flink's pom.xml. No FLIP is needed to make this change. I suppose we can also apply this rule for Flink ML?

[1] https://stackoverflow.com/questions/47260901/should-i-rely-on-transitive-dependencies-in-maven-if-they-come-from-other-sub-mo

Fanoid: Sorry I may missed something, as I can't figure out what the transitive dependency is in the case.  

If following the rule you cited, users have to **directly** specify `flink-ml-servable-core` as a dependency to use classes like `DenseVector`, `DenseVectorTypeInfo`, and so on.  If not following the rule, users still have the chance to **directly** depend on `flink-ml-servable-core` in their projects, because this module will be made public and be depended on without restrictions.

In both situations, after **directly** depending on `flink-ml-servable-core`, users may import `DenseVectorTypeInfo` in their code (because Flink requires type informations in some places). Note that there is no transitive dependency here.  Then comes the breaks if we move `DenseVectorTypeInfo` out of `flink-ml-servable-core` in future.

lindong28: > In both situations, after directly depending on flink-ml-servable-core, users may import DenseVectorTypeInfo in their code

Hmm... can you explain why user would import DenseVectorTypeInfo in their code? I think users would only need to use DenseVector to use Flink ML, right? Also note that DenseVectorTypeInfo is not marked `@public`.

BTW, I guess we need to mark classes such as `DenseVector` and `Vector` as `public`

lindong28: We say moduleA is transitively depending on moduleA, if moduleA is directly depending on moduleB in its pom.xml, and moduleB is directly depending on moduleC in its pom.xml.

Fanoid: > > In both situations, after directly depending on flink-ml-servable-core, users may import DenseVectorTypeInfo in their code
> 
> Hmm... can you explain why user would import DenseVectorTypeInfo in their code? I think users would only need to use DenseVector to use Flink ML, right? Also note that DenseVectorTypeInfo is not marked `@PublicEvolving`.
> 
> BTW, I guess we need to mark classes such as `DenseVector` and `Vector` as `public`

For example, the prediction `Row` of `LogisticRegressionModel` contains a field of type `DenseVector` (storing probability of being positive and negative).  It is possible that users call `flatMap(...).returns(Types.ROW(..., DenseVectorTypeInfo.INSTANCE, ...)` in their code to do some post-processing on the prediction results.

Fanoid: > We say moduleA is transitively depending on moduleC, if moduleA is directly depending on moduleB in its pom.xml, and moduleB is directly depending on moduleC in its pom.xml.

yeah. So in the two situations mentioned above, users are assumed to directly depend on `flink-ml-core-servable`, no matter following the cited rule or not.

lindong28: @Fanoid In the use-case mentioned above, users is using `LogisticRegressionModel`, which means they need to depend on `flink-ml-core` in their project's pom.xml. This means we can move `DenseVectorTypeInfo` from `flink-ml-servable-core` to `flink-ml-core` without breaking their code, right?

Fanoid: > n the use-case mentioned above, users are using `LogisticRegressionModel`, which means they need to depend on `flink-ml-core` in their project's pom.xml. This means we can move `DenseVectorTypeInfo` from `flink-ml-servable-core` to `flink-ml-core` without breaking their code, right?

I see your point now: it's meaningless to use `DenseVectorTypeInfo` solely without depending on `flink-ml-core`.  Actually, there is one rare situation that users call `DenseVectorTypeInfo#createSerializer` and use it just for serialization/deserialization of `DenseVector` with only depending on `flink-core`.  

Nevertheless, I agree that future movement of `DenseVectorTypeInfo` will hardly break end-users' codes.

lindong28: Cool. Thanks for all the comments!

BTW, with the current design of FLIP-289, users of TransformerServable will see types such as List, but not DenseVector. That means only users of Estimator/Transformer might need to call DenseVectorTypeInfo#createSerializer.

PR #202, opened 2023-02-02
[FLINK-30688] Disable Kryo fallback in Flink ML tests

jiangxin369:

Disable Kryo fallback in Flink ML tests to ensure that all Flink ML algorithms are not using Kryo.

## Brief change log

  - Configures StreamExecutionEnvironment using env.getConfig().disableGenericTypes().

## Documentation


flink-ml-core/src/test/java/org/apache/flink/ml/util/ReadWriteUtilsTest.java
```diff
@@ -54,7 +54,7 @@ public void before() throws IOException {
                 new org.apache.flink.configuration.Configuration();
         config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
         env = StreamExecutionEnvironment.getExecutionEnvironment(config);
-        env.getConfig().enableObjectReuse();
+        env.getConfig().enableObjectReuse().disableGenericTypes();
```
lindong28, line 57: Can we re-use TestUtils for consistency?

flink-ml-lib/src/main/java/org/apache/flink/ml/evaluation/binaryclassification/BinaryClassificationEvaluator.java
```diff
@@ -668,30 +669,64 @@ public Tuple3<Double, Boolean, Double> map(Row value) throws Exception {
...
+        public BinarySummary() {}
+
         public BinarySummary(Integer taskId, double maxScore, long curPositive, long curNegative) {
             this.taskId = taskId;
             this.maxScore = maxScore;
             this.curPositive = curPositive;
             this.curNegative = curNegative;
         }
+
+        public Integer getTaskId() {
```
lindong28, line 690: Would it be simpler to make these variables `public` than adding get/set methods?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizer.java
```diff
@@ -78,7 +80,11 @@ public CountVectorizerModel fit(Table... inputs) {
         DataStream<CountVectorizerModelData> modelData =
                 DataStreamUtils.aggregate(
                         inputData,
-                        new VocabularyAggregator(getMinDF(), getMaxDF(), getVocabularySize()));
+                        new VocabularyAggregator(getMinDF(), getMaxDF(), getVocabularySize()),
+                        Types.TUPLE(
+                                Types.LONG,
+                                Types.MAP(Types.STRING, Types.TUPLE(Types.LONG, Types.LONG))),
```
lindong28, line 86: It is surprising that `TypeExtractor.getAggregateFunctionAccumulatorType` can not automatically derive the right type from `Tuple2<Long, Map<String, Tuple2<Long, Long>>>`, which only contains common types.

Can you check with related developers whether this is expected? Maybe open a ticket in Flink to track this issue.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ImputerTest.java
```diff
@@ -265,9 +265,7 @@ public void testNoValidDataOnMedianStrategy() {
         final List<Row> trainData =
                 new ArrayList<>(
                         Arrays.asList(
-                                Row.of(Double.NaN, Float.NaN),
-                                Row.of(null, null),
-                                Row.of(1.0, 1.0f)));
+                                Row.of(Double.NaN, 3.0f), Row.of(null, 2.0f), Row.of(1.0, 1.0f)));
```
lindong28, line 268: Why do we need to change the value here as part of this PR?

lindong28 (commented):

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/ImputerTest.java
jiangxin369: Before the change, the 2 input columns both consisted of missing data, so the both 2 columns are invalid. And the aggregation summaries of columns are stored in a HashMap, which is not order-preserving, so we are not sure which column is detected first.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/countvectorizer/CountVectorizer.java
jiangxin369: IMO, since we are using generic types like Map, so the internal data types are erased during the compilation, and Flink doesn't know which TypeInformation to use in the Map.

Anyway, I agree to confirm with the related developers to find out if there's a better solution.

flink-ml-core/src/test/java/org/apache/flink/ml/api/GraphTest.java
```diff
@@ -49,13 +46,7 @@ public class GraphTest extends AbstractTestBase {
...
     @Before
     public void before() {
-        Configuration config = new Configuration();
-        config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
-        env = StreamExecutionEnvironment.getExecutionEnvironment(config);
-        env.getConfig().enableObjectReuse();
-        env.setParallelism(4);
-        env.enableCheckpointing(100);
-        env.setRestartStrategy(RestartStrategies.noRestart());
+        env = org.apache.flink.ml.util.TestUtils.getExecutionEnvironment();
```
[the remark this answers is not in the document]
jiangxin369: There may be some ambiguity here. The `org.apache.flink.ml.util.TestUtils` here is under `flink-ml-core`. We also have a `org.apache.flink.ml.util.TestUtils` under `flink-ml-lib`, and both of them contains a `getExecutionEnvironment` method. This cause some code duplicates but avoid introduce test-jar of `flink-ml-core` in `flink-ml-lib`. 

I'm combining the `TestUtil` in `org.apache.flink.ml.api` and `org.apache.flink.ml.util` under `flink-ml-core` to avoid using full package path.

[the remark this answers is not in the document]
jiangxin369: According to the offline discussion, we agreed to merge the `TestUtils` in `flink-ml-lib` to `flink-ml-core`, and make `flink-ml-lib` depends on the test-jar of `flink-ml-core` to reduce code duplicates.

jiangxin369:
@lindong28 Thanks for your review, I've updated the PR according to your comments.

lindong28 (commented):
Thanks for the update. LGTM overall. Left two minor comments.

PR #203, opened 2023-02-06
[FLINK-30688][followup] Disable Kryo fallback for tests in flink-ml-lib

jiangxin369:

Disable Kryo fallback in Flink ML tests to ensure that most Flink ML algorithms are not using Kryo.

With this PR, only the following algorithms are still using Kryo due to 2 types of reasons.

Iteration serialization issues:

- LogisticRegression
- OnlineLogisticRegression
- KMeans
- OnlineKMeans
- LinearSVC
- LinearRegression

Input rows with non-determined type:

- HashingTF

## Brief change log

*(for example:)*
  - Configures StreamExecutionEnvironment using env.getConfig().disableGenericTypes() in all Flink ML tests.
  - Add TypeInfoFactory for QuantileSummary
  - Add TypeInfoFactory for Vector

## Documentation

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/typeinfo/QuantileSummaryTypeInfoFactory.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/typeinfo/QuantileSummaryTypeInfoFactory.java
@@ -16,12 +16,13 @@
  * limitations under the License.
  */
 
-package org.apache.flink.ml.common.util.quantile;
+package org.apache.flink.ml.common.typeinfo;
 
 import org.apache.flink.api.common.typeinfo.TypeInfoFactory;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.api.java.typeutils.TypeExtractor;
+import org.apache.flink.ml.common.util.QuantileSummary;
 
 import java.lang.reflect.Type;
 import java.util.HashMap;
@@ -33,26 +34,14 @@ import java.util.Map;
  */
 public class QuantileSummaryTypeInfoFactory extends TypeInfoFactory<QuantileSummary> {
 
-    private static Map<String, TypeInformation<?>> statsTupleFields;
-
-    private static Map<String, TypeInformation<?>> fields;
-
-    static {
-        statsTupleFields = new HashMap<>();
-        statsTupleFields.put("value", Types.DOUBLE);
-        statsTupleFields.put("g", Types.LONG);
-        statsTupleFields.put("delta", Types.LONG);
-    }
-
-    private static final TypeInformation<QuantileSummary.StatsTuple> STATS_TUPLE_TYPE_INFO =
-            Types.POJO(QuantileSummary.StatsTuple.class, statsTupleFields);
+    private static final Map<String, TypeInformation<?>> fields;
 
     static {
         fields = new HashMap<>();
         fields.put("relativeError", Types.DOUBLE);
         fields.put("compressThreshold", Types.INT);
         fields.put("count", Types.LONG);
-        fields.put("sampled", Types.LIST(STATS_TUPLE_TYPE_INFO));
+        fields.put("sampled", Types.LIST(TypeInformation.of(QuantileSummary.StatsTuple.class)));
         fields.put("headBuffer", Types.LIST(Types.DOUBLE));
         fields.put("compressed", Types.BOOLEAN);
     }
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
@@ -16,9 +16,10 @@
  * limitations under the License.
  */
 
-package org.apache.flink.ml.common.util.quantile;
+package org.apache.flink.ml.common.util;
 
 import org.apache.flink.api.common.typeinfo.TypeInfo;
+import org.apache.flink.ml.common.typeinfo.QuantileSummaryTypeInfoFactory;
 import org.apache.flink.util.Preconditions;
 
 import java.io.Serializable;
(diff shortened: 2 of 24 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/quantile/QuantileSummary.java
```diff
@@ -395,12 +400,14 @@ public QueryResult(int index, long minRankAtIndex, double percentile) {
...
      */
-    private static class StatsTuple implements Serializable {
+    public static class StatsTuple implements Serializable {
         private static final long serialVersionUID = 1L;
-        private final double value;
+        private double value;
         private long g;
         private long delta;
 
+        public StatsTuple() {}
```
lindong28, line 409: Is this change needed to make this class POJO? If yes, then why its member variable is private?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/quantile/QuantileSummary.java
```diff
@@ -36,6 +37,7 @@
  * algorithm proposed in the paper: "Space-efficient Online Computation of Quantile Summaries" by
  * Greenwald, Michael and Khanna, Sanjeev. (https://doi.org/10.1145/375663.375670)
  */
+@TypeInfo(QuantileSummaryTypeInfoFactory.class)
```
lindong28, line 40: Would it be simpler to put both `QuantileSummary` and `QuantileSummaryTypeInfoFactory` directly under the package 
`common/util`?

Note that `QuantileSummaryTypeInfoFactory` should be an internal class that users should not need to access directly.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/lsh/LSHModel.java
```diff
@@ -373,34 +395,45 @@ public TopKFunction(String distCol, int numNearestNeighbors) {
...
-                }
+        public List<Row> add(Row value, List<Row> accumulator) {
+            if (topKRows == null) {
+                topKRows = new PriorityQueue<>(numNearestNeighbors, new DistColComparator(distCol));
+                topKRows.addAll(accumulator);
             }
-            accumulator.add(value);
-            return accumulator;
+            insert(value, topKRows);
+            return new ArrayList<>(topKRows);
```
lindong28, line 409: It would be pretty inefficient to copy the entire list every time we add a value.

How about we use `PriorityQueue<Row>` as the accumulator and let `getResult(List<Row> accumulator)` return `List<ROW>`? I think we will only need to serialize the output of `getResult` but not the accumulator itself.

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/lsh/LSHModel.java
lindong28: Note that if the aggregator needs to be serializable by either `kryo` or our custom serializer, we can define a specific container class that has the `PriorityQueue<ROW>` as member variable and define a custom serializer for it.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/quantile/QuantileSummary.java
jiangxin369: No, it isn't. We don't need to make its member variable public because I define the `TypeInfomation` for `StatsTuple` with `Types.POJO`. We need to

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
```diff
@@ -367,6 +373,26 @@ public double getRelativeError() {
         return relativeError;
     }
 
+    public int getCompressThreshold() {
```
lindong28, line 376: Is this needed to make this class POJO? Would it be simpler to still keep these member variables `public`?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/typeinfo/QuantileSummaryTypeInfoFactory.java
```diff
@@ -0,0 +1,57 @@
...
+        fields.put("relativeError", Types.DOUBLE);
+        fields.put("compressThreshold", Types.INT);
+        fields.put("count", Types.LONG);
+        fields.put("sampled", Types.LIST(TypeInformation.of(QuantileSummary.StatsTuple.class)));
+        fields.put("headBuffer", Types.LIST(Types.DOUBLE));
+        fields.put("compressed", Types.BOOLEAN);
+    }
+
+    private static final TypeInformation<QuantileSummary> TYPE_INFO =
+            Types.POJO(QuantileSummary.class, fields);
```
lindong28, line 50: Can we just use `TypeInformation.of(QuantileSummary.class)` here after making `QuantileSummary` POJO, instead of having to explicitly specifying the `fields` above?

And do we still need `QuantileSummaryTypeInfoFactory` if `QuantileSummary` is POJO?

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/common/typeinfo/QuantileSummaryTypeInfoFactory.java
jiangxin369: According to [1], a POJO must not contain fields that are not supported by registered serializers, while `QuantileSummary` has fields like `List<Double> headBuffer`, and `List` has no registered serializer and is treated as a generic type in Flink. So we still have to register the type info with `Types.POJO` and explicitly specify the fields.

[1] https://nightlies.apache.org/flink/flink-docs-master/docs/dev/datastream/fault-tolerance/serialization/types_serialization/

flink-ml-lib/src/main/java/org/apache/flink/ml/common/util/QuantileSummary.java
jiangxin369: No, it is for unittests currently. But I didn't annotate it with `VisibleForTesting` because I think it is reasonable for users to read the information of a `QuantileSummary`.

lindong28:
Thanks for the update. LGTM.

PR #205, opened 2023-02-07
[FLINK-30730] Fix null values handling in StringIndexer and StringIndexerModel

Fanoid:

This PR fixes null values handling in StringIndexer and StringIndexerModel.

## Brief change log

 - Skip null values in training, and handle null values in transforming.
 - Add tests for null values handling in corresponding test class.

## Documentation

### DIFF
--- a/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/stringindexer/StringIndexerTest.java
+++ b/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/stringindexer/StringIndexerTest.java
@@ -39,6 +39,8 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static org.junit.Assert.assertArrayEquals;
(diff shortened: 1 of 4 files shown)


flink-ml-lib/src/test/java/org/apache/flink/ml/feature/stringindexer/StringIndexerTest.java
```diff
@@ -53,6 +55,9 @@ public class StringIndexerTest extends AbstractTestBase {
     private Table trainTable;
     private Table predictTable;
 
+    private Table trainTableWithNull;
```
zhipeng93, line 58: Can you simplify the test by adding null values in `trainTable` and `predictTable`? In this case, the test case would be simpler.

Moreover, it would be great if we also update the python test.

Fanoid: Thanks for your suggestion. I've updated the PR.  Please check the changes.

zhipeng93:
Thanks for the update. LGTM.

PR #206, opened 2023-02-07
[FLINK-30933] Fix missing max watermark when executing join in iteration body

zhipeng93:

Currently, when executing join for bounded input streams in a perround iteration body, there exist two problems:
- The program may not exit even the result is collected using `datastream.executeAndCollect()`
- The max watermark of the output stream is losed.

This PR attempts to fix the above problems.

## Brief change log
- Set `Timer` thread as daemon in `HeadOperator`.
- Outputs maxwatermark when encountering max epoch watermark in `OutputOperator`.
- Added unit test for verify the change.

## Documentation

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
@@ -565,7 +565,7 @@ public class HeadOperator extends AbstractStreamOperator<IterationRecord<?>>
 
         private MailboxExecutorWithYieldTimeout(MailboxExecutor mailboxExecutor) {
             this.mailboxExecutor = mailboxExecutor;
-            this.timer = new Timer();
+            this.timer = new Timer(true);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


lindong28:
Thanks for the PR! LGTM.

@gaoyunhaii Can you help take another look at the change to the `OutputOperator`?

gaoyunhaii:
As a whole LGTM, and @zhipeng93 could you modify the test to also check the max watermark is indeed emitted before termination?

zhipeng93:
> As a whole LGTM, and @zhipeng93 could you modify the test to also check the max watermark is indeed emitted before termination?

Thanks for the reply. The unit test here [1] already verifies that the max watermark is emitted. If there is no max watermark, there should be no output and the unit test should fail. Do you mean we directly check that max watermark is emitted?

[1] https://github.com/apache/flink-ml/pull/206/files#diff-0d6c01f07abbec3796dfa449243437549c19615dd6439a27af08b6b68005e8a4R127

zhipeng93:

gaoyunhaii (approved):
LGTM, Thanks @zhipeng93 for the PR

PR #209, opened 2023-02-09
[FLINK-30688][followup] Disable Kryo fallback for tests in flink-ml-lib

lindong28:

Disable Kryo fallback for all tests (except `OnlineLogisticRegressionTest`) in flink-ml-lib to ensure that all Flink ML algorithms can run without relying on Kryo for serialization and deserialization.

In addition, this PR also updates `Benchmark#executeBenchmarks` to disable Kryo fallback to make sure that our benchmarks (including data generators) do not reply on Kryo.

## Brief change log

Disabled Kryo fallback in the following classes and methods:
- OnlineStandardScalerTest
- LogisticRegressionTest
- KMeansTest
- OnlineKMeansTest
- LinearSVCTest
- LinearRegressionTest
- HashingTFTest
- Benchmark#executeBenchmarks
- TableUtilsTest

## Documentation

### DIFF
--- a/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
+++ b/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.ml.benchmark;
 
+import org.apache.flink.api.common.restartstrategy.RestartStrategies;
 import org.apache.flink.ml.util.ReadWriteUtils;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
@@ -78,6 +79,8 @@ public class Benchmark {
 
         StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
         env.getConfig().enableObjectReuse();
+        env.getConfig().disableGenericTypes();
+        env.setRestartStrategy(RestartStrategies.noRestart());
         StreamTableEnvironment tEnv = StreamTableEnvironment.create(env);
 
         int index = 0;
--- a/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/HashingTFTest.java
+++ b/flink-ml-lib/src/test/java/org/apache/flink/ml/feature/HashingTFTest.java
@@ -18,13 +18,11 @@
 
 package org.apache.flink.ml.feature;
 
-import org.apache.flink.api.common.restartstrategy.RestartStrategies;
-import org.apache.flink.configuration.Configuration;
+import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.ml.feature.hashingtf.HashingTF;
 import org.apache.flink.ml.linalg.Vectors;
 import org.apache.flink.ml.util.TestUtils;
 import org.apache.flink.streaming.api.datastream.DataStream;
-import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.Expressions;
 import org.apache.flink.table.api.Table;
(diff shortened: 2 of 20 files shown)


lindong28:
@jiangxin369 @zhipeng93 Can you review this PR? Thanks.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/HashingTFTest.java
```diff
@@ -83,15 +81,9 @@ public class HashingTFTest extends AbstractTestBase {
...
-        config.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, true);
-        env = StreamExecutionEnvironment.getExecutionEnvironment(config);
-        env.getConfig().enableObjectReuse();
-        env.setParallelism(4);
-        env.enableCheckpointing(100);
-        env.setRestartStrategy(RestartStrategies.noRestart());
+        env = TestUtils.getExecutionEnvironment();
         tEnv = StreamTableEnvironment.create(env);
-        DataStream<Row> dataStream = env.fromCollection(INPUT);
+        DataStream<Row> dataStream = env.fromCollection(INPUT, Types.ROW(Types.LIST(Types.STRING)));
```
jiangxin369, line 86: So users need to specify the type info by themselves, right? if no type info is specified, then kryo is used.

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedAllRoundCheckpointITCase.java
```diff
@@ -131,16 +129,7 @@ static JobGraph createVariableAndConstantJobGraph(
...
-                        new Configuration() {
-                            {
-                                this.set(
-                                        ExecutionCheckpointingOptions
-                                                .ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH,
-                                        true);
-                            }
-                        });
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
         env.getConfig().enableObjectReuse();
```
jiangxin369, line 133: Would it be better if used `TestUtils.getExecutionEnvironment` here to disable kryo for these tests?

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
```diff
@@ -78,6 +79,8 @@ private static void executeBenchmarks(CommandLine commandLine) throws Exception
 
         StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
         env.getConfig().enableObjectReuse();
+        env.getConfig().disableGenericTypes();
```
jiangxin369, line 82: Do we need to `disableGenericTypes` in benchmark? First, not all algorithms work well with this config, e.g. HashingTF. Secondly, it is already guaranteed that kryo is disabled in unit tests of each algorithm.

jiangxin369 (commented):
@lindong28 Thanks for the PR, left some comments.

flink-ml-lib/src/test/java/org/apache/flink/ml/feature/HashingTFTest.java
lindong28: Yes that is correct.

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedAllRoundCheckpointITCase.java
lindong28: The purpose of FLINK-30688 is to ensure that all algorithms can be run without using kryo. And this can be verified by having all tests in flink-ml-lib use `TestUtils.getExecutionEnvironment` to disable kryo.

Now that all tests in flink-ml-lib (except OnlineLogisticRegression) have used `TestUtils.getExecutionEnvironment` to disable kryo, it seems that we don't have to use `TestUtils.getExecutionEnvironment ` for those tests in `flink-ml-iteration`. So I think it is simpler to keep those tests in `flink-ml-iteration` as is.

I removed the config `ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH` here because `ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH` is enabled by default since Flink 1.15.

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/Benchmark.java
lindong28: This change can additionally confirm that all DataGenerators have included TypeInformation properly so that the benchmarks can be run with kryo disabled.

Since the purpose of the benchmark is to get the maximum possible performance, I think we will always provide proper type information to make sure that the benchmark can be run with kryo disabled. This includes the benchmark for `HashingTF `.

jiangxin369: I see, thanks for the explanation.

zhipeng93:
Thanks for the PR. LGTM.

PR #210, opened 2023-02-10
[FLINK-31010] Add Transformer and Estimator for GBTClassifier and GBTRegressor

Fanoid:

Add Transformer and Estimator for GBTClassifier and GBTRegressor. They are expected to support most features provided by SparkML, except for some minor ones.

## Brief change log

 - Add implementation of gradient-boosting trees.
 - Add Transformer and Estimator for GBTClassifier and GBTRegressor.

  - Dependencies (does it add or upgrade a dependency): yes

## Documentation

  - Does this pull request introduce a new feature? yes
  - If yes, how is the feature documented? JavaDocs

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/gbtclassifier/GBTClassifier.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/classification/gbtclassifier/GBTClassifier.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.classification.gbtclassifier;
+
+import org.apache.flink.ml.api.Estimator;
+import org.apache.flink.ml.common.gbt.GBTModelData;
+import org.apache.flink.ml.common.gbt.GBTRunner;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.util.ParamUtils;
+import org.apache.flink.ml.util.ReadWriteUtils;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.table.api.Expressions.$;
+
+/** An Estimator which implements the gradient boosting trees classification algorithm. */
+public class GBTClassifier
+        implements Estimator<GBTClassifier, GBTClassifierModel>,
+                GBTClassifierParams<GBTClassifier> {
+
+    private final Map<Param<?>, Object> paramMap = new HashMap<>();
+
+    public GBTClassifier() {
+        ParamUtils.initializeMapWithDefaultValues(paramMap, this);
+    }
+
+    public static GBTClassifier load(StreamTableEnvironment tEnv, String path) throws IOException {
+        return ReadWriteUtils.loadStageParam(path);
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    @Override
+    public GBTClassifierModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<GBTModelData> modelData = GBTRunner.trainClassifier(inputs[0], this);
+        GBTClassifierModel model = new GBTClassifierModel();
+        model.setModelData(tEnv.fromDataStream(modelData).renameColumns($("f0").as("modelData")));
+        ReadWriteUtils.updateExistingParams(model, getParamMap());
+        return model;
+    }
+
+    @Override
+    public void save(String path) throws IOException {
+        ReadWriteUtils.saveMetadata(this, path);
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.common.gbt;
+
+import org.apache.flink.ml.classification.gbtclassifier.GBTClassifierModel;
+import org.apache.flink.ml.common.param.HasCategoricalCols;
+import org.apache.flink.ml.common.param.HasFeaturesCol;
+import org.apache.flink.ml.common.param.HasLabelCol;
+import org.apache.flink.ml.common.param.HasPredictionCol;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.StringArrayParam;
+
+/**
+ * Params of {@link GBTClassifierModel}.
+ *
+ * <p>If the input features come from 1 column of vector type, `featuresCol` should be used, and all
+ * features are treated as continuous features. Otherwise, `inputCols` should be used for multiple
+ * columns. Columns whose names specified in `categoricalCols` are treated as categorical features,
+ * while others are continuous features.
+ *
+ * <p>NOTE: `inputCols` and `featuresCol` are in conflict with each other, so they should not be set
+ * at the same time. In addition, `inputCols` has a higher precedence than `featuresCol`, that is,
+ * `featuresCol` is ignored when `inputCols` is not `null`.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface GBTModelParams<T>
+        extends HasFeaturesCol<T>, HasLabelCol<T>, HasCategoricalCols<T>, HasPredictionCol<T> {
+
+    Param<String[]> INPUT_COLS = new StringArrayParam("inputCols", "Input column names.", null);
+
+    default String[] getInputCols() {
+        return get(INPUT_COLS);
+    }
+
+    default T setInputCols(String... value) {
+        return set(INPUT_COLS, value);
+    }
+}
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
@@ -25,9 +25,10 @@ import org.apache.flink.ml.common.param.HasLabelCol;
 import org.apache.flink.ml.common.param.HasPredictionCol;
 import org.apache.flink.ml.param.Param;
 import org.apache.flink.ml.param.StringArrayParam;
+import org.apache.flink.ml.regression.gbtregressor.GBTRegressorModel;
 
 /**
- * Params of {@link GBTClassifierModel}.
+ * Params of {@link GBTClassifierModel} and {@link GBTRegressorModel}.
  *
  * <p>If the input features come from 1 column of vector type, `featuresCol` should be used, and all
  * features are treated as continuous features. Otherwise, `inputCols` should be used for multiple
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
@@ -1,0 +1,173 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.common.gbt;
+
+import org.apache.flink.api.common.functions.AggregateFunction;
+import org.apache.flink.api.common.functions.RichMapFunction;
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.iteration.DataStreamList;
+import org.apache.flink.iteration.IterationConfig;
+import org.apache.flink.iteration.IterationID;
+import org.apache.flink.iteration.Iterations;
+import org.apache.flink.iteration.ReplayableDataStreamList;
+import org.apache.flink.ml.common.broadcast.BroadcastUtils;
+import org.apache.flink.ml.common.datastream.DataStreamUtils;
+import org.apache.flink.ml.common.gbt.defs.FeatureMeta;
+import org.apache.flink.ml.common.gbt.defs.GbtParams;
+import org.apache.flink.ml.common.gbt.defs.LocalState;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.types.Row;
+
+import org.apache.commons.lang3.ArrayUtils;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+
+/** Runs a gradient boosting trees implementation. */
+public class GBTRunner {
+
+    /** Trains a gradient boosting tree model with given data and parameters. */
+    static DataStream<GBTModelData> train(Table dataTable, GbtParams p) {
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) dataTable).getTableEnvironment();
+        Tuple2<Table, DataStream<FeatureMeta>> preprocessResult =
+                p.isInputVector
+                        ? Preprocess.preprocessVecCol(dataTable, p)
+                        : Preprocess.preprocessCols(dataTable, p);
+        dataTable = preprocessResult.f0;
+        DataStream<FeatureMeta> featureMeta = preprocessResult.f1;
+
+        DataStream<Row> data = tEnv.toDataStream(dataTable);
+        DataStream<Tuple2<Double, Long>> labelSumCount =
+                DataStreamUtils.aggregate(data, new LabelSumCountFunction(p.labelCol));
+        return boost(dataTable, p, featureMeta, labelSumCount);
+    }
+
+    private static DataStream<GBTModelData> boost(
+            Table dataTable,
+            GbtParams p,
+            DataStream<FeatureMeta> featureMeta,
+            DataStream<Tuple2<Double, Long>> labelSumCount) {
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) dataTable).getTableEnvironment();
+
+        final String featureMetaBcName = "featureMeta";
+        final String labelSumCountBcName = "labelSumCount";
+        Map<String, DataStream<?>> bcMap = new HashMap<>();
+        bcMap.put(featureMetaBcName, featureMeta);
+        bcMap.put(labelSumCountBcName, labelSumCount);
+
+        DataStream<LocalState> initStates =
+                BroadcastUtils.withBroadcastStream(
+                        Collections.singletonList(
+                                tEnv.toDataStream(tEnv.fromValues(0), Integer.class)),
+                        bcMap,
+                        (inputs) -> {
+                            //noinspection unchecked
+                            DataStream<Integer> input = (DataStream<Integer>) (inputs.get(0));
+                            return input.map(
+                                    new InitLocalStateFunction(
+                                            featureMetaBcName, labelSumCountBcName, p));
+                        });
+
+        DataStream<Row> data = tEnv.toDataStream(dataTable);
+        final IterationID iterationID = new IterationID();
+        DataStreamList dataStreamList =
+                Iterations.iterateBoundedStreamsUntilTermination(
+                        DataStreamList.of(initStates.broadcast()),
+                        ReplayableDataStreamList.notReplay(data, featureMeta),
+                        IterationConfig.newBuilder()
+                                .setOperatorLifeCycle(IterationConfig.OperatorLifeCycle.ALL_ROUND)
+                                .build(),
+                        new BoostIterationBody(iterationID, p));
+        DataStream<LocalState> state = dataStreamList.get(0);
+        return state.map(GBTModelData::fromLocalState);
+    }
+
+    private static class InitLocalStateFunction extends RichMapFunction<Integer, LocalState> {
+        private final String featureMetaBcName;
+        private final String labelSumCountBcName;
+        private final GbtParams p;
+
+        private InitLocalStateFunction(
+                String featureMetaBcName, String labelSumCountBcName, GbtParams p) {
+            this.featureMetaBcName = featureMetaBcName;
+            this.labelSumCountBcName = labelSumCountBcName;
+            this.p = p;
+        }
+
+        @Override
+        public LocalState map(Integer value) {
+            LocalState.Statics statics = new LocalState.Statics();
+            statics.params = p;
+            statics.featureMetas =
+                    getRuntimeContext()
+                            .<FeatureMeta>getBroadcastVariable(featureMetaBcName)
+                            .toArray(new FeatureMeta[0]);
+            if (!statics.params.isInputVector) {
+                Arrays.sort(
+                        statics.featureMetas,
+                        Comparator.comparing(d -> ArrayUtils.indexOf(p.featureCols, d.name)));
+            }
+            statics.numFeatures = statics.featureMetas.length;
+            statics.labelSumCount =
+                    getRuntimeContext()
+                            .<Tuple2<Double, Long>>getBroadcastVariable(labelSumCountBcName)
+                            .get(0);
+            return new LocalState(statics, new LocalState.Dynamics());
+        }
+    }
+
+    private static class LabelSumCountFunction
+            implements AggregateFunction<Row, Tuple2<Double, Long>, Tuple2<Double, Long>> {
+
+        private final String labelCol;
+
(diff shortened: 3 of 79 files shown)


flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasLossType.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.StringParam;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for the shared maxBins param. */
+public interface HasLossType<T> extends WithParams<T> {
```
lindong28, line 27: This class is not used in this PR. Do we still need it?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
```diff
@@ -0,0 +1,247 @@
...
+        p.stepSize = estimator.getStepSize();
+        p.seed = estimator.getSeed();
+        p.subsamplingRate = estimator.getSubsamplingRate();
+        p.featureSubsetStrategy = estimator.getFeatureSubsetStrategy();
+        p.validationTol = estimator.getValidationTol();
+        p.gamma = estimator.getRegGamma();
+        p.lambda = estimator.getRegLambda();
+
+        if (TaskType.CLASSIFICATION.equals(p.taskType)) {
+            p.lossType = estimator.get(GBTClassifierParams.LOSS_TYPE);
```
lindong28, line 174: Would it be more readable to consistently get the parameter using get getXXX() API? For example, we can do `((GBTClassifier) estimator).getLossType()` here.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
```diff
@@ -0,0 +1,247 @@
...
+        p.maxIter = estimator.getMaxIter();
+        p.stepSize = estimator.getStepSize();
+        p.seed = estimator.getSeed();
+        p.subsamplingRate = estimator.getSubsamplingRate();
+        p.featureSubsetStrategy = estimator.getFeatureSubsetStrategy();
+        p.validationTol = estimator.getValidationTol();
+        p.gamma = estimator.getRegGamma();
+        p.lambda = estimator.getRegLambda();
+
+        if (TaskType.CLASSIFICATION.equals(p.taskType)) {
```
lindong28, line 173: Would it be simpler to use `estimator instanceof GBTClassifier`  instead of reading/writing an extra parameter `taskType` here?

Same for other parameters and methods (e.g. `trainClassifier()`) which can be replaced with the `instanceof`.

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/gbtclassifier/GBTClassifier.java
```diff
@@ -0,0 +1,75 @@
...
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    @Override
+    public GBTClassifierModel fit(Table... inputs) {
+        Preconditions.checkArgument(inputs.length == 1);
+        StreamTableEnvironment tEnv =
+                (StreamTableEnvironment) ((TableImpl) inputs[0]).getTableEnvironment();
+        DataStream<GBTModelData> modelData = GBTRunner.trainClassifier(inputs[0], this);
```
lindong28, line 64: It seems that both `GBTClassifier` and `GBTRegressor` are thin wrappers around `GBTRunner`. And `GBTRunner` is only used by these two classes.

Since we expect users to directly use `GBTClassifier` and `GBTRegressor`, would it be simpler to translate tests in this class into tests in e.g. `GBTClassifierTest` which uses only the public API of `GBTClassifier`?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
```diff
@@ -0,0 +1,247 @@
...
+                        DataStreamList.of(initTrainContext.broadcast()),
+                        ReplayableDataStreamList.notReplay(data, featureMeta),
+                        IterationConfig.newBuilder()
+                                .setOperatorLifeCycle(IterationConfig.OperatorLifeCycle.ALL_ROUND)
+                                .build(),
+                        new BoostIterationBody(iterationID, p));
+        return dataStreamList.get(0);
+    }
+
+    public static GbtParams fromEstimator(BaseGBTParams<?> estimator, TaskType taskType) {
```
lindong28, line 129: All information specified in `GbtParams` can obtained from `estimator` itself. Note that all existing Estimators get the parameters directly instead of having to pass them via a container (e.g. GbtParams).

Would it be simpler and more consistent with the existing code to follow the existing practice? Spark ML does not seem to need this container for parameters. Alternative, can you explain why GBT needs to follow a different approach?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasMinInfoGain.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.flink.ml.common.param;
+
+import org.apache.flink.ml.param.DoubleParam;
+import org.apache.flink.ml.param.Param;
+import org.apache.flink.ml.param.ParamValidators;
+import org.apache.flink.ml.param.WithParams;
+
+/** Interface for shared param minInfoGain. */
+public interface HasMinInfoGain<T> extends WithParams<T> {
```
lindong28, line 27: Since this is only used by `BaseGBTParams`, would it be simpler to define this parameter there instead of creating `HasMinInfoGain`?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/datastorage/IterationSharedStorage.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.flink.util.Preconditions;
+
+import org.apache.commons.collections.IteratorUtils;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** A shared storage across subtasks of different operators. */
+public class IterationSharedStorage {
```
lindong28, line 39: Since this class does not have gbt specific logic, we probably should move it to a package that does not belong to `gbt`.

Allowing operators of different subtask to communicate seems like a non-trivial infra work and is inconsistent with the existing Flink paradigm that operators mostly communicate via datastream edges.

Can you explain why we need this? It might be more efficient to discuss offline to understand the motivation and the API semantics provided by this class.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/defs/Split.java
```diff
@@ -0,0 +1,170 @@
...
+package org.apache.flink.ml.common.gbt.defs;
+
+import org.apache.flink.api.common.typeinfo.TypeInfo;
+import org.apache.flink.ml.common.gbt.typeinfo.SplitTypeInfoFactory;
+
+import org.eclipse.collections.impl.map.mutable.primitive.IntDoubleHashMap;
+
+import java.util.BitSet;
+
+/** Stores a split on a feature. */
```
lindong28, line 28: Does this class represents the same logical concept as `org.apache.spark.mllib.tree.model.Split`? If yes, would it be more intuitive to move it to a package similar to Spark's package path instead of putting it under `gbt` package?

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
```diff
@@ -0,0 +1,56 @@
...
+ * <p>NOTE: `inputCols` and `featuresCol` are in conflict with each other, so they should not be set
+ * at the same time. In addition, `inputCols` has a higher precedence than `featuresCol`, that is,
+ * `featuresCol` is ignored when `inputCols` is not `null`.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface GBTModelParams<T>
+        extends HasFeaturesCol<T>, HasLabelCol<T>, HasCategoricalCols<T>, HasPredictionCol<T> {
+
+    Param<String[]> INPUT_COLS = new StringArrayParam("inputCols", "Input column names.", null);
```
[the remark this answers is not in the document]
Fanoid: I'm OK to rename `HasInputCols` to `HasFeaturesCols`.

The reason to have both two parameters is related with the handling of categorical columns.

SparkML's implementation only supports a vector column as input, but it can store meta information in its schema [1], so the algorithm can distinguish between numerical and categorical values.

In Flink, there is no way to store meta information in the Table schema. So when end-users want to feed categorical values, our implementation can only accept the data as multiple columns, i.e., `HasFeaturesCols`.  Otherwise, `HasFeaturesCol` is used.

[1] https://github.com/apache/spark/blob/27ed89b7be5ebb91e4a0b106b1669a7867a6012d/mllib/src/main/scala/org/apache/spark/ml/classification/GBTClassifier.scala#L194

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
```diff
@@ -0,0 +1,56 @@
...
+ * while others are continuous features.
+ *
+ * <p>NOTE: `inputCols` and `featuresCol` are in conflict with each other, so they should not be set
+ * at the same time. In addition, `inputCols` has a higher precedence than `featuresCol`, that is,
+ * `featuresCol` is ignored when `inputCols` is not `null`.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface GBTModelParams<T>
+        extends HasFeaturesCol<T>, HasLabelCol<T>, HasCategoricalCols<T>, HasPredictionCol<T> {
```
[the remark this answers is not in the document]
Fanoid: It is a little complicated. I'll answer your questions one by one.

1. A more complete parameter list is in `BaseGBTParams`. `GBTModelParams` contains only parameters for the GBT models.

2. Generally, we support all algorithm-specific parameters from SparkML exception `maxMemoryInMB`, `cacheNodeIds`, and `checkpointInterval`.  (Note some parameters are added but corresponding functions are not implemented yet in this PR.) 

3. `HasVarianceImpurity` is not supported because SparkML actually does not use it [1]. But we may support it in the future.  `HasWeightCol` is added in the `BaseGBTParams` but corresponding function is not implemented now.

4. `HasCategoricalCols` is needed when multiple columns are used as input to help distinguish between numerical and categorical values.


[1] https://github.com/apache/spark/blob/27ed89b7be5ebb91e4a0b106b1669a7867a6012d/mllib/src/main/scala/org/apache/spark/ml/classification/GBTClassifier.scala#L118

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
```diff
@@ -0,0 +1,56 @@
...
+ * columns. Columns whose names specified in `categoricalCols` are treated as categorical features,
+ * while others are continuous features.
+ *
+ * <p>NOTE: `inputCols` and `featuresCol` are in conflict with each other, so they should not be set
+ * at the same time. In addition, `inputCols` has a higher precedence than `featuresCol`, that is,
+ * `featuresCol` is ignored when `inputCols` is not `null`.
+ *
+ * @param <T> The class type of this instance.
+ */
+public interface GBTModelParams<T>
```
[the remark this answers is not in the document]
Fanoid: Actually, there are no differences between the model parameters. But I can add `GBTClassifierModelParams` and `GBTRegressorModelParams`.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/loss/LogLoss.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.commons.math3.analysis.function.Sigmoid;
+
+/**
+ * The loss function for binary log loss.
+ *
+ * <p>The binary log loss defined as -y * pred + log(1 + exp(pred)) where y is a label in {0, 1} and
+ * pred is the predicted logit for the sample point.
+ */
+public class LogLoss implements Loss {
```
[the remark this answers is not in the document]
Fanoid: I didn't put `LogLoss` to the package `org.apache.flink.ml.common.lossfunc`, because all current loss functions in this package have assumed the data type as `LabeledPointWithWeight`.  But loss functions in GBT algorithms cannot use `LabeledPointWithWeight`.

Is it OK to mix loss functions with different data types?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasLossType.java
Fanoid: Thanks for your reminder. I'll remove it.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
Fanoid: SparkML's implementation has a class named `BoostingStrategy` [1] serving as a container like `GbtParams`.  I think its mainly purpose is to make training functions in `GBTRunner` able to be tested separately.

[1] https://github.com/apache/spark/blob/27ed89b7be5ebb91e4a0b106b1669a7867a6012d/mllib/src/main/scala/org/apache/spark/ml/classification/GBTClassifier.scala#L195

flink-ml-lib/src/main/java/org/apache/flink/ml/common/param/HasMinInfoGain.java
Fanoid: Good question! 

Actually, if more tree algorithms are planned in the future, `HasMinInfoGain` is then a common parameter. Same as many other parameters in GBT algorithms.   

BTW, in SparkML, these parameters are put in `DecisionTreeParams`. Maybe we can define `DecisionTreeParams`?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/defs/Split.java
Fanoid: If following SparkML's practice, we can add a package named `org.apache.flink.ml.tree`, and put all such classes and parameters mentioned above into this package. 

Are you OK with this?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/gbtclassifier/GBTClassifier.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.table.api.Expressions.$;
+
+/** An Estimator which implements the gradient boosting trees classification algorithm. */
```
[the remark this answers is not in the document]
Fanoid: Thanks for your advice. I'll improve the description.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/datastorage/IterationSharedStorage.java
Fanoid: Alright, we can discuss this issue offline.

The motivation is communicating data which is in same order of training samples via `DataStream`s is slow and memory intensive.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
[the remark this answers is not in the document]
Fanoid: Based on our offline discussion, I added `HasFeaturesCols` to cover two cases.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTModelParams.java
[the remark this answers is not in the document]
Fanoid: I've update the description of this PR with details about supported and unsupported features.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/datastorage/IterationSharedStorage.java
Fanoid: Based on our offline discussion, I've updated this part to a more complete infra. The entry point is `SharedStorageUtils#withSharedStorage` now.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/defs/Split.java
Fanoid: Sorry that I need to correct myself: the `Split` here is not a general definition: 
 - The `BinnedInstance` in `Split#shouldGoLeft(BinnedInstance binnedInstance)` is actually only applicable in GBT algorithms, as binning continuous features is a practice which normal decision trees algorithm don't do.
 - Same as `categoriesGoLeft` in `CategoricalSplit`.

I think it is better to improve these parts after more tree-based algorithms implemented.

Fanoid:
Hi, @lindong28 , thanks for your valuable comments. I've update the PR based on comments and offline discussions.  Please take a look.

lindong28 (commented):
Thanks for the update! Left some comments below.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/operators/CalcLocalSplitsOperator.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static final String SPLIT_FINDER_STATE_NAME = "split_finder";
+    private final String sharedStorageAccessorID;
+    // States of local data.
+    private transient ListStateWithCache<SplitFinder> splitFinderState;
+    private transient SplitFinder splitFinder;
+    private transient SharedStorageContext sharedStorageContext;
+
+    public CalcLocalSplitsOperator() {
+        sharedStorageAccessorID = getClass().getSimpleName() + "-" + UUID.randomUUID();
```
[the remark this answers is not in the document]
Fanoid: Actually, I've tried `StreamOperator#getOperatorID` before.  

Since the Operator ID cannot be obtained before execution, we are unable to specify the owner map of share data items when building graph. Without the owner map, it is difficult to control access in runtime.

flink-ml-core/src/main/java/org/apache/flink/ml/common/sharedstorage/SharedStorageContextImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    public <T extends AbstractStreamOperator<?> & SharedStorageStreamOperator> void initializeState(
+            T operator,
+            StreamingRuntimeContext runtimeContext,
+            StateInitializationContext context) {
+        String ownerId = operator.getSharedStorageAccessorID();
+        int subtaskId = runtimeContext.getIndexOfThisSubtask();
+        for (Map.Entry<ItemDescriptor<?>, String> entry : ownerMap.entrySet()) {
+            ItemDescriptor descriptor = entry.getKey();
+            if (ownerId.equals(entry.getValue())) {
+                writers.put(
```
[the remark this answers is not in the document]
Fanoid: There is no instances to be overwritten here, because every subtask owns a distinct instance of `SharedStorageContextImpl`.

The field of `context` in operators are serialized and then deserialized to different instances when operators being deployed. It just like other fields.

flink-ml-core/src/main/java/org/apache/flink/ml/common/sharedstorage/SharedStorage.java
```diff
@@ -0,0 +1,172 @@
...
+                String ownerId,
+                TypeSerializer<T> serializer,
+                StreamTask<?, ?> containingTask,
+                StreamingRuntimeContext runtimeContext,
+                StateInitializationContext stateInitializationContext,
+                OperatorID operatorID) {
+            super(t);
+            this.ownerId = ownerId;
+            try {
+                cache =
```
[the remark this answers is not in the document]
Fanoid: I'm not quite sure what you mean by your suggestion. Are you suggesting to put the in-memory value in the `Writer` instance to avoid the lock?  

Following the previous reply, every subtask owns a distinct instance of `SharedStorageContextImpl`. So, the `Writer` instance stored in `SharedStorageContextImpl` is only visible to its owner subtask of one operator. It cannot be accessed by subtasks of other operators.

flink-ml-core/src/main/java/org/apache/flink/ml/common/sharedstorage/SharedStorage.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.flink.util.Preconditions;
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** A shared storage to support access through subtasks of different operators. */
+class SharedStorage {
+    private static final Map<Tuple3<StorageID, Integer, String>, Object> m =
```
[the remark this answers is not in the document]
Fanoid: Sure! I'll rename variables in this class.

For the latter issue, I prefer to consider `SharedStorage` as a container of **references** to data items, and provide sharing utility among subtasks of operators. It is still the owner of data items to worry memory issues.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/operators/CalcLocalSplitsOperator.java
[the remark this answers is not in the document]
lindong28: It seems that there is a way to remove sharedStorageAccessorID and still pass all tests. We can discuss the code change offline.

flink-ml-core/src/main/java/org/apache/flink/ml/common/sharedstorage/SharedStorageContextImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.flink.util.function.BiConsumerWithException;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
+/** Default implementation of {@link SharedStorageContext} using {@link SharedStorage}. */
+@SuppressWarnings("rawtypes")
+class SharedStorageContextImpl implements SharedStorageContext, Serializable {
+    private final StorageID storageID;
```
lindong28, line 35: Instead of keeping the `storageID` here, an alternative approach is generate the storageId once during graph building phase and passes it to all operators. Then each operator can pass the `storageID` to `initializeState`.

The storageId can be pass to operators via either the constructor or the setGlobalJobParameters/getGlobalJobParameters of ExecutionConfig.

If we can do this, we might be able to simplify the code and remove e.g. `SharedStorageWrapper`.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
```diff
@@ -0,0 +1,304 @@
...
+                            DataStream<Integer> input = (DataStream<Integer>) (inputs.get(0));
+                            return input.map(
+                                    new InitTrainContextFunction(
+                                            featureMetaBcName, labelSumCountBcName, strategy));
+                        });
+
+        DataStream<Row> data = tEnv.toDataStream(dataTable);
+        DataStreamList dataStreamList =
+                Iterations.iterateBoundedStreamsUntilTermination(
+                        DataStreamList.of(initTrainContext.broadcast()),
```
zhipeng93, line 171: This `broadcast` in `initTrainContext#broadcast` is supposed to be removed during compilation. Is this expected?

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
```diff
@@ -0,0 +1,304 @@
...
+                        ReplayableDataStreamList.notReplay(data, featureMeta),
```
zhipeng93, line 172: The `featureMeta` seems never used later. Can you remove it?

flink-ml-core/src/main/java/org/apache/flink/ml/common/sharedobjects/SharedObjectsStreamOperator.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.common.sharedobjects;
+
+/** Interface for all operators that need to access the shared objects. */
+public interface SharedObjectsStreamOperator {
```
zhipeng93, line 22: As I understand, this PR tries to provide an infrastructure for sharing objects among multiple Flink operators, through java static variables. 

To achieve this, it empoys one specific Flink operator for each sharing object as the writer and others operators as the reader. Based on this, the GBDT implementation relies on the Flink events to guarantee the read/write order of each object. 

However, can you explain some other machine learning algorithms that would use `SharedObjects` in the future? And is there a general way that developers can guarantee the order of read/writes is correct? If a reader of an object changes the value of that object, does it still follows the assumption of `SharedObjects`?

There is another possible solution [1] that we put all the computation logic into one operator (i.e., WorkerOperator) and all the computation logic into another operator (i.e., ServerOperator). In this case, we would not need shared objects anymore. Let's have a thorough comparison between these two options.

[1] https://github.com/apache/flink-ml/pull/237

zhipeng93 (commented):
Thanks for the PR. I left some comments about the `SharedObjects` infra here.

flink-ml-core/src/main/java/org/apache/flink/ml/common/sharedobjects/SharedObjectsStreamOperator.java
Fanoid: Thanks for your valuable comments. You mentioned several issues in your comments, and I will answer them one by one.

Q: "can you explain some other machine learning algorithms that would use SharedObjects in the future?"

One common pattern where SharedObjects can be used is same datasets are needed in operators before and after a `reduce` operator.  Here I list some algorithms (correct me if I made a mistake):
‒  All distributed algorithms based on decision trees: both model and training data are required during the nodes splitting after reducing the intermediate data.
‒ Second-order gradient optimizer, e.g. Newton, L-BFGS: require the use of old gradient data after the reduce operation.
‒ Algorithms that have two rounds of (unmergeable) `reduce` operation in each iteration: GBDT, ALS, GMM, and LDA.
‒ Evaluation metrics are calculated every few rounds of iterations: evaluation metrics has to be calculated with `reduce` after model is updated.

Q: "is there a general way that developers can guarantee the order of read/writes is correct?"

A simple approach is to connect the reader and writer with an addition dummy stream.  Readers read the data only after receiving elements from the stream.
This approach assumes the reads and writers are not interleaved which should be true in most algorithms. If not (please give some examples), multi-threading techniques like atomics can be used.

Q："If a reader of an object changes the value of that object, does it still follows the assumption of SharedObjects?"

I think this is a common issue in Java, and one solution is to give readers a deep clone of the object, which can be expensive. Balancing efficiency and safety, I chose efficiency.

Fanoid: #237 proposes a brilliant solution about abstracting iterative computations inspired from Parameter Server. I believe both solutions can work well in many algorithms/scenarios.

Before comparing two solutions, there are two facts about PS infra I must emphasize:

One fact is the PS infra is built on the DataStream APIs, which means there will be no performance improvement compared to implement with raw DataStream APIs. So we mainly discuss its usability with aspect to developers.

The other fact is the current status of functionalities shown in #237 cannot fully meet the requirements of GBDT implementation.  `MessageType`, model format, reduce logic of messages, etc. are all fixed/hard-coded with respect to gradient-based algorithms. The usability will drop significantly if forcing GBDT implementation to use current APIs.

Therefore, to make a reasonable comparison between two solutions, I assume an extended version of current PS infra which supports POJO message types and POJO model data, user-defined `reduce` function, etc. Here are my thoughts under this assumption:

1. Framework Intrusiveness
Using PS infra means developers cannot use DataStream APIs in iterations anymore. Then there are cases where PS infra cannot implement: 
  - side outputs: evaluation result streams; prediction and model streams in online cases.
  - partition/join/coGroup of training data sets: AUC calculation after model update, ALS, SimRank.

As for SharedObjects, it is an augment to DataStream APIs. There is no extra limitation to developers.

The intrusiveness also influences the observation of operators when job running as, in PS infra, multiple computations are merged in to one operator, like in/out stats, checkpoint status. This decreases usability to both developers and end-users.

2. Applicable scenarios
Besides inapplicable cases mentioned above, PS infra cannot work in non-iteration cases. But SharedObjects can work. One possible case is to improve consecutive joins with a same dataset by reducing a copy of dataset.

3. Learning curve
PS infra provides a whole set of concepts and interfaces such as Message, ModelUpdater, ProcessStage, TrainingUtils, etc., which are not related to the existing DataStream API and have a steeper learning curve.

SharedObjects provides two interfaces, SharedObjectsUtils and SharedObjectsContext, and can be developed directly based on the existing DataStream API code, making it easier for developers to accept.


Overall speaking, I think both solutions can coexist because they are on different levels of APIs and have no conflicts. How about you? @zhipeng93

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
Fanoid: Thanks for your remind. I've moved the broadcast to iteration body.

flink-ml-lib/src/main/java/org/apache/flink/ml/common/gbt/GBTRunner.java
Fanoid: No problem.

PR #211, opened 2023-02-13
[FLINK-31026] Fix KBinsDiscretizer when the input values of one column are same

Fanoid:

KBinsDiscretizer produces `{Double.MIN_VALUE, Double.MAX_VALUE}` for columns whose input values are same which is not correct, as it doesn't cover negative values and 0. A more reasonable value should be `{Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY}`.

## Brief change log

  - Change bin edges for columns whose input values are same to `{Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY}`.

## Documentation


Fanoid:
Hi, @zhipeng93 could you help review this commit?

zhipeng93:
Thanks for the fix. LGTM.

PR #212, opened 2023-02-17
[FLINK-31125] Flink ML benchmark framework should minimize the source operator overhead

lindong28:

Minimize the source operator overhead in Flink ML benchmark and optimize the bucketizer performance.

Here is the total execution time of the Bucketizer benchmark for 3 different setups. The execution time is obtained by taking the average execution time across 5 runs for each setup.

- Execution time is 33.26 sec prior to this PR.
- Execution time is 31.14 sec after optimizing the benchmark framework.
- Execution time is 30.88 sec after optimizing both the benchmark framework and the Bucketizer.

Thus the benchmark framework optimization reduces the total time by 6.4%. The Bucketizer reduces the total time by another 0.83%.

## Brief change log

- Updated RowGenerator to pre-generate at least 1/000 of the expected number of rows  and re-use those rows as the source function output.
- Renamed RowGenerator#nextRow to RowGenerator#getRow. This is because there is no order of rows emitted by `nextRow`. 
- Updated Bucketizer#FindBucketFunction to fetch values from input rows using column index.

## Documentation

### DIFF
--- a/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DenseVectorArrayGenerator.java
+++ b/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DenseVectorArrayGenerator.java
@@ -42,7 +42,7 @@ public class DenseVectorArrayGenerator extends InputTableGenerator<DenseVectorAr
         return new RowGenerator[] {
             new RowGenerator(getNumValues(), getSeed()) {
                 @Override
-                protected Row nextRow() {
+                protected Row getRow() {
                     DenseVector[] result = new DenseVector[arraySize];
                     for (int i = 0; i < arraySize; i++) {
                         result[i] = new DenseVector(vectorDim);
--- a/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DenseVectorGenerator.java
+++ b/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DenseVectorGenerator.java
@@ -41,7 +41,7 @@ public class DenseVectorGenerator extends InputTableGenerator<DenseVectorGenerat
             new RowGenerator(getNumValues(), getSeed()) {
 
                 @Override
-                protected Row nextRow() {
+                protected Row getRow() {
                     double[] values = new double[vectorDim];
                     for (int i = 0; i < values.length; i++) {
                         values[i] = random.nextDouble();
--- a/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DoubleGenerator.java
+++ b/flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DoubleGenerator.java
@@ -59,7 +59,7 @@ public class DoubleGenerator extends InputTableGenerator<DoubleGenerator> {
         return new RowGenerator[] {
             new RowGenerator(getNumValues(), getSeed()) {
                 @Override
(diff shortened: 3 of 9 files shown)


lindong28:
@zhipeng93 Can you help review this PR?

flink-ml-benchmark/src/main/java/org/apache/flink/ml/benchmark/datagenerator/common/DenseVectorArrayGenerator.java
```diff
@@ -42,7 +42,7 @@ protected RowGenerator[] getRowGenerators() {
         return new RowGenerator[] {
             new RowGenerator(getNumValues(), getSeed()) {
                 @Override
-                protected Row nextRow() {
+                protected Row getRow() {
```
zhipeng93, line 45: Reusing one row to simulate the input may cause the benchmark result not stable, since processing two different rows may take different time.

The two optimizations introduced in this PR give us 12% speedup. Is there a breakdown analysis of this two optimization?

If the data generation cost is indeed high, can we sample a fixed number of rows and reuse it later?

lindong28: Good point. I have updated the PR as suggested. The RowGenerator would generate up to 64 rows and re-use those rows as the source function output.

I have updated the PR description with benchmark results for 3 different setups. It shows the percentage improvement provided by each optimization. The overall execution time is smaller than what I observed last time and the percentage difference is smaller.

zhipeng93: Thanks for the update. As generating 10^8 rows takes ~2s, do you think generating data is still a big overhead that needs to be addressed? Since reusing some of the rows does introduce some stability issues.

lindong28: I think we need to make this optimization since 2 sec still a big overhead. And previously I have seen bigger difference (more than 3 sec) after we make this optimization.

We expect Flink ML's performance to be close to Spark ML. That means we expect Flink ML's total time for this benchmark to be less than 6 sec after we finish all optimizations. Then 2 sec is going to take a big percentage of the total time.

Can you explain why this PR will introduce stability issues? Did you observe additional stability issues with this PR?

lindong28: I have updated the PR such that it will pre-generate 1/1000 of the expected number of rows and re-use those rows as the source function output. This allows the number of unique values to increase with the total number of expected values. Can you see if this address the concern?

zhipeng93: Thanks for the update. Setting the number of pre-generated data as 1/1000 does reduce the variance of the benchmark result here. LGTM.

zhipeng93:
Thanks for the update. There are many other data generators in the benchmark settings, do you think we should apply same optimization here?

lindong28:
@zhipeng93 Thanks for the review. The change made in this PR should be able to improve all existing data generators via `RowGenerator`. Can you check if there is anything missing?

zhipeng93:
> @zhipeng93 Thanks for the review. The change made in this PR should be able to improve all existing data generators via `RowGenerator`. Can you check if there is anything missing?

Ah my bad... Thanks for the explanation. LGTM.

PR #214, opened 2023-02-19
[FLINK-31127] Add public API classes for FLIP-289

lindong28:

Add public API classes for FLIP-289 with the exception specified below.

## Brief change log

Added most public API classes specified in FLIP-289.

The following class and methods are not added:
- GraphModel#supportServable
- PipelineModel#supportServable
- PipelineModelServable

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes

## Documentation

  - Does this pull request introduce a new feature? yes
  - If yes, how is the feature documented? JavaDocs

### DIFF
--- a/flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
+++ b/flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.servable.api;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.ml.servable.types.DataType;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * A DataFrame consists of several rows, each of which has the same column names and data types.
+ *
+ * <p>Values in the same column must have the same data type: integer, float, string etc.
+ */
+@PublicEvolving
+public class DataFrame {
+
+    private final List<String> columnNames;
+    private final List<DataType> dataTypes;
+    private final List<Row> rows;
+
+    /**
+     * The given columnNames and dataTypes should be mutable in order for TransformerServable
+     * classes to update DataFrame with the serving results.
+     */
+    public DataFrame(List<String> columnNames, List<DataType> dataTypes, List<Row> rows) {
+        int numColumns = columnNames.size();
+        if (dataTypes.size() != numColumns) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "The number of data types %d is different from the number of column names %d.",
+                            dataTypes.size(), numColumns));
+        }
+        for (Row row : rows) {
+            if (row.size() != numColumns) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "The row size %d is different from the number of column names %d.",
+                                row.size(), numColumns));
+            }
+        }
+
+        this.columnNames = columnNames;
+        this.dataTypes = dataTypes;
+        this.rows = rows;
+    }
+
+    /** Returns a list of the names of all the columns in this DataFrame. */
+    public List<String> getColumnNames() {
+        return columnNames;
+    }
+
+    /**
+     * Returns the index of the column with the given name.
+     *
+     * @throws IllegalArgumentException if the column is not present in this table
+     */
+    public int getIndex(String name) {
+        int index = columnNames.indexOf(name);
+        if (index == -1) {
+            throw new IllegalArgumentException(
+                    String.format("Failed to find the column with the name %s.", name));
+        }
+        return index;
+    }
+
+    /**
+     * Returns the data type of the column with the given name.
+     *
+     * @throws IllegalArgumentException if the column is not present in this table
+     */
+    public DataType getDataType(String name) {
+        int index = getIndex(name);
+        return dataTypes.get(index);
+    }
+
+    /**
+     * Adds to this DataFrame a column with the given name, data type, and values.
+     *
+     * @throws IllegalArgumentException if the number of values is different from the number of
+     *     rows.
+     */
+    public DataFrame addColumn(String columnName, DataType dataType, List<Object> values) {
+        if (values.size() != rows.size()) {
+            throw new RuntimeException(
+                    String.format(
+                            "The number of values %d is different from the number of rows %d.",
+                            values.size(), rows.size()));
+        }
+        columnNames.add(columnName);
+        dataTypes.add(dataType);
+
+        Iterator<Object> iter = values.iterator();
+        for (Row row : rows) {
+            Object value = iter.next();
+            row.add(value);
+        }
+        return this;
+    }
+
+    /** Returns all rows of this table. */
+    public List<Row> collect() {
+        return rows;
+    }
+}
--- a/flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/ModelServable.java
+++ b/flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/ModelServable.java
@@ -16,25 +16,23 @@
  * limitations under the License.
  */
 
-package org.apache.flink.ml.linalg;
+package org.apache.flink.ml.servable.api;
 
 import org.apache.flink.annotation.PublicEvolving;
 
-import java.io.Serializable;
+import java.io.IOException;
+import java.io.InputStream;
 
-/** A matrix of double values. */
+/**
+ * A ModelServable is a TransformerServable with the extra API to set model data.
+ *
+ * @param <T> The class type of the ModelServable implementation itself.
+ */
 @PublicEvolving
-public interface Matrix extends Serializable {
+public interface ModelServable<T extends ModelServable<T>> extends TransformerServable<T> {
 
-    /** Gets number of rows. */
-    int numRows();
-
-    /** Gets number of columns. */
-    int numCols();
-
-    /** Gets value of the (i,j) element. */
-    double get(int i, int j);
-
-    /** Converts the instance to a dense matrix. */
-    DenseMatrix toDense();
+    /** Sets model data using the serialized model data from the given input streams. */
+    default T setModelData(InputStream... modelDataInputs) throws IOException {
+        throw new UnsupportedOperationException("This operation is not supported.");
+    }
 }
--- a/flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/TransformerServable.java
+++ b/flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/TransformerServable.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.servable.api;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.ml.param.WithParams;
+
+/**
+ * A TransformerServable takes a DataFrame as input and produces a DataFrame as the result. It can
+ * be used to encode online inference computation logic.
+ *
+ * <p>NOTE: Every TransformerServable subclass should have a no-arg constructor.
+ *
+ * <p>NOTE: Every TransformerServable subclass should implement a static method with signature
+ * {@code static T load(String path) throws IOException;}, where {@code T} refers to the concrete
+ * subclass. This static method should instantiate a new TransformerServable instance based on the
+ * data read from the given path.
+ *
+ * @param <T> The class type of the TransformerServable implementation itself.
+ */
+@PublicEvolving
+public interface TransformerServable<T extends TransformerServable<T>> extends WithParams<T> {
+    /**
+     * Applies the TransformerServable on the given input DataFrame and returns the result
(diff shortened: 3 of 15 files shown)


lindong28:
@jiangxin369 Can you help review this PR?

jiangxin369 (commented):
@lindong28 Thanks for the PR, left some comments.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/ModelServable.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * A ModelServable is a TransformerServable with the extra API to set model data.
+ *
+ * @param <T> The class type of the ModelServable implementation itself.
+ */
+@PublicEvolving
+public interface ModelServable<T extends ModelServable<T>> extends TransformerServable<T> {
+
+    /** Sets model data using the serialized model data from the given input stream. */
+    default T setModelData(InputStream modelData) {
```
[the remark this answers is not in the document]
lindong28: Good point. I agree we need to make setModelData() take multiple InputSreams. And we need to let these methods throw IOException.

I will send an email update to the voting thread when this PR is mostly ready for merge.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
```diff
@@ -0,0 +1,120 @@
...
+     * Returns the index of the column with the given name.
+     *
+     * @throws IllegalArgumentException if the column is not present in this table
+     */
+    public int getIndex(String name) {
+        for (int i = 0; i < columnNames.size(); i++) {
+            if (columnNames.get(i).equals(name)) {
+                return i;
+            }
+        }
```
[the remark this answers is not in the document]
lindong28: Good point. I have updated the PR as suggested.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
```diff
@@ -0,0 +1,120 @@
...
+    public DataFrame(List<String> columnNames, List<DataType> dataTypes, List<Row> rows) {
+        int numColumns = columnNames.size();
+        if (dataTypes.size() != numColumns) {
+            throw new IllegalArgumentException(
+                    "The number of data types is different from the number of column names.");
+        }
+        for (Row row : rows) {
+            if (row.size() != numColumns) {
+                throw new IllegalArgumentException(
+                        "The number of values in some row is different from the number of column names");
```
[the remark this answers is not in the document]
lindong28: Sure. I have updated the PR as suggested.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
```diff
@@ -0,0 +1,120 @@
...
+     * @throws IllegalArgumentException if the number of values is different from the number of
+     *     rows.
+     */
+    public DataFrame addColumn(String columnName, DataType dataType, List<Object> values) {
+        if (values.size() != rows.size()) {
+            throw new RuntimeException(
+                    "The number of values is different from the number of rows.");
+        }
+        columnNames.add(columnName);
+        dataTypes.add(dataType);
```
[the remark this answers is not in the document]
lindong28: I am not sure it is worth adding additional code to do this check. Note that there is no need to differentiate between graph build time and graph execution time since there is no Flink runtime involved in using DataFrame. 

Thus the behavior observed by users will be the same regardless of whether we do this check before add elements in this List -- users will see exception throw by flink-ml-servable classes.

I added the following statement to the Java doc of `DataFrame`. Does this address your concern?

```
 * NOTE: the columnNames and dataTypes provided to the DataFrame constructor should be immutable in order for
 * TransformerServable classes to update DataFrame with the serving results.
```

lindong28:
@jiangxin369 Thanks for the review. Can you take another look?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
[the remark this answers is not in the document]
jiangxin369: The benefit of doing this check is that we can give users a more intuitive error message instead of `UnsupportedOperationException`, but I agree it is not necessary.

Adding docs also addresses my concern, but you need to change the `should be immutable` to `should be mutable`and also add such docs for `Row` constructor.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/ModelServable.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A ModelServable is a TransformerServable with the extra API to set model data.
+ *
+ * @param <T> The class type of the ModelServable implementation itself.
+ */
+@PublicEvolving
+public interface ModelServable<T extends ModelServable<T>> extends TransformerServable<T> {
+
+    /** Sets model data using the serialized model data from the given input stream. */
```
jiangxin369, line 34:
```suggestion
    /** Sets model data using the serialized model data from the given input streams. */
```

jiangxin369 (commented):
@lindong28 Thanks for the updates. LGTM overall only left 2 minor comments.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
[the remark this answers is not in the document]
lindong28: Thanks for catching this. It is fixed now.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/ModelServable.java
lindong28: Thanks for catching this. It is fixed now.

lindong28:
@zhipeng93 Can you help review this PR? If this PR looks good to you, I will need to send an email to the FLIP-289 voting thread to discuss the proposed API change (e.g. let `setModelData()` take multiple input streams) before this PR is merged.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
```diff
@@ -0,0 +1,132 @@
...
+package org.apache.flink.ml.servable.api;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.ml.servable.types.DataType;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * A DataFrame consists of some number of rows, each of which has the same list of column names and
```
zhipeng93, line 28: nit: 
`A DataFrame consists of several rows, each of which has the same column names and data types.`

`Values in the same column must have the same data type: integer, float, string etc.`

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Returns the index of the column with the given name.
+     *
+     * @throws IllegalArgumentException if the column is not present in this table
+     */
+    public int getIndex(String name) {
+        int index = columnNames.indexOf(name);
+        if (index == -1) {
+            throw new IllegalArgumentException("Failed to find the column with the given name.");
```
zhipeng93, line 79: nit: Shall we also print the given name for better debuging?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
```diff
@@ -0,0 +1,132 @@
...
+                            "The number of values %d is different from the number of rows %d.",
+                            values.size(), rows.size()));
+        }
+        columnNames.add(columnName);
+        dataTypes.add(dataType);
+
+        int rowSize = -1;
+        Iterator<Object> iter = values.iterator();
+        for (Row row : rows) {
+            if (rowSize < 0) {
```
zhipeng93, line 113: The check for row size here seems uncessary since we already checked them in the constructor.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/ModelServable.java
```diff
@@ -0,0 +1,38 @@
...
+ * A ModelServable is a TransformerServable with the extra API to set model data.
+ *
+ * @param <T> The class type of the ModelServable implementation itself.
+ */
+@PublicEvolving
+public interface ModelServable<T extends ModelServable<T>> extends TransformerServable<T> {
+
+    /** Sets model data using the serialized model data from the given input streams. */
+    default T setModelData(InputStream... modelDataInputs) throws IOException {
+        throw new UnsupportedOperationException("this operation is not supported");
```
zhipeng93, line 36: nit: this -> This

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/TransformerServable.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param <T> The class type of the TransformerServable implementation itself.
+ */
+@PublicEvolving
+public interface TransformerServable<T extends TransformerServable<T>> extends WithParams<T> {
+    /**
+     * Applies the TransformerServable on the given input DataFrame and returns the result
+     * DataFrame.
+     *
+     * @param input the input data
```
zhipeng93, line 43: nit: Let's capitalize the first character here.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/types/MatrixType.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.flink.ml.servable.types;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * A DataType representing a map where all keys have the same BasicType and all values have the same
+ * BasicType.
+ */
+@PublicEvolving
+public final class MatrixType extends DataType {
```
zhipeng93, line 28: Should we distinguish DataType for SparseMatrix/DenseMatrix and SparseVector/DenseVector?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/types/MatrixType.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.servable.types;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * A DataType representing a map where all keys have the same BasicType and all values have the same
```
zhipeng93, line 24: The java doc seems wrong here.

zhipeng93 (commented):

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
lindong28: Thanks. It is fixed now.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
lindong28: Thanks. It is fixed now.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/TransformerServable.java
lindong28: Do you mean to change it to `@param input The input data`?

I am not sure whether it is the right stye to use. The examples provided in [1] [2] as well as the existing doc for e.g. `AlgoOperator#transform` uses the lower case character. And the usage is also inconsistent in Flink codebase. For example, `DataStream#transform` uses lower case character.

How about we use the lower case for now and have a followup PR to change the existing Java doc later?

[1] https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html
[2] https://www.baeldung.com/javadoc

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/api/DataFrame.java
lindong28: Thanks for catching this! It is fixed now.

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/types/MatrixType.java
lindong28: For now, I think it is simpler to keep it as is without distinguishing between SparseMatrix/DenseMatrix. The code that reads the Matrix can just use the public API of Matrix without knowing whether it is SparseMatrix or DenseMatrix. And the code can use `instance of ` if needed.

I expect this to work well since `mleap` uses the same approach and seems to be OK with it. We can change it if needed in the future.

What do you think?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/types/MatrixType.java
lindong28: Ah my bad. Thanks for catching this. It is updated to the following doc now.

```
/** A matrix of double values. */
```

lindong28:
@zhipeng93 Thanks for the review. Can you help take another look?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/servable/types/MatrixType.java
zhipeng93: Thanks for the explanation. LGTM.

zhipeng93:
Thanks for the update. LGTM.

Fanoid:
> Update `ModelServable#setModelData` to use the signature `default T setModelData(InputStream... modelDataInputs) throws IOException`.

Hi, I just found we have made this change.  I have a question: how do we obtain the corresponding input stream for each `GraphNode` in the `GraphModel`?  `modelDataInputs` are linearly organized, but `GraphNode` are in a form of DAG.

It seems additional parameters are required in the signature of `setModelData` to help map `InputStream`s to `GraphNode`s.

jiangxin369:
> GraphModel

IMO, indeed we need to map `InputStream`s to `GraphNode`s, but it is implementation, not required in API. The existing `GraphModel` class has shown a way to implement it.

Fanoid:
> 



> > It seems additional parameters are required in the signature of setModelData to help map InputStreams to GraphNodes.
> 
> IMO, indeed we need to map `InputStream`s to `GraphNode`s, but it is implementation, not required in API. The existing `GraphModel` class has shown a way to implement it.

If not specifying the mapping in this method,  we may need another public API to obtain the orders of `GraphNode`s. Otherwise, end-users can't figure out how to put the model data of all `GraphNode`s into an array/list.

lindong28:
> Hi, I just found we have made this change. I have a question: how do we obtain the corresponding input stream for each GraphNode in the GraphModel? modelDataInputs are linearly organized, but GraphNode are in a form of DAG.

> It seems additional parameters are required in the signature of setModelData to help map InputStreams to GraphNodes.


@Fanoid we will need to add extra APIs in GraphBuilder and extra class GraphModelServable in order to compose multiple Estimator/Transformer into one Servable. We can obtain the input stream for `ModelServable#setModelData` similar to how the existing `GraphBuilder` obtain input streams for `Model#setModelData`. The high level idea is to use placeholder to provide this information to GraphBuilder. Maybe checkout `testGraphWithSetGetModelData()` [1] for example.

[1] https://github.com/apache/flink-ml/blob/master/flink-ml-core/src/test/java/org/apache/flink/ml/api/GraphTest.java#L213

Fanoid:
@lindong28  Thanks for your explanation.  I missed the point that we already have the orders in the `GraphBuilder#buildEstimator`.  Then, I'm OK with this change.

lindong28:
Thanks @zhipeng93 @Fanoid @jiangxin369 for the review.

PR #215, opened 2023-02-22
[FLINK-31160] Support join/cogroup in BroadcastUtils.withBroadcastStream

zhipeng93:
Currently BroadcastUtils::withBroadcastStream does not support DataStream::cogroup/join, since we restricted that users can only introduce one extra operator in the specified lambda function.

To support using join/cogroup in BroadcastUtils::withBroadcastStream, we would like to relax the restriction such that users can introduce more than more extra operator, but only the output operator can access the broadcast variables.

## Brief change log
- Relaxes the restrictions of BroadcastUtils::withBroadcastStream such that users can introduce more than one operator in the specified lambda function.
- Sets keyContext to deal with the case that the non-broadcast input is a keyed stream.
- Adds unit test to verify the change.

## Documentation


zhipeng93:
@gaoyunhaii @lindong28 Can you review this PR?

zhipeng93:

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -269,138 +200,162 @@ private OperatorMetricGroup createOperatorMetricGroup(
...
+            ThrowingConsumer<Watermark, Exception> watermarkConsumer,
+            ThrowingConsumer<StreamRecord, Exception> keyContextSetter)
             throws Exception {
-        if (!isBlocked[inputIndex]) {
-            if (areBroadcastVariablesReady()) {
-                if (hasPendingElements[inputIndex]) {
-                    processPendingElementsAndWatermarks(
-                            inputIndex, elementConsumer, watermarkConsumer);
-                    hasPendingElements[inputIndex] = false;
+        if (richContext.hasRichFunction) {
```
lindong28, line 223: It is not very readable to have 4 stacked `if` statements.

How about using something like this:

```
if (!richContext.hasRichFunction) {
    elementConsumer.accept(streamRecord);
} else if (richContext.isBlocked[inputIndex]) {
    while (!areBroadcastVariablesReady()) {
        richContext.mailboxExecutor.yield();
    }
    elementConsumer.accept(streamRecord);
} else if (!areBroadcastVariablesReady()) {
    richContext.dataCacheWriters[inputIndex].addRecord(
            CacheElement.newRecord(streamRecord.getValue()));
} else {
    if (richContext.hasPendingElements[inputIndex]) {
        processPendingElementsAndWatermarks(
                inputIndex, elementConsumer, watermarkConsumer, keyContextSetter);
        richContext.hasPendingElements[inputIndex] = false;
        keyContextSetter.accept(streamRecord);
    }
    elementConsumer.accept(streamRecord);
}
```

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastUtils.java
```diff
@@ -152,41 +155,52 @@ private static <OUT> DataStream<OUT> cacheBroadcastVariables(
...
         }
         DataStream<OUT> draftOutStream = graphBuilder.apply(draftSources);
-        Preconditions.checkState(
-                draftEnv.getStreamGraph(false).getStreamNodes().size() == 1 + inputList.size(),
-                "cannot add more than one operator in withBroadcastStream's lambda function.");
-        draftEnv.copyToActualEnvironment();
-        return draftEnv.getActualStream(draftOutStream.getId());
+
+        List<Transformation<?>> realNonBroadcastInputs =
+                draftOutStream.getTransformation().getInputs();
```
lindong28, line 185: Can you explain why the inputs of the transformation of datastream generated by the `graphBuilder` is equivalent to the non-broadcast element in the original `inputList`?

Note that `graphBuilder` can include multiple join/map operators such that the inputs of the datastream computed by it might not even be a subset of the original `inputList`.

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -468,7 +425,9 @@ public void initializeState(StreamTaskStateInitializer streamTaskStateManager)
 
         timeServiceManager = streamOperatorStateContext.internalTimerServiceManager();
 
-        broadcastVariablesReady = false;
+        if (richContext.hasRichFunction) {
+            richContext.broadcastVariablesReady = false;
```
lindong28, line 429: Would it be more readable to explicitly initialize `broadcastVariablesReady` when `hasRichFunction == false`?

lindong28 (commented):

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastUtils.java
zhipeng93: The `realNonBroadcastInputs` is not equivalent to the `inputList` here. 

When processing the non-broadcast streams, we need to block until all the broadacast streams are consumed. However, the `inputList` are not always the real non-broadcast streams.

For example, when calling `withBroadcastStreams()` with join as a udf (see [1]), the inputList are `source1` and `source2`, but the real non-broadcast input is a keyedStream generated following [2].

[1] https://github.com/apache/flink-ml/pull/215/files#diff-92d87f6ba02ee90f721294e1d720d7a56b6b911ee8b5c2e00e0f0ce39fb21359R134
[2] https://github.com/apache/flink/blob/master/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/datastream/CoGroupedStreams.java#L388

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
zhipeng93: Logically if `hasRichFunction == false`, we will never access the value of `broadcastVariablesReady`. So it seems unnecessary to set the value here.

But thanks for the catch, I found it unnecessary to call this function here since it is always initialized in `RichContext`. 

I will remove these three lines later.

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -269,138 +200,162 @@ private OperatorMetricGroup createOperatorMetricGroup(
...
-                            inputIndex, elementConsumer, watermarkConsumer);
-                    hasPendingElements[inputIndex] = false;
+        if (richContext.hasRichFunction) {
+            if (!richContext.isBlocked[inputIndex]) {
+                if (areBroadcastVariablesReady()) {
+                    if (richContext.hasPendingElements[inputIndex]) {
+                        processPendingElementsAndWatermarks(
+                                inputIndex, elementConsumer, watermarkConsumer, keyContextSetter);
+                        richContext.hasPendingElements[inputIndex] = false;
+                        keyContextSetter.accept(streamRecord);
```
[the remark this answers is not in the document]
zhipeng93: I am afraid not. `keyContextSetter.accept(streamRecord);` tries to set the key for the received streamRecord, not the cached ones.

We also set key context for pending elements in `processPendingElementsAndWatermarks`. [1]

[1] https://github.com/apache/flink-ml/pull/215/files/a4e1f0576384762b22e9bbc264a020a6ec7bda39#diff-001f424b1d7ebde26cc633756f68c9c83d71b33fce0e9058d5b22ba2746e2959R357

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
zhipeng93: The stacked `if` statements here represent the real execution logic here. For example in the code snippet you provided here, the `if` statement means that this operator does not need to access broadcast varaibles, while the three `else` statement are different conditions when this operator needs to access broadcast variables.

So the original code snippet seems more readable to me, what do you think?

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -169,83 +144,39 @@
...
-                OperatorUtils.getDataCachePath(
-                        containingTask.getEnvironment().getTaskManagerInfo().getConfiguration(),
-                        containingTask
-                                .getEnvironment()
-                                .getIOManager()
-                                .getSpillingDirectoriesPaths());
-        dataCacheWriters = new DataCacheWriter[numInputs];
-        hasPendingElements = new boolean[numInputs];
-        Arrays.fill(hasPendingElements, true);
+        richContext = new RichContext(hasRichFunction, broadcastStreamNames, inTypes, isBlocked);
```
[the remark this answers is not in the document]
zhipeng93: `RichContext` stores the variables that this operator needs to access the broadcast variables. If this operator does not needs to access broadcast variables (i.e., has no rich functions), then all the variables in `RichContext` (except `HasRichFunction`) should be null.

I can extract it as a separate commit when we get aligned on this PR.

zhipeng93:
@lindong28 Thanks for the review. I have addressed your comments in the latest PR.

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastUtils.java
```diff
@@ -152,41 +155,52 @@ private static <OUT> DataStream<OUT> cacheBroadcastVariables(
...
+            inTypes[i] = realNonBroadcastInputs.get(i).getOutputType();
+        }
+        // Does not block all non-broadcast input edges by default.
+        boolean[] isBlocked = new boolean[realNonBroadcastInputs.size()];
+        Arrays.fill(isBlocked, false);
+
+        // Executes the graph builder and adds the operators to real execution environment.
+        DraftExecutionEnvironment draftEnv2 =
+                new DraftExecutionEnvironment(
+                        env, new BroadcastWrapper<>(broadcastStreamNames, inTypes, isBlocked));
```
lindong28, line 197: Is it possible for `AbstractBroadcastWrapperOperator` to retrieve inTypes from `wrappedOperator` directly instead of getting this information from the constructor? 

If we can do this, we can make the code more readable by only invoking `graphBuilder` only once here.

Maybe add a TODO here if it is not straightforward to make this change in this PR.

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -169,13 +182,16 @@
...
+        this.numInputs = inTypes.length;
+
+        this.hasRichFunction =
                 wrappedOperator instanceof AbstractUdfStreamOperator
                         && ((AbstractUdfStreamOperator) wrappedOperator).getUserFunction()
                                 instanceof RichFunction;
 
         if (hasRichFunction) {
-            wrappedOperatorRuntimeContext =
+            this.wrappedOperatorRuntimeContext =
```
lindong28, line 194: Is it possible to raise exception if `wrappedOperatorRuntimeContext#getBroadcastVariable` or `wrappedOperatorRuntimeContext#getBroadcastVariableWithInitializer` is invoked on an operator other than the output operator of the `graphBuilder`?

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -269,125 +279,136 @@ private OperatorMetricGroup createOperatorMetricGroup(
...
-                elementConsumer.accept(streamRecord);
-
-            } else {
-                dataCacheWriters[inputIndex].addRecord(
-                        CacheElement.newRecord(streamRecord.getValue()));
-            }
-
-        } else {
+        if (!hasRichFunction) {
+            elementConsumer.accept(streamRecord);
```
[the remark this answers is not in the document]
zhipeng93: We do not need to set the key context for records that are not cached. It is set by upstream functions, e.g., [1].

[1] https://github.com/apache/flink/blob/release-1.15/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/ChainingOutput.java#L98

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -269,125 +279,136 @@ private OperatorMetricGroup createOperatorMetricGroup(
...
+            ThrowingConsumer<Watermark, Exception> watermarkConsumer,
+            ThrowingConsumer<StreamRecord, Exception> keyContextSetter)
             throws Exception {
-        while (!areBroadcastVariablesReady()) {
-            mailboxExecutor.yield();
-        }
-        if (hasPendingElements[inputIndex]) {
-            processPendingElementsAndWatermarks(inputIndex, elementConsumer, watermarkConsumer);
-            hasPendingElements[inputIndex] = false;
+        if (hasRichFunction) {
```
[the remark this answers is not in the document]
zhipeng93: Thanks for the comment. I update updated all of the `if (hasRichFunction)` except the one in the constructor, which I think it provides better readability.

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
zhipeng93: It does throws NPE when it is not the output operator of `graphBuilder`. [1]

[1] https://github.com/apache/flink-ml/blob/master/flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastStreamingRuntimeContext.java#L87

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastUtils.java
zhipeng93: Good catch. I have updated the implementation as you suggested. Please checkout the latest commit.

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/operator/AbstractBroadcastWrapperOperator.java
```diff
@@ -269,138 +295,151 @@ private OperatorMetricGroup createOperatorMetricGroup(
...
+        } else if (isBlocked[inputIndex]) {
             while (!areBroadcastVariablesReady()) {
                 mailboxExecutor.yield();
             }
             elementConsumer.accept(streamRecord);
+        } else if (!areBroadcastVariablesReady()) {
+            dataCacheWriters[inputIndex].addRecord(CacheElement.newRecord(streamRecord.getValue()));
+        } else {
+            if (hasPendingElements[inputIndex]) {
+                processPendingElementsAndWatermarks(
```
lindong28, line 329: Would it be useful to clear the cached records after they are processed?

flink-ml-core/src/main/java/org/apache/flink/ml/common/broadcast/BroadcastUtils.java
```diff
@@ -57,8 +56,9 @@ public class BroadcastUtils {
      * @param bcStreams map of the broadcast data streams, where the key is the name and the value
      *     is the corresponding data stream.
      * @param userDefinedFunction the user defined logic in which users can access the broadcast
-     *     data streams and produce the output data stream. Note that users can add only one
-     *     operator in this function, otherwise it raises an exception.
+     *     data streams and produce the output data stream. Note that though users can add more than
+     *     one operator in this logic, but only the operator that generates the result stream can
+     *     contain a rich function and access the broadcast variables.
```
lindong28, line 61: Would it be useful to include the following:

Other operators will encounter NPE when accessing the broadcast variables.

lindong28 (commented):
Thanks for the update. LGTM overall. Left two minor comments.

zhipeng93:
Thanks for the review. The PR is updated :)

lindong28:
Thanks for the update! LGTM.

PR #216, opened 2023-02-22
[FLINK-31173] Fix serveral iteration bugs

zhipeng93:
This PR fixes several iteration bugs.

## Brief change log
  - Restrict that TailOperator should only have one input since it does not align the epoch watermarks.
  - Fix wrong typeinfo in ProxyOperatorStateBackend.
  - Added unit test to verify the above changes.

## Documentation

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/proxy/state/ProxyOperatorStateBackend.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/proxy/state/ProxyOperatorStateBackend.java
@@ -21,6 +21,11 @@ import org.apache.flink.api.common.state.BroadcastState;
 import org.apache.flink.api.common.state.ListState;
 import org.apache.flink.api.common.state.ListStateDescriptor;
 import org.apache.flink.api.common.state.MapStateDescriptor;
+import org.apache.flink.api.common.state.StateDescriptor;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.ListTypeInfo;
+import org.apache.flink.api.java.typeutils.MapTypeInfo;
+import org.apache.flink.iteration.utils.ReflectionUtils;
 import org.apache.flink.runtime.checkpoint.CheckpointOptions;
 import org.apache.flink.runtime.state.CheckpointStreamFactory;
 import org.apache.flink.runtime.state.OperatorStateBackend;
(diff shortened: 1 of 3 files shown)


zhipeng93:

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedIterationWithBroadcastITCase.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @After
+    public void teardown() throws Exception {
+        if (miniCluster != null) {
+            miniCluster.close();
+        }
+    }
+
+    @Test
+    public void testIterationWithBroadcastJobGraph() throws Exception {
```
lindong28, line 81: Can you help explain why `ListStateDescriptor` is not initialized in this test but initialized in other tests?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/proxy/state/ProxyOperatorStateBackend.java
```diff
@@ -58,12 +62,24 @@ public <K, V> BroadcastState<K, V> getBroadcastState(MapStateDescriptor<K, V> st
...
+        return ((ListTypeInfo) f.get(stateDescriptor)).getElementTypeInfo();
+    }
+
     @Override
     public <S> ListState<S> getListState(ListStateDescriptor<S> stateDescriptor) throws Exception {
         ListStateDescriptor<S> newDescriptor =
-                new ListStateDescriptor<>(
-                        stateNamePrefix.prefix(stateDescriptor.getName()),
-                        stateDescriptor.getElementSerializer());
+                stateDescriptor.isSerializerInitialized()
```
lindong28, line 75: Does `getUnionListState()` also need to check this?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/proxy/state/ProxyOperatorStateBackend.java
```diff
@@ -58,12 +62,24 @@ public <K, V> BroadcastState<K, V> getBroadcastState(MapStateDescriptor<K, V> st
         return wrappedBackend.getBroadcastState(newDescriptor);
     }
 
+    private <S> TypeInformation<S> getTypeInfo(ListStateDescriptor<S> stateDescriptor)
+            throws Exception {
+        Field f = StateDescriptor.class.getDeclaredField("typeInfo");
```
lindong28, line 67: It is probably more readable to use `ReflectionUtils` in a way similar to `OutputReflectionContext`.

lindong28 (commented):

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedTailOperatorWithUnionInputIterationITCase.java
```diff
@@ -0,0 +1,75 @@
...
+        @Override
+        public IterationBodyResult process(
+                DataStreamList variableStreams, DataStreamList dataStreams) {
+
+            DataStream<Long> variableStream = variableStreams.get(0);
+            DataStream<Long> constantStream = dataStreams.get(0);
+            DataStream<Integer> terminationCriteria =
+                    constantStream.flatMap(new TerminateOnMaxIter<>(3));
+            return new IterationBodyResult(
+                    DataStreamList.of(variableStream.union(constantStream)),
```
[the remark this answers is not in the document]
zhipeng93: Currently `TailOperator` does not align the epoch watermarks from its inputs, thus will cause the iteration incorrect.

To support using `opA union opB` as the feedback, users can now use `(opA union opB).map(x -> x)` for a workaround.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/proxy/state/ProxyOperatorStateBackend.java
zhipeng93: Good catch. I have updated the check for `UnionListState` and `BroadcastState`.

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedIterationWithBroadcastITCase.java
zhipeng93: The `getListState` in `DefaultOperatorStateBackend` does the initialization here [1] while `ProxyOperatorStateBackend` does not.

lindong28: Suppose that `ProxyOperatorStateBackend#wrappedBackend` is a `DefaultOperatorStateBackend`. Then `ProxyOperatorStateBackend#getUnionListState` invokes `DefaultOperatorStateBackend#getUnionListState`, which should in turn invoke `DefaultOperatorStateBackend#getListState` and initialize the state descriptor, right?

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedTailOperatorWithUnionInputIterationITCase.java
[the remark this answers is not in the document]
lindong28: Given that we have this work around, is it possible for `Iterations#addTails` to add the `map(x -> x)` instead of throwing exception?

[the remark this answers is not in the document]
zhipeng93: I am afraid adding `map(x -> x)` to the wrapped operator does not work here. Currently iterations do not support `watermark` alignment. The bug comes in because `TailOperator` does not align the `epochwatermark`, rather than watermarks. `epochwatermark` should be viewed as a normal record from Flink framework's view.

The current situation is that:
- The tailOperator sends the epochwatermark to headOperator.
- headOperator expects that the epochwatermark should be strictly increasing.
- The tail operator made an assumption that it has only one input and it does not align the epochwatermarks. When receiving epochwatermarks, it directly pushes to the head operator.

If the tail operator has two inputs, it will push two same epochwatermarks to headOperator, which breaks the assumption.

I also added a TODO there.

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/BoundedIterationWithBroadcastITCase.java
zhipeng93: Before calling `DefaultOperatorStateBackend#getListState`, we need to call `stateDescriptor.getElementSerializer()`. If it is null, it throws NPE here.

zhipeng93: Before calling `DefaultOperatorStateBackend#getListState`, `ProxyOperatorStateBackend` will call `getElementSerializer` first, this may lead to NPE.

lindong28:
Thanks for the update. LGTM.

PR #219, opened 2023-03-03
[FLINK-31422] Add Servable for Logistic Regression Model

jiangxin369:

Add Servable for Logistic Regression.

## Brief change log

  - Adds Servable for Logistic Regression
  - Move common params from flink-ml-lib to flink-ml-servable

## Documentation


flink-ml-servable-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServable.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * The main logic that predicts one input data point.
+     *
+     * @param feature The input feature.
+     * @param coefficient The model parameters.
+     * @return The prediction label and the raw probabilities.
+     */
+    public static Tuple2<Double, DenseVector> predictOneDataPoint(
```
lindong28, line 114: Should we make this method protected so that it won't be used by code outside Flink ML repo to do online prediction?

Would it be better to name this method `Tuple2<Double, DenseVector> transform(Vector feature)` so that it is more consistent with `DataFrame transform(DataFrame input)`?

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModel.java
```diff
@@ -150,21 +153,9 @@ public Row map(Row dataPoint) {
                 coefficient = modelData.coefficient;
             }
             DenseVector features = ((Vector) dataPoint.getField(featuresCol)).toDense();
-            Row predictionResult = predictOneDataPoint(features, coefficient);
-            return Row.join(dataPoint, predictionResult);
+            Tuple2<Double, DenseVector> predictionResult =
+                    LogisticRegressionModelServable.predictOneDataPoint(features, coefficient);
```
lindong28, line 157: Would it be more readable to instantiate an instance of `LogisticRegressionModelServable` and call `LogisticRegressionModelServable#transform(features)` here?

flink-ml-servable-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServable.java
```diff
@@ -0,0 +1,148 @@
...
+
+        return outputStream.toByteArray();
+    }
+
+    @Override
+    public Map<Param<?>, Object> getParamMap() {
+        return paramMap;
+    }
+
+    public DenseVector getCoefficient() {
```
lindong28, line 145: Is it necessary to add this public API for users to get servable model data? This public API was not discussed in any existing FLIP.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/LogisticRegressionTest.java
```diff
@@ -253,6 +255,25 @@ public void testSaveLoadAndPredict() throws Exception {
...
+                        tEnv,
+                        model,
+                        tempFolder.newFolder().getAbsolutePath(),
+                        LogisticRegressionModel::loadServable);
+
+        assertEquals("features", servable.getFeaturesCol());
+        assertEquals("prediction", servable.getPredictionCol());
+        assertEquals("rawPrediction", servable.getRawPredictionCol());
+
+        assertArrayEquals(expectedCoefficient, servable.getCoefficient().values, 0.1);
```
lindong28, line 274: Should we verify the result of the `servable.transform(...)` similar to `PipelineTest#testPipelineModelServable`?

flink-ml-servable-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServable.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Serializes the model data into byte array which can be saved to external storage and then be
+     * used to update the Servable by `TransformerServable::setModelData` method.
+     *
+     * @param modelData The model data to be serialized.
+     * @return The serialized model data in byte array.
+     */
+    public static byte[] serialize(DenseVector modelData) throws IOException {
```
lindong28, line 128: Does this mean that users need to know the internal data structure/type of the model data in order to use this public method?

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServableTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+        verifyPredictionResult(
+                output,
+                servable.getFeaturesCol(),
+                servable.getPredictionCol(),
+                servable.getRawPredictionCol());
+    }
+
+    @Test
+    public void testSetModelData() throws IOException {
```
lindong28, line 118: This test looks **exactly** the same as `testTransform()`. What is the purpose of having both tests?

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServableTest.java
```diff
@@ -0,0 +1,153 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/** Tests the {@link LogisticRegressionModelServable}. */
+public class LogisticRegressionModelServableTest {
```
lindong28, line 40: Should we have a test that covers the processing of saving model and loading servable, similar to `PipelineTest#testPipelineModelServable`?

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServableTest.java
jiangxin369: Yes, I should have deleted it.

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServableTest.java
jiangxin369: It is covered in `LogisticRegressionTest#testSaveLoadServable`.

flink-ml-lib/src/test/java/org/apache/flink/ml/classification/LogisticRegressionTest.java
jiangxin369: I don't think we really need to verify the result of the `servable.transform(...)`, because it has been covered in `LogisticRegressionModelServableTest#testTransform`. If you are worrying about the consistency of test cases of `PipelineTest` and `LogisticRegressionModelServableTest`, I think we should change the `PipelineTest#testPipelineModelServable` to only verify the model data of each stages.

lindong28: `LogisticRegressionModelServableTest#testTransform` does not save/load servable.

We need to have a test that shows that we can save/load servable and use the loaded servable for online prediction, similar to why we have both `LogisticRegressionTest#testFitAndPredict` and `LogisticRegressionTest#testSaveLoadAndPredict`.

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelServableTest.java
lindong28: `LogisticRegressionTest#testSaveLoadServable` currently does not verify the result of `LogisticRegressionModelServable#transform`.

jiangxin369:
@lindong28 The PR is updated, could you have another look?

lindong28 (commented):

flink-ml-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModel.java
```diff
@@ -149,22 +152,14 @@ public Row map(Row dataPoint) {
...
             }
-            DenseVector features = ((Vector) dataPoint.getField(featuresCol)).toDense();
-            Row predictionResult = predictOneDataPoint(features, coefficient);
-            return Row.join(dataPoint, predictionResult);
-        }
-    }
+            Vector features = (Vector) dataPoint.getField(featuresCol);
+
+            LogisticRegressionModelServable servable =
+                    new LogisticRegressionModelServable(
```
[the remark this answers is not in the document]
jiangxin369: 1. Without the `LogisticRegressionModelServable(LogisticRegressionModelData)` constructor, every time we want to construct the `LogisticRegressionModelServable` with an existing `LogisticRegressionModelData`, we need to serialize the instance of `LogisticRegressionModelData` to bytes and then deserialize the bytes via `setModelData(InputStream... modelDataInputs)`, it may be expensive if the model data is large. I prefer to keep the constructor but make it `protected` or `default`.
2.   Yes we should forward parameters from model to servable, but the `LogisticRegressionModelServable` doesn't need any parameters from model to do predictions. If we are supporting `MinMaxScalerModelServable`, developers can pass the params by `servable.setMin()` directly.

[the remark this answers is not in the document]
lindong28: Making `LogisticRegressionModelServable(LogisticRegressionModelData) ` protected sounds good to me.

However, asking developers to identify and forward the selectively parameters from model to servable is not very convenient and introduces chances of human error. It will much more convenient for Flink ML to provide infra that algorithm developers can use to forward parameters, similar to how the existing Estimators forward parameters to Model using `ParamUtils.updateExistingParams`.

jiangxin369:
@lindong28 Thanks for the review, could you help review again?

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/LogisticRegressionModelServableTest.java
```diff
@@ -0,0 +1,142 @@
...
+                .setPredictionCol("test_predictionCol")
+                .setRawPredictionCol("test_rawPredictionCol");
+        assertEquals("test_features", servable.getFeaturesCol());
+        assertEquals("test_predictionCol", servable.getPredictionCol());
+        assertEquals("test_rawPredictionCol", servable.getRawPredictionCol());
+    }
+
+    @Test
+    public void testTransform() throws IOException {
+        LogisticRegressionModelData modelData = new LogisticRegressionModelData(COEFFICIENT);
```
lindong28, line 106: I am not sure we need this test given that we don't expect users to construct model data directly like this.

Note that we do not have a similar test for LogisticRegressionModel where we explicitly construct model data and give it to the LogisticRegressionModel.

lindong28 (commented):

flink-ml-servable-lib/src/test/java/org/apache/flink/ml/classification/LogisticRegressionModelServableTest.java
jiangxin369: I think it is expected to construct model data like this, e.g. for users who want to test if the servable transformation logic is correct without training a model. 

We do not have a similar test for LogisticRegressionModel because the `LogisticRegression` and `LogisticRegressionModel` are in the same module, it is reasonable to validate `fit` and `transform` in the same test case. However, `LogisticRegressionModelServable` is in a separate module, which means it can be used and tested independently.

lindong28: The separate of `flink-ml-lib` and `flink-ml-servable-lib` is an implementation detail and should not affect how users use `Servable` classes, right?

jiangxin369: I agree. In this way, I could remove all tests in `flink-ml-servable-lib` currently. 

I've updated the PR based on your advice, could you have a look again?

flink-ml-servable-lib/src/main/java/org/apache/flink/ml/classification/logisticregression/LogisticRegressionModelData.java
```diff
@@ -0,0 +1,80 @@
...
+/** Model data of {@link LogisticRegressionModelServable}. */
+public class LogisticRegressionModelData {
+
+    public DenseVector coefficient;
+
+    public long modelVersion;
+
+    public LogisticRegressionModelData() {}
+
+    public LogisticRegressionModelData(DenseVector coefficient) {
```
lindong28, line 40: Would it be simpler to remove this constructor and let users instantiate `LogisticRegressionModelData(.., 0)` similar to what is done in `LogisticRegression#fit`?

lindong28 (commented):
Thanks for the update. LGTM overall. Left one minor comment.

PR #222, opened 2023-03-09
[FLINK-31029] Fix bug when using quantile in KbinsDiscretizer

zhipeng93:

This PR aims to fix the bug when using quantile strategy in KbinsDiscretizer. 

- In the original version, we remove bins with zero width by simply merging it with the next bin. However it is not always true. For example, if the histograms computed are [0, 0, 0, 1], then it is transformed into [0, 1]. Thus is 0 and 1 are mapped in to the same bin, which is wrong.
- In this PR, the histogram will be transformed into [0, 0.5, 1]. It is computed with the following logic:
    - Remove the repeated elements in the histgram such that there are no 3+ consecutive same elements.
    - If there two consecutive same elements, the second one is updated as the average of the element before and after.
    - If the last two elements in the histogram are the same, the last one is removed.

## Brief change log
  - Changed the computing logic of handling zero-width bins.
  - Added unit test to verify it.

## Documentation

### DIFF
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
@@ -67,7 +67,10 @@ import java.util.Set;
  *       and the {@link KBinsDiscretizerParams#STRATEGY} is set as {@link
  *       KBinsDiscretizerParams#KMEANS}, we switch to {@link KBinsDiscretizerParams#UNIFORM}.
  *   <li>When the width of one output bin is zero, i.e., the left edge equals to the right edge of
- *       the bin, we remove it.
+ *       the bin, we replace the right edge as the average value of its two neighbors. One exception
+ *       is that the last two edges are the same --- in this case, the left edge is updated as the
+ *       average of its two neighbors. For example, the bin edges {0, 1, 1, 2, 2} are transformed
+ *       into {0, 1, 1.5, 1.75, 2}.
  * </ul>
  */
 public class KBinsDiscretizer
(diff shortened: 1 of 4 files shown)


zhipeng93:
@Fanoid Can you help to review this PR?

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
```diff
@@ -220,22 +220,46 @@ private static double[][] findBinEdgesWithQuantileStrategy(
...
+                tempBinEdges = edges.stream().mapToDouble(Double::doubleValue).toArray();
+                Arrays.sort(tempBinEdges);
+                int i = 1;
+                for (; i < tempBinEdges.length - 1; i++) {
+                    if (tempBinEdges[i] == tempBinEdges[i - 1]) {
+                        tempBinEdges[i] = (tempBinEdges[i + 1] + tempBinEdges[i - 1]) / 2;
+                    }
+                }
+                if (tempBinEdges[i] == tempBinEdges[i - 1]) {
+                    binEdges[columnId] = Arrays.copyOfRange(tempBinEdges, 0, i);
```
Fanoid, line 259: Does this line means the last bin edge is dropped?  

If so, when the `tempBinEdges ` are `[0, 1, 1]`, `binEdges ` will be `[0, 1]`. Then all values are in same bin in `KBinsDiscretizerModel#transform`.   

It seems this case should be handle separately when there are only two distinct bin edges.

Fanoid (commented):
Hi, thanks for your fix. I've checked the implementation and found a special case. Please take a look at it.

BTW, can we clearly state our strategy for duplicated edges in comments? It will be more convenient to judge the strategy.

flink-ml-lib/src/main/java/org/apache/flink/ml/feature/kbinsdiscretizer/KBinsDiscretizer.java
zhipeng93: Good catch. I have updated the logic and also updated the java doc.

In the updated PR, [0, 1, 1] should be transformed into [0, 0.5, 1]

Fanoid:
Thanks for your updates.  Current version can meet my needs now. LGTM.

lindong28:
Thanks for the PR. LGTM.

PR #227, opened 2023-03-28
[FLINK-31623] Fix DataStreamUtils#sample with approximate uniform sampling

Fanoid:

This modification will resolve the issue with `DataStreamUtils#sample` by implementing uniform sampling.

The current implementation utilizes a two-level sampling method. However, if the elements are unevenly distributed among partitions (subtasks), the sampling probabilities may differ across partitions, resulting in non-uniform sampling.

Furthermore, the current implementation has a side-effect where one subtask has a memory footprint of `2 * numSamples * sizeof(element)`, potentially leading to unexpected OOM errors in certain scenarios.

## Brief change log

  - Adds `rebalance` to the input data, and decrease the number of samples in the first round.
  - Adds unit tests for this method.

  - Dependencies (does it add or upgrade a dependency): (yes / no) no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / no) no

## Documentation

  - Does this pull request introduce a new feature? (yes / no) no
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) not applicable

### DIFF
--- a/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
+++ b/flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
@@ -280,6 +280,15 @@ public class DataStreamUtils {
      * <p>This method takes samples without replacement. If the number of elements in the stream is
      * smaller than expected number of samples, all elements will be included in the sample.
      *
+     * <p>Technical details about this method: Firstly, the input elements are rebalanced. Then, in
+     * the first-round sampling, `firstRoundNumSamples` samples are selected from each data
+     * partitions. In the second-round sampling, `numSamples` elements are sampled from the results
+     * of the first round.
+     *
+     * <p>If the input elements are evenly distributed after rebalancing each element is selected
+     * with an equal probability during both rounds of sampling. Otherwise, elements are picks based
+     * on probabilities with tolerable differences.
+     *
      * @param input The input data stream.
      * @param numSamples The number of elements to be sampled.
      * @param randomSeed The seed to randomly pick elements as sample.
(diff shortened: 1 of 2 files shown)


flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -280,6 +280,15 @@ public static <IN, ACC, OUT> DataStream<OUT> aggregate(
      * <p>This method takes samples without replacement. If the number of elements in the stream is
      * smaller than expected number of samples, all elements will be included in the sample.
      *
+     * <p>Technical details about this method: Firstly, the input elements are rebalanced. Then, in
```
zhipeng93, line 283: nit: How about we update the java doc as follows and move the technique details to the implementation part or remove it?

`Performs an approximate uniform sampling over the elements in a bounded data stream. The difference of probablity of two data points been sampled is bounded by O(numSamples * p * p / (M * M)), where p is the paralellism of the input stream, M is the total number of data points that the input stream contains.`

This method takes samples without replacement. If the number of elements in the stream is smaller than expected number of samples, all elements will be included in the sample.

Fanoid: Thanks for your comments. I've updated the PR.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -288,13 +291,20 @@ public static <IN, ACC, OUT> DataStream<OUT> aggregate(
     public static <T> DataStream<T> sample(DataStream<T> input, int numSamples, long randomSeed) {
         int inputParallelism = input.getParallelism();
 
-        return input.transform(
+        // In a worst-case scenario, the data partition with the greatest number of elements has
+        // `inputParallelism` additional elements compared to the one with the fewest elements even
+        // after `rebalance` performed. Therefore, additional elements are sampled from each
+        // partition in case some partitions has insufficient elements.
+        int firstRoundNumSamples =
```
zhipeng93, line 298: nit: The max difference of number of data points in each partition after calling `rebalance` is `inputParallism`. 
As a result, extra `inputParallelism` data points are sampled for each partition in the first round.

flink-ml-core/src/main/java/org/apache/flink/ml/common/datastream/DataStreamUtils.java
```diff
@@ -288,13 +291,20 @@ public static <IN, ACC, OUT> DataStream<OUT> aggregate(
...
+        return input.rebalance()
+                .transform(
                         "samplingOperator",
                         input.getType(),
-                        new SamplingOperator<>(numSamples, randomSeed))
+                        new SamplingOperator<>(firstRoundNumSamples, randomSeed))
                 .setParallelism(inputParallelism)
                 .transform(
-                        "samplingOperator",
+                        "samplingOperator-2nd-round",
```
zhipeng93, line 307: nit: Let's make the samping operator name consistent.

zhipeng93 (commented):
LGTM overall. Left two minor comments here.

Fanoid:
Hi, I've updated the codes according to your comments. Please check it.

PR #229, opened 2023-04-04
[FLINK-31255] OperatorUtils#createWrappedOperatorConfig should update input and sideOutput serializers

jiangxin369:

Wraps the operator config about serializer.

## Brief change log

*(for example:)*
  - Wraps the operator config about serializer.

## Documentation


jiangxin369:
@zhipeng93  Could you help review this PR? And I have no idea why the workflow needs additional approval to run in this PR, could you help approve the workflow running?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
```diff
@@ -104,6 +107,33 @@ public static StreamConfig createWrappedOperatorConfig(
             }
         }
 
+        StreamConfig.InputConfig[] inputs = wrapperConfig.getInputs(cl);
+        for (int i = 0; i < inputs.length; ++i) {
+            if (inputs[i] instanceof NetworkInputConfig) {
+                TypeSerializer<?> typeSerializerIn =
+                        ((NetworkInputConfig) inputs[i]).getTypeSerializer();
+                if (typeSerializerIn instanceof IterationRecordSerializer) {
```
zhipeng93, line 115: It seems that `typeSerializerIn` and `typeSerializerOut` should always be `IterationRecordSerializer` for wrapped iteration operators if they are not null.

Is there any case that this condition does not hold?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
```diff
@@ -104,6 +107,33 @@ public static StreamConfig createWrappedOperatorConfig(
...
+
+        TypeSerializer<?> typeSerializerOut = wrapperConfig.getTypeSerializerOut(cl);
+        if (typeSerializerOut instanceof IterationRecordSerializer) {
+            wrappedConfig.setTypeSerializerOut(
+                    ((IterationRecordSerializer<?>) typeSerializerOut).getInnerSerializer());
+        }
+
+        TypeSerializer<?> stateKeySerializer = wrapperConfig.getStateKeySerializer(cl);
+        if (stateKeySerializer instanceof IterationRecordSerializer) {
+            wrappedConfig.setTypeSerializerOut(
```
zhipeng93, line 133: It should be `wrappedConfig.setStateKeySerializer(...)` here.

zhipeng93 (commented):
Thanks for the fix. Left two comments below. Please take a look.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
jiangxin369: Yes, you're right. I removed the  `if ... instanceOf ...` check. And seems that we did not modify the `StateKeySerializer` at all, so I also removed this part.

jiangxin369:
@zhipeng93 Thanks for the review, I've updated the PR.

zhipeng93:
The update LGTM except the unresolved one. @lindong28 Would you like to take  a look?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
jiangxin369: Yes, the `typeSerializerIn` and `typeSerializerOut` should be `IterationRecordSerializer` in the current usage. But since we are providing a public utility method, I think we'd better check if the `serializer`s are not null and an instance of `IterationRecordSerializer`.

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/OperatorUtilsTest.java
```diff
@@ -0,0 +1,141 @@
...
+                    SingleOutputStreamOperator transformed =
+                            variable.transform(
+                                    "side-output",
+                                    Types.INT,
+                                    new MockSideOutputOperator<>(outputTag));
+                    return new IterationBodyResult(
+                            DataStreamList.of(variable),
+                            DataStreamList.of(transformed.getSideOutput(outputTag)));
+                });
+        env.execute();
```
lindong28, line 96: Can we additionally check the value of the output?

Since the test actually starts a job in mini-cluster, would it be simpler to add the test to an test class in flink-ml-tests?

lindong28 (commented):

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/OperatorUtilsTest.java
jiangxin369: I added the check of the value of the output.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
```diff
@@ -104,6 +110,50 @@ public static StreamConfig createWrappedOperatorConfig(
             }
         }
 
+        StreamConfig.InputConfig[] inputs = wrapperConfig.getInputs(cl);
```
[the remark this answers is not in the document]
jiangxin369: It's a good idea to rename `wrapperConfig` to `config`, but I don't think it matters to use `wrapperConfig` or `wrappedConfig`. Because the `wrappedConfig` is just a clone from `wrapperConfig` and we are getting the new configurations from the `wrapperConfig` and set to the `wrappedConfig`, just like the above code,
```
KeySelector keySelector = config.getStatePartitioner(i, cl);
            if (keySelector != null) {
                checkState(
                        keySelector instanceof ProxyKeySelector,
                        "The state partitioner for the wrapper operator should always be ProxyKeySelector, but it is "
                                + keySelector);
                wrappedConfig.setStatePartitioner(
                        i, ((ProxyKeySelector) keySelector).getWrappedKeySelector());
            }
```

jiangxin369:
@lindong28 Thanks for the review. I updated the PR, could you have another look?

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/OperatorUtilsTest.java
lindong28: `BroadcastOutputTest.java` uses `BroadcastOutput` directly and `OperatorEpochWatermarkTrackerFactoryTest.java` uses `OperatorEpochWatermarkTracker` directly. But `OperatorUtilsTest` does not use `OperatorUtilsTest` directly. Thus it seems better to move this test to e.g. `UnboundedStreamIterationITCase`.

flink-ml-tests/src/test/java/org/apache/flink/test/iteration/UnboundedStreamIterationITCase.java
```diff
@@ -152,6 +160,39 @@ public void testVariableAndConstantBoundedIteration() throws Exception {
         assertEquals(OutputRecord.Event.TERMINATED, result.get().take().getEvent());
     }
 
+    @Test
+    public void testUnwrapOperatorConfig() throws Exception {
```
lindong28, line 164: Can we name this test based on the key property of the API used by this test?

For example, if the key property here is the use of SideOutput, maybe name it as `testBoundedIterationWithSideoutput`.

lindong28 (commented):
Thanks for the update. Left one minor comment.

lindong28:
Thanks for the update. LGTM.

PR #234, opened 2023-04-23
[FLINK-31887] Upgrade Flink version of Flink ML to 1.16.1

jiangxin369:

Upgrade Flink version of Flink ML to 1.16.1

## Brief change log

Upgrade Flink version of Flink ML to 1.16.1. The main changes made for compatibility are as below.
  - 1.16 adds a field `createdKVStates` in KeyedStateBackend to store states, which should be cleared in PerRound Iteration. Otherwise the tests like `PerRoundOperatorStateTest#testStateIsolationWithHashMapKeyedStateBackend` would fail.
  - 1.16 removes the `getNonChainedOutputs` API from `StreamConfig` which is used during creating wrapped operator configs.
  - `KeyedStateFactory#createInternalState` interface changed to `KeyedStateFactory#createOrUpdateInternalState`.
  - 1.16 modifies some interfaces of `OperatorCoordinator`, we have to re-implement `HeadOperatorCoordinator`.
  - The `setXXX` functions in `StreamConfig` are changed to lazy, we need to trigger the serialization eagerly during creating wrapped operator configs.
  - 1.16 doesn't support `table.as("colA, colB")` to rename multiple columns anymore, we must use `table.as("colA", "colB")`
  - `AbstractTestBase` doesn't extend `TestBaseUtils` anymore, some tests are affected.

## Documentation


jiangxin369:
@lindong28 Could you have a look at this PR?

flink-ml-iteration/src/test/java/org/apache/flink/iteration/operator/HeadOperatorTest.java
```diff
@@ -318,6 +322,10 @@ public void testPostponeGloballyAlignedEventsAfterSnapshot() throws Exception {
                                                     OperatorUtils.getUniqueSenderId(operatorId, 0)),
                                             0)),
                             new ArrayList<>(harness.getOutput()));
+                    // TODO There might be a potential bug makes the testing hang, which could be
```
lindong28, line 325: The root cause is that `RecordingHeadOperatorFactory.latestHeadOperator` is not properly closed. The issue can be fixed by having the following code in `HeadOperatorTest#createHarnessAndRun`:

```
try {
    return runnable.apply(harness);
} finally {
    RecordingHeadOperatorFactory.latestHeadOperator.close();
}
```

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
```diff
@@ -136,31 +135,20 @@ public static StreamConfig createWrappedOperatorConfig(StreamConfig config, Clas
...
-                                                                outputTag.getTypeInfo())
-                                                        .getInnerTypeInfo()),
-                                        ((IterationRecordSerializer) typeSerializerSideOut)
-                                                .getInnerSerializer());
-                            }
+                        chainedOutput -> {
+                            OutputTag<?> outputTag = chainedOutput.getOutputTag();
+                            setTypeSerializerSideOut(outputTag, config, wrappedConfig, cl);
+                        });
+        config.getOperatorNonChainedOutputs(cl)
```
lindong28, line 144: Would it be simpler to keep the previous style and do something like this:

```
Stream.concat(
        config.getChainedOutputs(cl).stream(),
        config.getOperatorNonChainedOutputs(cl).stream())
        .forEach(
                output -> {
                    OutputTag<?> outputTag = output.getOutputTag();
                    setTypeSerializerSideOut(outputTag, config, wrappedConfig, cl);
                }
        );
```

lindong28 (commented):
Thanks for the PR. LGTM overall. Left some comments below.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/operator/OperatorUtils.java
jiangxin369: The reason why I separate these two parts is `config.getChainedOutputs(cl)` returns `List<StreamEdge>` while `config.getOperatorNonChainedOutputs(cl)` returns `List<NonChainedOutput>`, they are of different types and are not suitable to iterate in the same loop.

jiangxin369:
@lindong28 Could you have another look?

lindong28:
Thanks for the update. LGTM.

PR #235, opened 2023-04-24
[FLINK-31899] Upgrade Flink version of Flink ML to 1.17.0

jiangxin369:

Upgrade Flink version of Flink ML to 1.17.0.

## Brief change log

Upgrade Flink version of Flink ML to 1.17.0. The main changes about compatibility are sd below.
  - 1.17 introduces `CountingOutput` which is used to update metrics, we need to adapt to it.
  - 1.17 upgrades the version of Py4j to 0.10.9.7, which introduces a [bug](https://github.com/py4j/py4j/issues/506), we have to change the way to call Java methods in Python temporarily.

## Documentation

### DIFF
--- a/flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
+++ b/flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
@@ -18,8 +18,11 @@
 
 package org.apache.flink.iteration.utils;
 
+import org.apache.flink.util.Preconditions;
+
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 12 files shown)


flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
```diff
@@ -80,4 +83,55 @@ public static <T> T callMethod(
...
     }
+
+    /**
+     * The utility method to call method with the specific name and parameters.
+     *
+     * <p>Note that this method is added only for bypassing the existing bug in Py4j. It doesn't
+     * validate the classes of parameters so it can only deal with the classes that have only one
+     * method with the specific name.
+     *
+     * <p>TODO Remove this method after the Py4j bug is fixed.
```
lindong28, line 94: nits: `TODO` -> `TODO: `

flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
```diff
@@ -80,4 +83,55 @@ public static <T> T callMethod(
...
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T callMethodByName(
+            Object targetObject, Class<?> declaredClass, String methodName, Object[] parameters) {
+        List<Method> methods = new ArrayList<>();
+        for (Method m : declaredClass.getMethods()) {
```
lindong28, line 100: Is it possible to locate the method by calling `Class#getMethod(String name, Class<?>... parameterTypes)`?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
```diff
@@ -80,4 +83,55 @@ public static <T> T callMethod(
...
+        }
+    }
+
+    /**
+     * The utility method to call method with the specific name.
+     *
+     * <p>TODO Remove this method after the Py4j bug is fixed.
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T callMethodByName(
```
lindong28, line 125: Would it be simpler to update the existing `callMethod(...)` to also use the inherited methods?

flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
```diff
@@ -80,4 +83,55 @@ public static <T> T callMethod(
...
+            Object targetObject, Class<?> declaredClass, String methodName) {
+
+        try {
+            Method method = declaredClass.getMethod(methodName);
+            method.setAccessible(true);
```
lindong28, line 130: It is not necessary to use `method.setAccessible(true)` here since `Class#getMethod(...)` only returns public member method.

lindong28 (commented):
Thanks for the PR. Left some minor comments below.

flink-ml-iteration/src/main/java/org/apache/flink/iteration/utils/ReflectionUtils.java
jiangxin369: Would it be better if we keep the logic that is to be removed in a separate method? We don't have the need to use the inherited methods except for Py4j, right? So we don't have to introduce this change in the existing method.

Based on this reason, I add the method called `callMethodByName(Object targetObject, Class<?> declaredClass, String methodName)` to differ from the existing `callMethod(Object targetObject, Class<?> declaredClass, String methodName)` which has the same parameters. To keep the same naming style in the two methods I added, they are all named with `callMethodByName`, and this answers your above question.

jiangxin369:
@lindong28 I've updated the PR, could you have another look?

lindong28:
Thanks for the update. LGTM.

PR #242, opened 2023-06-09
[FLINK-32293] Support sparse vector with long index

zhipeng93:

Add support for sparse vector with long as index and double as value.

## Brief change log
- Renamed `SparseVector` as `SparseIntDoubleVector` and `DenseVector` as `DenseIntDoubleVector`.
- Added generics in `Vector` interface.
- Added `IntDoubleVector` and `LongDoubleVector` for vectors with int keys and double values, long keys and double values, respectively.
- Added test for SparseLongDouble vector.
- Added the coresponding python changes.
- Updated the corresponding flink ml docs.

## Documentation


zhipeng93:
@lindong28 @HuangXingBo Can you help to review this PR?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
```diff
@@ -24,29 +24,29 @@
...
-public interface Vector extends Serializable {
+public interface Vector<K extends Number, V extends Number> extends Serializable {
 
     /** Gets the size of the vector. */
-    int size();
+    K size();
 
     /** Gets the value of the ith element. */
-    double get(int i);
+    V get(K i);
```
lindong28, line 36: Would it be simpler to just use `double get(long i)`?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
```diff
@@ -24,29 +24,29 @@
...
 
     /** Sets the value of the ith element. */
-    void set(int i, double value);
+    void set(K i, V value);
```
lindong28, line 39: Would it be simpler to use `void set(long i, double value)`?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
```diff
@@ -24,29 +24,29 @@
...
-/** A vector of double values. */
+/** A vector representation of numbers. */
 @TypeInfo(VectorTypeInfoFactory.class)
 @PublicEvolving
-public interface Vector extends Serializable {
+public interface Vector<K extends Number, V extends Number> extends Serializable {
 
     /** Gets the size of the vector. */
-    int size();
+    K size();
```
lindong28, line 33: Would it be simpler to use `long size()`?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
```diff
@@ -24,29 +24,29 @@
...
+    V get(K i);
 
     /** Sets the value of the ith element. */
-    void set(int i, double value);
+    void set(K i, V value);
 
-    /** Converts the instance to a double array. */
-    double[] toArray();
+    /** Converts the instance to a primitive array. */
+    Object toArray();
```
lindong28, line 42: The implementation of this API always return `double[]`. Would it be simpler to still use `double[] toArray()`?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/SparseLongDoubleVector.java
```diff
@@ -0,0 +1,174 @@
...
+ * A sparse vector with long as keys and double as values.
+ *
+ * <p>TODO: Add processing logic for {@link SparseLongDoubleVector} for existing algorithms.
+ */
+@TypeInfo(SparseLongDoubleVectorTypeInfoFactory.class)
+@PublicEvolving
+public class SparseLongDoubleVector implements LongDoubleVector {
+
+    public final long n;
+    public long[] indices;
```
lindong28, line 39: Would it be useful to make these fields private and expose them via methods (e.g. `getIndices()` or `getIndicesAsLongArray()`) so that we can use the same caller code for different implementations of SparseVector?

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
```diff
@@ -24,29 +24,29 @@
...
 
     /** Converts the instance to a dense vector. */
-    DenseVector toDense();
+    Vector<K, V> toDense();
 
     /** Converts the instance to a sparse vector. */
-    SparseVector toSparse();
+    Vector<K, V> toSparse();
```
lindong28, line 48: The intention of defining this method in an interface is to have the caller use the return value via public APIs of `Vector<K, V>`.

However, it appears that caller of this method uses return values via public fields/APIs that do not belong to `Vector<K, V>`. For example, `MinHashLSHModelData#hashFunction()` accesses the field `SparseIntDoubleVector#indices` via `vec.toSparse().indices`. The code will break if we change the concrete class from `SparseIntDoubleVector` to `SparseLongDoubleVector`.

lindong28 (commented):

I created a commit to demonstrate an alternative way to support SparseVector with long index that hopefully can address most of comments mentioned below.

https://github.com/lindong28/flink-ml/commits/FLINK-32293-backup

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
```diff
@@ -24,29 +24,29 @@
 
 import java.io.Serializable;
 
-/** A vector of double values. */
+/** A vector representation of numbers. */
 @TypeInfo(VectorTypeInfoFactory.class)
 @PublicEvolving
-public interface Vector extends Serializable {
+public interface Vector<K extends Number, V extends Number> extends Serializable {
```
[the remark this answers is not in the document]
zhipeng93: I think keeping `V extends Number` would make this more scalable. Though Spark does not support float32 as value, but storing data in other data types (e.g., float32) does save the space. 

I searched the data types used in some machine learning libraries and math lib.
- Breeze supports int-key and double/float/int/long values [1]. Breeze is used by Spark ML, but Spark ML restricts the value type as double.
- Sklearn uses numpy array as input, the data types of numpy array could be float16, single, double etc. [2]
- TensorFlow/PyTorch supports different data types of tensors [3]

[1] https://github.com/scalanlp/breeze/blob/master/math/src/main/scala/breeze/linalg/DenseVector.scala
[2] https://numpy.org/doc/stable/user/basics.types.html
[3] https://www.tensorflow.org/guide/tensor#more_on_dtypes

flink-ml-servable-core/src/main/java/org/apache/flink/ml/linalg/Vector.java
zhipeng93: I have overrode the return type in `IntDoubleVector` and `LongDoubleVector`, since for different instances `toSparse` should return different types of vectors.

We can talk about this offline.

PR #246, opened 2023-07-05
[FLINK-32545] Improves the performance by optimizing row operations

jiangxin369:

Improves the performance by optimizing row operations.

Here is a performance comparison of some algorithms before and after optimization using `flink-ml-benchmark` by analyzing the seconds that each algorithm costs on average. We can see that for these algorithms, the time consumption is reduced by about 10%-25%.

|    | Before  | After  | 
|  ----  | ----  | ----  |
| Bucketizer | 48.481 | 38.185 |
| ElementWiseProduct | 43.437 | 31.094 |
| Normalizer | 55.811 | 44.629 |
| PolynomialExpansion | 72.050 | 65.050 |
| VectorSlicer | 40.515 | 31.186 |

For the other 31 algorithms, their benchmarks are configured with a small dataset, however, the optimization effect is positively correlated with dataset size, so they are of the same performance as the original on my workstation.

## Brief change log

  - Adds RowUtils to help with Row operations.
  - Replaces all Row.join() function call with new utility.

## Documentation

### DIFF
--- a/flink-ml-benchmark/src/main/resources/elementwiseproduct-benchmark.json
+++ b/flink-ml-benchmark/src/main/resources/elementwiseproduct-benchmark.json
@@ -15,18 +15,18 @@
 
 {
   "version": 1,
-  "elementwiseproduct10000000": {
+  "elementwiseproduct100000000": {
     "inputData": {
       "className": "org.apache.flink.ml.benchmark.datagenerator.common.DenseVectorGenerator",
       "paramMap": {
         "vectorDim": 5,
         "colNames": [
           [
-            "featuresCol"
+            "input"
           ]
         ],
         "seed": 2,
-        "numValues": 10000000
+        "numValues": 100000000
       }
     },
     "stage": {
--- a/flink-ml-benchmark/src/main/resources/maxabsscaler-benchmark.json
+++ b/flink-ml-benchmark/src/main/resources/maxabsscaler-benchmark.json
@@ -22,7 +22,7 @@
         "vectorDim": 100,
         "colNames": [
           [
-            "featuresCol"
+            "input"
           ]
         ],
         "seed": 2,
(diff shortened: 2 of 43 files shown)


jiangxin369:
@lindong28 Could you have a review at this PR?

lindong28:
Thanks for the PR!

For those algorithms which have the corresponding benchmark defined in ./flink-ml-benchmark/src/main/resources, can you run the benchmarks and document the performance improvements in the PR description?

Can you confirm that Bucketizer benchmark results are obtained by running `./flink-ml-benchmark/src/main/resources/bucketizer-benchmark.json`?

jiangxin369:
@lindong28 Thanks for the reply.

> For those algorithms which have the corresponding benchmark defined in ./flink-ml-benchmark/src/main/resources, can you run the benchmarks and document the performance improvements in the PR description?

For those algorithms which are already defined in "./flink-ml-benchmark/src/main/resources", I have run the benchmark for them on my workstation, but most of them remain a similar performance as before. The reason is that the other algorithms are configured with a small dataset(1/10 of Bucketizer) to make it not run so long. The PR is optimized by reducing the cost of creating rows, the effect is positively correlated with dataset size.

> Can you confirm that Bucketizer benchmark results are obtained by running `./flink-ml-benchmark/src/main/resources/bucketizer-benchmark.json`?

Yes, it is.

lindong28:
@jiangxin369 Can you help update the configuration of these machines so that they can run for e.g 30-60 seconds and validate the performance improvement of this PR?

jiangxin369:
> @jiangxin369 Can you help update the configuration of these benchmarks so that they can run for e.g 30-60 seconds and validate the performance improvement of this PR?

Sure, I have updated the PR and the description. Btw,  I'm surprised that some of the benchmarks are corrupted at runtime, I hotfix them in this PR but I think we need some strategies to detect it in advance, maybe run all benchmarks with a tiny input dataset in CI.

lindong28:
Thanks for the PR! LGTM.

PR #248, opened 2023-07-31
[FLINK-32704] Supports spilling to disk when feedback channel memory buffer is full

jiangxin369:

Supports spilling to disk when feedback channel memory buffer is full.

## Brief change log

  - Adds config to allow user to set the in-memory buffer size.
  - Adds SpillableFeedbackChannel to support spilling to disk automatically.

## Documentation


Fanoid (commented):
Thanks for your contribution. 
LGTM overall. Just leave some minor comments.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/ListStateWithCache.java
```diff
@@ -82,11 +82,13 @@ public ListStateWithCache(
...
         if (fraction > 0) {
             MemoryManager memoryManager = containingTask.getEnvironment().getMemoryManager();
+            // TODO Adds a configuration to set the fraction of the memory that the
+            // ListStateWithCache can use.
             segmentPool =
                     new LazyMemorySegmentPool(
                             containingTask,
                             memoryManager,
-                            memoryManager.computeNumberOfPages(fraction));
+                            memoryManager.computeNumberOfPages(fraction) / 2);
```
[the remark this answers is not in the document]
jiangxin369: That's a good point. I also think that the `iteration.memory.feedback-buffer.fraction` we just added is inappropriate because it is not defined in `ManagedMemoryUseCase` of Flink. So it would grab the managed memory with `ListStateWithCache` which may cause memory allocation failed. 

I changed the implementation to also use the `ManagedMemoryUseCase.OPERATOR` to share the managed memory with `ListStateWithCache`. Would you have another look?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/utils/DataStreamUtils.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public static <T> void setManagedMemoryWeight(DataStream<T> dataStream, long memoryBytes) {
+        if (memoryBytes > 0) {
+            final int weightInMebibyte = Math.max(1, (int) (memoryBytes >> 20));
+            final Optional<Integer> previousWeight =
+                    dataStream
+                            .getTransformation()
+                            .declareManagedMemoryUseCaseAtOperatorScope(
+                                    ManagedMemoryUseCase.OPERATOR, weightInMebibyte);
+            if (previousWeight.isPresent()) {
```
zhipeng93, line 46: Is there a case that one operator would set managed memory twice?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/datacache/nonkeyed/ListStateWithCache.java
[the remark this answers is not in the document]
Fanoid: The changes are acceptable to me. Thank you for your efforts in improving them!

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackChannel.java
```diff
@@ -0,0 +1,147 @@
...
+    public void initialize(
+            IOManager ioManager,
+            MemoryManager memoryManager,
+            TypeSerializer<T> serializer,
+            long inMemoryBufferSize,
+            long pageSize)
+            throws MemoryAllocationException {
+        this.queue =
+                new MpscQueue<>(ioManager, memoryManager, serializer, inMemoryBufferSize, pageSize);
+        this.initialized = true;
```
zhipeng93, line 69: Is there a case that this object is not initialized?

zhipeng93 (commented):

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
```diff
@@ -498,6 +534,11 @@ public HeadOperatorStatus getStatus() {
         return status;
     }
 
+    @VisibleForTesting
+    public SpillableFeedbackChannel getFeedbackChannel() {
```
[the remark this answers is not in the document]
jiangxin369: The feedback channel is used by both `HeadOperator` and `TailOperator`, so we cannot close it in `HeadOperator` in case of `TailOperator` is still using it.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/MpscQueue.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+            this.activeQueue = this.standByQueue;
+            this.standByQueue = ready;
+            return ready.iterate();
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    public void resetStandBy() throws Exception {
```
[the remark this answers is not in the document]
jiangxin369: I agree. Now that I have changed these methods to package private, maybe it is not necessary to rename the `resetStandBy` because it is not visible to users anymore.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackChannel.java
jiangxin369: In the current code base, no. But this is a public utility class, I would like the developers who use this class not to forget to call the `initialize` method.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/MpscQueue.java
```diff
@@ -0,0 +1,128 @@
...
+ * @param <T> The element type.
+ */
+@Internal
+public final class MpscQueue<T> implements Closeable {
+    private final Lock lock = Locks.spinLock();
+
+    private SpillableFeedbackQueue<T> activeQueue;
+    private SpillableFeedbackQueue<T> standByQueue;
+
+    private volatile boolean isClosed;
```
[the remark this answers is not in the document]
jiangxin369: There are some tests like `HeadOperatorTest#testSnapshotAndRestoreBeforeRoundZeroFinish` that are closing the operator without processing any events. So the `ConsumerTask` may be scheduled after the operator is closed and still allocates managed memory, which would cause errors.

But in an actual Flink program, the above case would not happen. So yes, the `volatile isClosed` can be removed, and just need to make the current tests process all events before closing.

jiangxin369:
@zhipeng93 Thanks for the review. I've updated the PR, could you have another look?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
```diff
@@ -0,0 +1,115 @@
...
+    private final DataOutputSerializer output = new DataOutputSerializer(256);
+    private final TypeSerializer<T> serializer;
+    private final IOManager ioManager;
+    private final MemoryManager memoryManager;
+    private final int numPages;
+
+    private List<MemorySegment> segments;
+    private ListMemorySegmentPool segmentPool;
+
+    private SpillingBuffer target;
```
lindong28, line 54: Would it be more readable to name it `buffer`?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    long size() {
+        return size;
+    }
+
+    void reset() throws Exception {
+        size = 0;
+        close();
```
lindong28, line 98: close() is generally used to release all resources used by this instance and there should be no further invocation of any method of this instance after close() is invoked.

It would be useful to follow this pattern.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
```diff
@@ -0,0 +1,115 @@
...
+        output.clear();
+        List<MemorySegment> toRelease = target.close();
+        toRelease.addAll(segments);
+        memoryManager.release(toRelease);
+        segmentPool.clear();
+    }
+
+    private void resetSpillingBuffer() throws MemoryAllocationException {
+        this.segments = memoryManager.allocatePages(this, numPages);
+        this.segmentPool = new ListMemorySegmentPool(segments);
```
lindong28, line 112: Would it be simpler to use `ListMemorySegmentSource` (defined in flink-runtime) rather than ListMemorySegmentPool (defined in flink-table-runtime)?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
```diff
@@ -0,0 +1,115 @@
...
+        size = 0;
+        close();
+        resetSpillingBuffer();
+    }
+
+    void close() throws IOException {
+        output.clear();
+        List<MemorySegment> toRelease = target.close();
+        toRelease.addAll(segments);
+        memoryManager.release(toRelease);
```
lindong28, line 106: Would it be simpler to just do `memoryManager.release(segments)`?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    long size() {
+        return size;
+    }
+
+    void reset() throws Exception {
+        size = 0;
+        close();
+        resetSpillingBuffer();
```
lindong28, line 99: Instead of repeatedly release and re-allocate this list of buffers, can we re-use the allocated buffers before/after the reset, similar to what is done in `SpillingResettableIterator#reset`?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackChannel.java
```diff
@@ -0,0 +1,142 @@
...
+    private final SubtaskFeedbackKey<T> key;
+
+    /** A single registered consumer. */
+    private final AtomicReference<ConsumerTask<T>> consumerRef = new AtomicReference<>();
+
+    /** The underlying queue used to hold the feedback results. */
+    private MpscQueue<T> queue;
+
+    /** Whether the feedback channel is initialized. */
+    private boolean initialized;
```
lindong28, line 54: Instead of adding this variable, would it be simpler to check whether `queue == null`?

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
```diff
@@ -422,16 +430,44 @@ public void close() throws Exception {
...
         int attemptNum = getRuntimeContext().getAttemptNumber();
         FeedbackKey<StreamRecord<IterationRecord<?>>> feedbackKey =
                 OperatorUtils.createFeedbackKey(iterationId, feedbackIndex);
         SubtaskFeedbackKey<StreamRecord<IterationRecord<?>>> key =
                 feedbackKey.withSubTaskIndex(indexOfThisSubtask, attemptNum);
-        FeedbackChannelBroker broker = FeedbackChannelBroker.get();
-        FeedbackChannel<StreamRecord<IterationRecord<?>>> channel = broker.getChannel(key);
-        OperatorUtils.registerFeedbackConsumer(channel, this, mailboxExecutor);
+        SpillableFeedbackChannelBroker broker = SpillableFeedbackChannelBroker.get();
+        this.feedbackChannel = broker.getChannel(key);
```
lindong28, line 444: Would it be more readable to pass the initialization logic as a parameter of type `Consumer<SpillableFeedbackChannel>` to `getChannel(...)`?

We can do the same for TailOperator. The consumer will be invoked iff a new channel needs to be created.

lindong28 (commented):

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/typeinfo/IterationRecordSerializer.java
```diff
@@ -106,7 +106,7 @@ public int getLength() {
     @Override
     public void serialize(IterationRecord<T> record, DataOutputView target) throws IOException {
         target.writeByte((byte) record.getType().ordinal());
-        serializerNumber(record.getEpoch(), target);
+        target.writeInt(record.getEpoch());
```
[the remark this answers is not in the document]
jiangxin369: The original `serializerNumber` cannot serialize `Integer.MAX_VALUE + 1`, which is the epoch of the feedback record in the last iteration. In the previous implementation, the feedback records are just passed in memory and don't need to serialize/deserialize, so the bug doesn't appear.

I also did a benchmark to serialize 1 billion integers with `serializerNumber` and `DataOutputSerializer#writeInt`, their performances are quite close, so I think we can remove the current `serializerNumber` implementation.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
jiangxin369: No, because every segment would be removed one by one from the `segments` when it is used, it can be seen in `ListMemorySegmentSource#nextSegment`. So the return value of `SpillingBuffer.close()` and the remaining `segments` together are the memories that should be freed.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
jiangxin369: `SpillingResettableIterator` is not similar to our requirement because it just iterates an existing collection and doesn't need to re-allocate more memories, however, the `SpillableFeedbackQueue` needs to receive, iterate and free items again and again. With the given API of `SpillingBuffer`, I think we need to release and re-allocate the buffers. 
Btw, I think the allocation is just taking out a piece of memories from the reserved ones which doesn't cost much.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/typeinfo/IterationRecordSerializer.java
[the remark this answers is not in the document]
lindong28: Regarding "the original serializerNumber cannot serialize Integer.MAX_VALUE + 1", can you explain why it is necessary to pass `Integer.MAX_VALUE + 1` as the serializerNumber? Is it possible to change the caller side code to avoid that?

I agree the 1-byte overhead is likely not a big deal for common cases where iteration is needed. On the other hand, if it is possible to change the caller side code as appropriate without any downside, then that approach is better than increasing the epoch overhead from 1 byte to 4 bytes.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
lindong28: With the current implementation, both HeadOperator and TailOperator will get channel via `broker.getChannel(key)`. HeadOperator is responsible for initializing this channel by calling its `initialize()` API. TailOperator will encounter exception if the channel has not been initialized.

This approach effectively relies on the callers (i.e. HeadOperator and TailOperator) to coordinator how they should access this shared object. It would be simpler if the caller code does not rely on this coordination to work.

And this code is less readable. The reason is that `SpillableFeedbackChannel#getChannel` effectively means "getOrCreateChannel" since it creates a channel if it does not already exist for the given key. However, this channel can only be used after it is initialized, which is not guaranteed by `getChannel`. By passing the initialization logic as an extra parameter to `getChannel(..)`, this method will always return a usable channel. This can make the code more readable and more self-contained. And it can also address the issue mentioned above.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/typeinfo/IterationRecordSerializer.java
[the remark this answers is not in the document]
jiangxin369: Currently, the `Integer.MAX_VALUE + 1` is passed because the termination check is in the `HeadOperator` . When the `TailOperator` processed the last epoch_watermark, it needs to increment the epoch which is already `Integer.MAX_VALUE`, and send the increased epoch to the `HeadOperator`. Then the `TerminatingHeadOperatorRecordProcessor#processFeedbackElement` would check if the epoch is equal to `Integer.MAX_VALUE + 1` to determine whether to terminate the iteration.

Of course, we can avoid it. Maybe simply change the last epoch_watermark number to `Integer.MAX_VALUE - 1`, but I don't think we need to make the code less readable for this.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/HeadOperator.java
jiangxin369: I agree with your 2nd point and I will pass the initialization logic as a `Consumer`.

But the HeadOperator and TailOperator still need to coordinate on accessing the shared channel. Because only the operator that called `declareManagedMemoryUseCaseAtOperatorScope` can initialize the Channel, while the other cannot since it has no managed memory to allocate.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/typeinfo/IterationRecordSerializer.java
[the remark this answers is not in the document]
lindong28: How about we create something like `public final END_EPOCH_WATERMARK = Integer.MAX_VALUE - 1 ` and use this variable in e.g. `TerminatingHeadOperatorRecordProcessor`? Then the code can be readable without increasing metadata overhead.

flink-ml-iteration/flink-ml-iteration-common/src/main/java/org/apache/flink/iteration/operator/feedback/SpillableFeedbackQueue.java
lindong28: Just to summarize our offline discussions:
- Currently MemoryManager#allocatePages will allocate memory from JVM rather than re-using allocated memory from a pool. We should avoid this repeated memory allocation.
- AbstractPagedOutputView, the parent class of SpillingBuffer, provides clear() and advance() to support writing to a buffer after it has been read. It might be possible to let `SpillingBuffer` implement these two APIs so that we can re-use it for writing after it is read.
- NormalizedKeySorter#reset supports re-using a buffer after it is read.

jiangxin369: After reading the code in Flink, I found that the `SerializedUpdateBuffer` is just what we need, so I replaced the SpillingBuffer with SerializedUpdateBuffer to avoid allocating memory multiple times.

lindong28:
Thanks for the update! LGTM.

PR #251, opened 2023-08-15
[FLINK-27286] Add infra to support training high dimension models

zhipeng93:
This PR introduces a communication infra that can distribute model parameters to multiple servers, so as to support training high dimension models in Flink ML. Moreover, it also introduces a programming API that can simplify the programming of iterative machine learning training process. The motivation cases include:

- Train a high dimension model whose model data cannot be fitted on a single machine.
- Support the case where the model data of each key as a double or double[].
- Support model data with discrete index.
- Use a set of index to retrieve the model values. 
- Use a set of index and UDF to retrieve the statistics computed from model values.
- Use a set of index/value to update the model values.
- Use a set of index/value to update the model, with user specified aggregation logics.
- Supports AllReduce/ReduceScatter for java generics.
- Supports AllReduce/ReduceScatter to execute every K iterations.
- Support the case that user can access the training data in each iteration.
- Output model data to downstream tasks.
- Output the intermediate result to downstream tasks.

## Brief change log
- Introduced `WorkerOperator` and `ServerOperator` to execute the training logic and maintain the model data, respectively.
- Added `MLSession` to store the information that is alive during the training process on `WorkerOperator`.
- Added `Message` to represent the information that can be transfered among workers and servers.
- Added `IterationStageList` and `IterationStage` to describe the iterative machine learning process as a chain of computation/communication stages. Also, computation stage (`ProcessStage`) and commonly used communication stage (Push/Pull/AllReduce/ReduceScatter) are also added.
- Added `ModelUpdater` to let developers describe model updating logic.
- Added `TrainingUtils` to ease the programming of iterative machine learning process.
- Added `SharedLongArray` and `SharedDoubleArray` to allow developers reuse the memory accross different iterations.
- Added unit test to cover the above functionalities.

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / no)

## Documentation

### DIFF
--- a/flink-ml-lib/pom.xml
+++ b/flink-ml-lib/pom.xml
@@ -138,6 +138,11 @@ under the License.
       <scope>test</scope>
       <type>test-jar</type>
     </dependency>
+    <dependency>
+      <groupId>it.unimi.dsi</groupId>
+      <artifactId>fastutil</artifactId>
+      <version>8.5.12</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/flink-ml-lib/src/main/java/org/apache/flink/ml/common/ps/Message.java
+++ b/flink-ml-lib/src/main/java/org/apache/flink/ml/common/ps/Message.java
@@ -1,0 +1,242 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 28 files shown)


zhipeng93:
Hi @lindong28 @Fanoid @weibozhao , can you help to review this PR?

I feel that this PR [1] is a bit complicated and it involves too many changes (i.e., infra, vectors, algorithms). So I have addressed the comments there and extract the infra here as a separate PR. 

[1] https://github.com/apache/flink-ml/pull/237

